Issue #8, opened 2018-03-22
Error when using multiple referenced libraries


neurospeech:
Lets say we have 3 projects.

```
Library
   |
   +--(Referenced by) Application
            |
            +--(Referenced by) Tests
```

In `AppData\Temp`, coverage is generated for "Application" but not for "Library", or even if it is generated, it fails with an error that could not find `Library_SOME_GUID` file. But there are multiple `Application_SOME_GUID` files in `AppData\Temp`.

I suspect, `Library_SOME_GUID` is written as `Application_SOME_GUID`

tonerdo:
I'll take a look shortly and try to fix it.

jbright:
I think that's the same issue I'm having.

wlmiller:
I was having the same issue so I dug in to see what I could find - I believe it's actually two issues.

1. If the module is never hit by a test, `CoverageTracker.MarkExecuted` isn't called for that module and no hits file is created (this would be addressed by @jbright's PR).
2. `CoverageTracker` has a single static `_path` and a single list of `_markers`, which means that all the hits get written to the same file. So even if you have hits across multiple modules, they all get dumped into one file at the end.

This fixed it for me - https://github.com/wlmiller/coverlet/commit/c9e9afb835cb25dfe006c82851fc6dcb99810703. I'm happy to submit a PR but didn't want to step on the existing PR for this issue since it's still open.

tonerdo:
@wlmiller by all means I'll love to have that PR, thanks for pointing out the issue.
You can just remove this line `if (!File.Exists(result.HitsFilePath)) { continue; }` if you're concerned about stepping on the existing PR and I also have my reservations about it

tonerdo:
This has been fixed in #14. New NuGet bits drop by Monday

tonerdo:
@wlmiller did some tests locally and found out that the implementation of `CoverageTracker.MarkExecuted` wasn't thread-safe from the onset. Your change simply brought that to light, here's the fix if you're interested https://github.com/tonerdo/coverlet/commit/c055f95f300179dd862893205cb88af53493732b

PR #11, opened 2018-03-23
CalculateCoverage: Fix file not foundexception

sbalant:
Check if the file HitsFilePath exists

--
I have the same issue as https://github.com/tonerdo/coverlet/issues/8
This simple check fix my problem

### DIFF
--- a/src/coverlet.core/Coverage.cs
+++ b/src/coverlet.core/Coverage.cs
@@ -93,6 +93,7 @@ namespace Coverlet.Core
         {
             foreach (var result in _results)
             {
+                if (!File.Exists(result.HitsFilePath)) { continue; }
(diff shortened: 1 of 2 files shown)


tonerdo:
Hi @sbalant, thanks for your contribution. I have some thoughts:

This approach just ignores the problem and it's not obvious to the user that something went wrong. A better approach will be to figure out why that error is even being thrown in the first place and try to prevent that. Possibly the file isn't being created in the first place? (https://github.com/tonerdo/coverlet/blob/master/src/coverlet.core/CoverageTracker.cs#L31)

sbalant:
No problem ;)

StillLearnin:
Any idea when this will be released to Nuget. I'm hitting this issue also.

StillLearnin:
Splendid! Thanks!

PR #44, opened 2018-04-15
Update to location of build tasks

brewsoftware:
When loading this project in VS2017 and Rider the unit tests are  currently failing due to a missing reference


tonerdo:
See https://github.com/tonerdo/coverlet#building-the-project for instructions on how to build

alexanderkozlenko:
Despite the fact that the solution is building properly using the provided instructions, it will be more convenient to have an ability for building the solution w/o additional steps. It seems that the provided changes provide a simpler way to import `props` and `targets`.

Issue #56, opened 2018-04-19
Is it possible to specify a directory to exclude from coverage?


abe545:
Hi, we would love to start using coverlet to generate code coverage, but we really don't want to include code that gets automatically generated in our coverage results. Since we generate it during CI, it wouldn't be feasible for us to apply attributes to every class that we want to exclude from coverage. We do, however, have a common folder in which the code gets generated. 

Is there a flag we can pass to exclude the files in said folder from coverage?

ryanmvt:
We're also using gRPC code generation which makes the attribute-based exclusion feature unrealistic.  A configuration or flag based solution would really help out here.

abe545:
@tonerdo the syntax suggested in #55 looks like it could be a winner - if that were a globbing pattern (and we'd have to support multiple globs). My suggestion would be a semicolon separated list of globs (probably have to quote it)

```
/p:ExcludeFromCoverage="some/**/glob;c:\some\complex rooted path\*.glob"
```

I think it would need to support both absolute and relative paths. Thoughts?

tonerdo:
I totally agree! My only thought is whether to limit this feature to just whole assemblies or to extend it to allow specifying namespaces/classes/methods and how that will work.

MarcoRossignoli:
In my scenario could be very useful also to have an "include" switch(a lot of  auxiliary class i don't test). I like the idea of json config file(in addition) because we can address both. Another concern is the case with a lot of include/exclude, command line could become hard to use and i don't like the idea to organize code for coverage, again this isn't possible on existing code. My idea is to chose the code i want to test with minimum % of coverage or fail for new or existing code.

abe545:
@tonerdo I think the easiest would be path based. This will cover the majority of the scenarios, and devs can use the attribute method to exclude methods still. I don't see a scenario where I'd want to cover *most* code generated code, but not one particular method. At the very least, I'd expect such code to be generated in its own file.

@MarcoRossignoli I think the ability to both include and exclude is valuable. I think if we help out with implementing an msbuild switch, it would be easy to include an msbuild xml file in your project (via project includes), which could define these in a file for you.

ryanmvt:
@abe545 Path-based meaning everything inside the directory specified in the path, as well as all sub-directories?  That idea would be flexible enough to encompass generated code perfectly.  It also allows you to target large swaths of code or down to a single file at a time.

[...]

PR #64, opened 2018-04-26
Implementation exclusion of files according to source file paths 

ido-namely:
(including file globbing)
Goes to #56


ido-namely:
Would love to hear what you think is the best way to test this feature

MarcoRossignoli:
> Would love to hear what you think is the best way to test this feature

@tonerdo i'm not a "msbuild workflow" expert...but if we could have a simple guide/resouce link "how to test/debug" this project will grow much faster(very likely this is obvious for a msbuild workflow dev).

src/coverlet.core/Helpers/InstrumentationHelper.cs
```diff
@@ -106,5 +108,26 @@ public static void DeleteHitsFile(string path)
...
+                matcher.AddInclude(exclusionRule);
+            }
+        
+            DirectoryInfo directoryInfo = new DirectoryInfo(parentDir);
+            
+            var fileMatchResult = matcher.Execute(new DirectoryInfoWrapper(directoryInfo));
+            return fileMatchResult.Files
+                .Select(
+                    f => System.IO.Path.GetFullPath(
+                        System.IO.Path.Combine(directoryInfo.ToString(), f.Path)
```
MarcoRossignoli: i think  instead of `directoryInfo.ToString()`  you can use `parentDir` right?
https://source.dot.net/#System.IO.FileSystem/System/IO/DirectoryInfo.cs,29
https://source.dot.net/#System.IO.FileSystem/System/IO/FileSystemInfo.cs,108

src/coverlet.core/Helpers/InstrumentationHelper.cs
```diff
@@ -3,12 +3,14 @@
...
 using System.Reflection.PortableExecutable;
+using Microsoft.Extensions.FileSystemGlobbing;
 
 using Coverlet.Core.Instrumentation;
+using Microsoft.Extensions.FileSystemGlobbing.Abstractions;
 
 namespace Coverlet.Core.Helpers
 {
-    internal static class InstrumentationHelper
+    public static class InstrumentationHelper
```
MarcoRossignoli: @tonerdo maybe`[InternalsVisibleToAttribute]`?

src/coverlet.core/Helpers/InstrumentationHelper.cs
MarcoRossignoli: nit: you can avoid `System.IO.` and use `Path.xxx`(using `System.IO;` )

src/coverlet.core/Helpers/InstrumentationHelper.cs
```diff
@@ -106,5 +108,26 @@ public static void DeleteHitsFile(string path)
 
             RetryHelper.Retry(() => File.Delete(path), retryStrategy, 10);
         }
+        
+        public static string[] GetExcludedFiles(string[] exclusionRules, string parentDir) {
+            if (exclusionRules == null || exclusionRules.Length == 0 ) return null;
```
MarcoRossignoli: IMHO usually i prefer `Array.Empty<string>()` https://docs.microsoft.com/en-us/dotnet/api/system.array.empty?view=netcore-2.0 less check condition for users.

src/coverlet.core/Helpers/InstrumentationHelper.cs
```diff
@@ -3,12 +3,14 @@
 using System.IO;
 using System.Linq;
 using System.Reflection.PortableExecutable;
+using Microsoft.Extensions.FileSystemGlobbing;
```
MarcoRossignoli: i did't know this ext...thank's!

MarcoRossignoli:
@ido-namely one useful think i've learn from my contributions(on very first one) is to use "special key" to link PR to issue for simple "navigation".

https://help.github.com/articles/closing-issues-using-keywords/

if for instance on PR explanation you write:
```
closes #56  (including file globbing)
```

a link is automatically created on issue thread and you don't need to write anything on issue thread anymore but discussion move on PR.

src/coverlet.core/Instrumentation/Instrumenter.cs
```diff
@@ -8,19 +8,23 @@
...
         private string _identifier;
+        private IEnumerable<string> _excludedFiles;
         private InstrumenterResult _result;
 
-        public Instrumenter(string module, string identifier)
+        public Instrumenter(string module, string identifier, IEnumerable<string> excludedFiles = null)
         {
             _module = module;
             _identifier = identifier;
+            _excludedFiles = excludedFiles;
```
abe545: if you change this to `_excludedFiles = excludedFiles ?? Enumerable.Empty<string>()` you won't need the null checks elsewhere

src/coverlet.core/Instrumentation/Instrumenter.cs
```diff
@@ -61,6 +65,10 @@ private void InstrumentModule()
 
                     foreach (var method in type.Methods)
                     {
+                        var sourceFiles = method.DebugInformation.SequencePoints.Select(s => s.Document.Url).Distinct();
+                        if (_excludedFiles != null && sourceFiles.Any(_excludedFiles.Contains)) {
```
abe545: Odd that the method could be declared in multiple files (I guess this is the case with `partial methods` in `partial classes`). I'm wondering if in this case, you'd only want to exclude the method if `sourceFiles.All(...)`.  @tonerdo you're obviously a lot more familiar with this - am I interpreting this correctly?

src/coverlet.core/Instrumentation/Instrumenter.cs
MarcoRossignoli: +1

tonerdo:
@ido-namely @MarcoRossignoli See steps to build and test project here: https://github.com/tonerdo/coverlet#building-the-project

src/coverlet.core/Instrumentation/Instrumenter.cs
abe545: They are part of the language: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/partial-method

If it isn't something that is supported in coverlet, then I guess we don't need to worry about it.

tonerdo: OMG I've never even heard of `partial methods` until today. I guess we learn everyday. Taking my new found knowledge into consideration I'll still go with `.First(...)`, reason being that sequence points start at the body of the method and no matter how many partials of the same method you have, you can only have one implementation and the document that implementation is contained in is where all sequence points for that method will point to

src/coverlet.core/Helpers/InstrumentationHelper.cs
tonerdo: Please lets leave this as `internal`

ido-namely:
Hey @tonerdo , found some problems, please don't merge yet

MarcoRossignoli:
@tonerdo  maybe you already know this app https://github.com/apps/wip useful to permit to PR writer to block merge.

ido-namely:
@tonerdo I fixe the problem I faced.
I also added support for absolute paths, updated the readme file and the build.proj file to use the new flag.

I see there is some problem in appveyor, not sure if it's related to the project or not.

README.md
```diff
@@ -78,13 +78,24 @@ The above command will automatically fail the build if the average code coverage
...
+#### Attributes  
 You can ignore a method or an entire class from code coverage by creating and applying any of the following attributes:
 
 * ExcludeFromCoverage
 * ExcludeFromCoverageAttribute
 
 Coverlet just uses the type name, so the attributes can be created under any namespace of your choosing.
 
+#### File Path  
+You can also ignore code coverage by specifing files using the `Exclude` property
```
tonerdo: Nit: You can also ignore specific source files from code coverage using the `Exclude` property

ryanmvt:
awesome! this looks great guys. Thank you @ido-namely for taking this on.

PR #65, opened 2018-04-26
Avoid leaving file handles open

sjp:
When calculating code coverage for a solution of mine, I found that two projects where reliably stuck on file locking.

The only process using the files were the `dotnet` process running code coverage reporting. After wrapping disposable objects in `using` statements, the problem appears to be resolved.

Additionally, I think this may also allow the `RetryHelper` code to be removed, but this has not been done yet.

### DIFF
--- a/src/coverlet.core/Instrumentation/Instrumenter.cs
+++ b/src/coverlet.core/Instrumentation/Instrumenter.cs
@@ -1,9 +1,7 @@
-using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 
 using Coverlet.Core.Helpers;
-using Coverlet.Core.Extensions;
 
 using Mono.Cecil;
 using Mono.Cecil.Cil;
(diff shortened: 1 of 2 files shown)


src/coverlet.core/Instrumentation/Instrumenter.cs
```diff
@@ -155,7 +154,7 @@ private bool IsBranchTarget(ILProcessor processor, Instruction instruction)
             return false;
         }
 
-        private void ReplaceInstructionTarget(Instruction instruction, Instruction oldTarget, Instruction newTarget)
+        private static void ReplaceInstructionTarget(Instruction instruction, Instruction oldTarget, Instruction newTarget)
```
tonerdo: Any particular reason you made this `static`?

sjp: No reason aside from them not relying on any fields in the class. It's not a functional change in behavior, mostly cosmetic.

Issue #39, opened 2018-04-11
BUG: Missing coverage


pms1969:
Just noticed this in my build pipeline 
```
...
NUnit Adapter 3.8.0.0: Test execution complete

Total tests: 163. Passed: 144. Failed: 0. Skipped: 0.
Test Run Successful.
Test execution time: 2.4027 Minutes
Results File: /tmp/build/5128ae76/.output/_85bded69-6543-4650-7e86-6797d5d60ecd_2018-04-11_12_48_55.trx

Calculating coverage result...

real	5m55.490s
user	3m20.960s
sys	0m22.580s
```

Basically, ran 144 tests, but obviously failed silently when generating the coverage file.  This happened on a number of test assemblies. :-(

Any ideas what might be causing that.  I've not been able to see that happen locally.

tonerdo:
What's `.trx`?

pms1969:
it's the logger output for dotnet test.  

To run my tests, I do the following:
```
    time \
        dotnet test \
            --no-build \
            --no-restore \
            --filter 'TestCategory!=Integration&TestCategory!=Explicit&TestCategory!=Performance' \
            --logger:trx \
            -r .output/ \
            $testproj \
            /nowarn:NU1701,NU1603,NU1605 \
            /p:CollectCoverage=true \
            /p:CoverletOutputFormat=opencover
```

the --logger:trx, produces the trx output file (which is vstest output format?).  This is then picked up by sonarqube and processed to identify failed tests.

sampwil:
@pms1969 

Once again it seems we are have some of the same issues. I think it is because we are attempting to run this in a CI/CD environment. I encountered the same no output if I have the  --no-build  and --no-restore switch. When I removed the switch I would I get the "because it is being used by another process" issue again. I discovered if I let the process build into empty directory it would fail. When I ran a dotnet publish and specified the configuration switch to it previously run publish directory it would succeed. I have put my successful patterns below.
-----------------------------------------------------------------------------------------------
dotnet publish src/Server.Tests/Server.Tests.csproj -c Debug
dotnet add src/Server.Tests/Server.Tests.csproj package coverlet.msbuild --version 1.1.0
dotnet test src/Server.Tests/Server.Tests.csproj  /p:CollectCoverage=true /p:CoverletOutputFormat=json -c Debug

-----------------------------------------------------------------------------------------------
dotnet publish src/Server.Tests/Server.Tests.csproj -c Release
[...]

PR #68, opened 2018-04-30
Additional coverage reporting for properties and nested types

### LINKED ISSUE #7: Incorrect coverage reporting 🐛
The coverage report for 87135a5d0c3c5a27db83509fba01774dd20201d9 claims that [`DictionaryExtensions.cs` has 75% coverage](https://coveralls.io/builds/16102295/source?filename=/src/coverlet.core/Extensions/DictionaryExtensions.cs). It only has one method:

https://github.com/tonerdo/coverlet/blob/87135a5d0c3c5a27db83509fba01774dd20201d9/src/coverlet.core/Extensions/DictionaryExtensions.cs#L7-L14

According to the report, lines 9, 10 and 13 are covered but not line 12 (thus the 75%). How can 13 be covered but not 12? Is this a bug or am I misreading something? The only test for this method…

https://github.com/tonerdo/coverlet/blob/87135a5d0c3c5a27db83509fba01774dd20201d9/test/coverlet.core.tests/Extensions/DictionaryExtensionsTests.cs#L10-L17

…seems to cover it 100%.

sjp:
This should fix #7 and #39.

Nested types were not being covered, which is probably why some LINQ code and anonymous types were not being marked as covered.

In addition to this I've also modified `ExcludeFromCodeCoverageAttribute` so that it can be applied to properties.

There have also been some other minor refactorings:
* Inline functions to local functions.
* Extract method for obtaining backup module path.
* Avoid hard-coding coverlet assembly name.
* Place the RetryHelper into an explicit `namespace`

Finally, I've improved performance by caching some reflection that retrieves `CoverageTracker.MarkExecuted()`. This should be a one-off operation instead of (potentially) once per sequence point.

### DIFF
--- a/src/coverlet.core/Helpers/InstrumentationHelper.cs
+++ b/src/coverlet.core/Helpers/InstrumentationHelper.cs
@@ -2,10 +2,10 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Reflection;
 using System.Reflection.PortableExecutable;
-using Microsoft.Extensions.FileSystemGlobbing;
 
-using Coverlet.Core.Instrumentation;
+using Microsoft.Extensions.FileSystemGlobbing;
 using Microsoft.Extensions.FileSystemGlobbing.Abstractions;
 
 namespace Coverlet.Core.Helpers
@@ -15,7 +15,7 @@ namespace Coverlet.Core.Helpers
         public static string[] GetDependencies(string module)
         {
             IEnumerable<string> modules = Directory.GetFiles(Path.GetDirectoryName(module), "*.dll");
-            modules = modules.Where(a => Path.GetFileName(a) != Path.GetFileName(module));
+            modules = modules.Where(a => IsAssembly(a) && Path.GetFileName(a) != Path.GetFileName(module));
             return modules.ToArray();
         }
 
@@ -41,39 +41,35 @@ namespace Coverlet.Core.Helpers
         public static void CopyCoverletDependency(string module)
         {
             var directory = Path.GetDirectoryName(module);
-            if (Path.GetFileNameWithoutExtension(module) == "coverlet.core")
-                return;
+            var moduleFileName = Path.GetFileName(module);
 
             var assembly = typeof(Coverage).Assembly;
             string name = Path.GetFileName(assembly.Location);
+            if (name == moduleFileName)
+                return;
+
             File.Copy(assembly.Location, Path.Combine(directory, name), true);
         }
 
         public static void BackupOriginalModule(string module, string identifier)
         {
-            var backupPath = Path.Combine(
-                Path.GetTempPath(),
-                Path.GetFileNameWithoutExtension(module) + "_" + identifier + ".dll"
-            );
-
+            var backupPath = GetBackupPath(module, identifier);
             File.Copy(module, backupPath);
         }
 
         public static void RestoreOriginalModule(string module, string identifier)
         {
-            var backupPath = Path.Combine(
-                Path.GetTempPath(),
-                Path.GetFileNameWithoutExtension(module) + "_" + identifier + ".dll"
-            );
+            var backupPath = GetBackupPath(module, identifier);
 
             // Restore the original module - retry up to 10 times, since the destination file could be locked
             // See: https://github.com/tonerdo/coverlet/issues/25
             var currentSleep = 6;
-            Func<TimeSpan> retryStrategy = () => {
+            TimeSpan retryStrategy()
+            {
                 var sleep = TimeSpan.FromMilliseconds(currentSleep);
                 currentSleep *= 2;
                 return sleep;
-            };
+            }
 
             RetryHelper.Retry(() => {
                 File.Copy(backupPath, module, true);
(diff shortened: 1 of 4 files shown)


sjp:
@hunterjm Yeah that's true. I'm fine with merging as your changes look to be good improvements too.

tonerdo:
@hunterjm you're gonna have a lot of merge fun then 😛. This PR seems straightforward enough, need to review yours over coffee

hunterjm:
That's fine @tonerdo, I'll keep an eye on this PR and when it get's merged I'll update mine.

src/coverlet.core/Instrumentation/Instrumenter.cs
```diff
@@ -95,7 +133,7 @@ private void InstrumentIL(MethodDefinition method)
             {
                 var instruction = processor.Body.Instructions[index];
                 var sequencePoint = method.DebugInformation.GetSequencePoint(instruction);
-                if (sequencePoint == null || sequencePoint.StartLine == 16707566)
+                if (sequencePoint == null || sequencePoint.StartLine == HiddenSequencePoint)
```
hunterjm: I would just use `!sequencePoint.IsHidden`.  Cecil already provides this info for us.

src/coverlet.core/Instrumentation/Instrumenter.cs
```diff
@@ -135,7 +173,7 @@ private Instruction AddInstrumentationCode(MethodDefinition method, ILProcessor
 
             var pathInstr = Instruction.Create(OpCodes.Ldstr, _result.HitsFilePath);
             var markInstr = Instruction.Create(OpCodes.Ldstr, marker);
-            var callInstr = Instruction.Create(OpCodes.Call, processor.Body.Method.Module.ImportReference(typeof(CoverageTracker).GetMethod("MarkExecuted")));
+            var callInstr = Instruction.Create(OpCodes.Call, processor.Body.Method.Module.ImportReference(_markExecutedMethodLoader.Value));
```
tonerdo: Any reason we can't just put `typeof(CoverageTracker).GetMethod(nameof(CoverageTracker.MarkExecuted))` here?

src/coverlet.core/Instrumentation/Instrumenter.cs
```diff
@@ -57,28 +61,62 @@ private void InstrumentModule()
...
+            if (type.CustomAttributes.Any(IsExcludeAttribute))
+                return;
+
+            foreach (var method in type.Methods)
+            {
+                if (!method.CustomAttributes.Any(IsExcludeAttribute))
+                    InstrumentMethod(method);
+            }
+
+            foreach (var property in type.Properties)
```
tonerdo: The C# compiler generates methods for properties. Those methods are returned as part of `type.Methods`, no need to explicitly instrument properties

src/coverlet.core/Helpers/InstrumentationHelper.cs
```diff
@@ -41,39 +41,35 @@ public static bool HasPdb(string module)
...
-                Path.GetTempPath(),
-                Path.GetFileNameWithoutExtension(module) + "_" + identifier + ".dll"
-            );
+            var backupPath = GetBackupPath(module, identifier);
 
             // Restore the original module - retry up to 10 times, since the destination file could be locked
             // See: https://github.com/tonerdo/coverlet/issues/25
             var currentSleep = 6;
-            Func<TimeSpan> retryStrategy = () => {
+            TimeSpan retryStrategy()
```
tonerdo: Since the retry strategy is the same in `InstrumentationHelper.cs`, we can just create a single method and use that everywhere

src/coverlet.core/Instrumentation/Instrumenter.cs
```diff
@@ -57,28 +61,62 @@ private void InstrumentModule()
...
+            {
+                if (!property.CustomAttributes.Any(IsExcludeAttribute))
+                    InstrumentProperty(property);
+            }
+
+            if (type.HasNestedTypes)
+            {
+                foreach (var nestedType in type.NestedTypes)
+                    InstrumentType(nestedType);
```
tonerdo: Did you confirm that this actually fixes the specified issues. It was my thinking that Cecil returned every type regardless of whether they were nested or not

hunterjm: When I was adding the branch logic, I saw something in OpenCover's symbol manager: https://github.com/OpenCover/opencover/blob/35f1299e136c9f91241839b85e20d7eb6ed24416/main/OpenCover.Framework/Symbols/CecilSymbolManager.cs#L183

src/coverlet.core/Instrumentation/Instrumenter.cs
sjp: Thanks. You're right. I've removed enumeration of property methods.

src/coverlet.core/Instrumentation/Instrumenter.cs
sjp: Sure. I've just changed the logic slightly to use `module.Types` instead of `module.GetTypes()` so that this logic for nested types can be retained without processing nested types more than once.

hunterjm: @sjp - Actually, I think @tonerdo is correct.  `GetTypes()` from `Mono.Cecil.Rocks` already handled nested types.  He was also correct that Properties were created as methods as seen in the attached screenshot (all that went up was hit count when properties were added).

I actually would kind of prefer letting Cecil handle type nesting.

slack - red ventures 2018-05-01 16-10-30

PR #69, opened 2018-05-01
Branch Enhancements

hunterjm:
## Description

This PR covers a lot of changes in order to better support branch coverage.  The main purpose of this refactor is two-fold.  First, to add improved support for branch coverage.  Second, to improve the output of the 3 supported report formats by adding previously missing information.

At a high level, this PR covers the objectives by:

- Resolving #67.
  - Updated all report generators with new branch logic.
- Changing Coverage Calculation from average of averages to a single average of the sums of covered/total points (for line, branch, and method).
  - Utilized the new `CoverageDetails` result model from `CoverageSummary` in report generators in lieu of manually tracking and incrementing visits and counts.

Issue #67, opened 2018-04-30
Enhanced Branch Coverage


hunterjm:
## Expected Result

When running code coverage on a project, all branches in a method will be determined, and their targets will be instrumented.

## Actual Result

Currently, only sequence points that are targets of branches are instrumented.

## Proposed Solution

Implement a branch point/target search using logic similar to [OpenCover.Framework.Symbols.CecilSymbolManager](https://github.com/OpenCover/opencover/blob/master/main/OpenCover.Framework/Symbols/CecilSymbolManager.cs#L421).  This would allow us to:

- Track Branches separately from Sequence Points.
- Instrument all Branch Targets.
- Provide more detailed branch information to the supported reporters.

## Discussion

I've already taken an initial stab at this functionality in my fork and everything seems to be working.  I am in the process of cleaning up the code and adding tests, but wanted to open this issue before opening a PR.

Major high-level changes to the structure are as follows:

- Removed `IsBranchTarget` from `Line` and `LineInfo` as branches are now tracked separately from sequence points.
- Changed `InstrumenterResult`
  - Added `Branch` which extends `Line` and adds additional properties such as `Path`
  - Changed `Document` to have both `Lines` and `Branches`
- Changed `CoverageResult`
  - Updated `Methods` to be a Dictionary with a new class `Method` as the Value
  - `Method` contains `Lines` and `Branches`
- Changed all supporting code (Instrumenter, CoverageSummary, Reporters) to handle the new model structure.

tonerdo:
@hunterjm Thanks for opening this issue and taking a stab at fixing it. I look forward to your PR. Cheers

tonerdo:
New release here: https://www.nuget.org/packages/coverlet.msbuild/2.0.0

PR #73, opened 2018-05-02
Branch Enhancements (take 2)

hunterjm:
## Description

This PR covers a lot of changes in order to better support branch coverage.  The main purpose of this refactor is two-fold.  First, to add improved support for branch coverage.  Second, to improve the output of the 3 supported report formats by adding previously missing information.

At a high level, this PR covers the objectives by:

- Resolving #67.
  - Updated all report generators with new branch logic.
- Changing Coverage Calculation from average of averages to a single average of the sums of covered/total points (for line, branch, and method).
  - Utilized the new `CoverageDetails` result model from `CoverageSummary` in report generators in lieu of manually tracking and incrementing visits and counts.

### Note

I re-forked and re-applied the changed from #69 because I managed to royally screw up the git history in that fork.  What that means is I've effectively squashed all those commits into one, so you don't get to see my fun debugging journey, and for that - I'm sorry.

### DIFF
--- a/src/coverlet.core/Instrumentation/Instrumenter.cs
+++ b/src/coverlet.core/Instrumentation/Instrumenter.cs
@@ -126,28 +126,25 @@ namespace Coverlet.Core.Instrumentation
                     index += 3;
                 }
 
-                if (targetedBranchPoints.Count() > 0)
+                foreach (var _branchTarget in targetedBranchPoints)
                 {
-                    foreach (var _branchTarget in targetedBranchPoints)
-                    {
-                        /*
-                         * Skip branches with no sequence point reference for now.
-                         * In this case for an anonymous class the compiler will dynamically create an Equals 'utility' method.
-                         * The CecilSymbolHelper will create branch points with a start line of -1 and no document, which
-                         * I am currently not sure how to handle.
-                         */
-                        if (_branchTarget.StartLine == -1 || _branchTarget.Document == null)
-                            continue;
-                        
-                        var target = AddInstrumentationCode(method, processor, instruction, _branchTarget);
-                        foreach (var _instruction in processor.Body.Instructions)
-                            ReplaceInstructionTarget(_instruction, instruction, target);
+                    /*
+                        * Skip branches with no sequence point reference for now.
+                        * In this case for an anonymous class the compiler will dynamically create an Equals 'utility' method.
+                        * The CecilSymbolHelper will create branch points with a start line of -1 and no document, which
+                        * I am currently not sure how to handle.
+                        */
+                    if (_branchTarget.StartLine == -1 || _branchTarget.Document == null)
+                        continue;
+                    
+                    var target = AddInstrumentationCode(method, processor, instruction, _branchTarget);
+                    foreach (var _instruction in processor.Body.Instructions)
+                        ReplaceInstructionTarget(_instruction, instruction, target);
 
-                        foreach (ExceptionHandler handler in processor.Body.ExceptionHandlers)
-                            ReplaceExceptionHandlerBoundary(handler, instruction, target);
+                    foreach (ExceptionHandler handler in processor.Body.ExceptionHandlers)
+                        ReplaceExceptionHandlerBoundary(handler, instruction, target);
 
-                        index += 3;
-                    }
+                    index += 3;
                 }
 
                 index++;
@@ -171,7 +168,6 @@ namespace Coverlet.Core.Instrumentation
                     document.Lines.Add(new Line { Number = i, Class = method.DeclaringType.FullName, Method = method.FullName });
             }
 
-            // string flag = branchPoints.Count > 0 ? "B" : "L";
             string marker = $"L,{document.Path},{sequencePoint.StartLine},{sequencePoint.EndLine}";
 
             var pathInstr = Instruction.Create(OpCodes.Ldstr, _result.HitsFilePath);
--- a/src/coverlet.core/Instrumentation/Instrumenter.cs
+++ b/src/coverlet.core/Instrumentation/Instrumenter.cs
@@ -6,6 +6,7 @@ using System.Reflection;
 
 using Coverlet.Core.Attributes;
 using Coverlet.Core.Helpers;
+using Coverlet.Core.Symbols;
 
 using Mono.Cecil;
 using Mono.Cecil.Cil;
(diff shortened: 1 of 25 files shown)


tonerdo:
No worries. Will take a look later today

src/coverlet.core/Instrumentation/Instrumenter.cs
```diff
@@ -99,32 +100,59 @@ private void InstrumentMethod(MethodDefinition method)
...
-                    ReplaceInstructionTarget(_instruction, instruction, target);
+                    foreach (ExceptionHandler handler in processor.Body.ExceptionHandlers)
+                        ReplaceExceptionHandlerBoundary(handler, instruction, target);
 
-                foreach (ExceptionHandler handler in processor.Body.ExceptionHandlers)
-                    ReplaceExceptionHandlerBoundary(handler, instruction, target);
+                    index += 3;
+                }
+
+                if (targetedBranchPoints.Count() > 0)
```
tonerdo: There doesn't seem to be a reason to check the count here. If it's empty the next loop simply won't run

src/coverlet.core/Instrumentation/Instrumenter.cs
```diff
@@ -143,8 +171,8 @@ private Instruction AddInstrumentationCode(MethodDefinition method, ILProcessor
                     document.Lines.Add(new Line { Number = i, Class = method.DeclaringType.FullName, Method = method.FullName });
             }
 
-            string flag = IsBranchTarget(processor, instruction) ? "B" : "L";
-            string marker = $"{document.Path},{sequencePoint.StartLine},{sequencePoint.EndLine},{flag}";
+            // string flag = branchPoints.Count > 0 ? "B" : "L";
```
tonerdo: No need for this line

src/coverlet.core/Symbols/CecilSymbolHelper.cs
```diff
@@ -0,0 +1,290 @@
...
+using Mono.Cecil;
+using Mono.Cecil.Cil;
+using Mono.Collections.Generic;
+
+namespace Coverlet.Core.Symbols
+{
+    public static class CecilSymbolHelper
+    {
+        private const int StepOverLineCode = 0xFEEFEE;
+        private static readonly Regex IsMovenext = new Regex(@"\<[^\s>]+\>\w__\w(\w)?::MoveNext\(\)$", RegexOptions.Compiled | RegexOptions.ExplicitCapture);
```
tonerdo: Can you give an example of text that'll match this pattern

tonerdo: Nit: Empty line after this

src/coverlet.core/Symbols/CecilSymbolHelper.cs
```diff
@@ -0,0 +1,290 @@
...
+                    if (null == instruction.Next)
+                        return;
+
+                    if (!BuildPointsForConditionalBranch(list, instruction, branchingInstructionLine, document, branchOffset, pathCounter, instructions, ref ordinal, methodDefinition)) 
+                        return;
+                }
+            }
+            catch (Exception ex)
+            {
+                throw new InvalidOperationException(
```
tonerdo: We should have a sane fallback instead of throwing an exception.

src/coverlet.core/Symbols/CecilSymbolHelper.cs
```diff
@@ -0,0 +1,290 @@
...
+
+                    var pathCounter = 0;
+
+                    // store branch origin offset
+                    var branchOffset = instruction.Offset;
+                    var closestSeqPt = FindClosestInstructionWithSequencePoint(methodDefinition.Body, instruction).Maybe(i => methodDefinition.DebugInformation.GetSequencePoint(i));
+                    var branchingInstructionLine = closestSeqPt.Maybe(x => x.StartLine, -1);
+                    var document = closestSeqPt.Maybe(x => x.Document.Url);
+
+                    if (null == instruction.Next)
```
tonerdo: Nit: `instruction.Next == null`

src/coverlet.core/Symbols/CecilSymbolHelper.cs
```diff
@@ -0,0 +1,290 @@
...
+    {
+        private const int StepOverLineCode = 0xFEEFEE;
+        private static readonly Regex IsMovenext = new Regex(@"\<[^\s>]+\>\w__\w(\w)?::MoveNext\(\)$", RegexOptions.Compiled | RegexOptions.ExplicitCapture);
+        public static List<BranchPoint> GetBranchPoints(MethodDefinition methodDefinition)
+        {
+            var list = new List<BranchPoint>();
+            GetBranchPoints(methodDefinition, list);
+            return list;
+        }
+        private static void GetBranchPoints(MethodDefinition methodDefinition, List<BranchPoint> list)
```
tonerdo: Why not just make this method signature `public static List<BranchPoint> GetBranchPoints(MethodDefinition methodDefinition)` and initialize the `BranchPoint` list in it

src/coverlet.core/Symbols/CecilSymbolHelper.cs
hunterjm: https://github.com/OpenCover/opencover/issues/310

src/coverlet.core/Symbols/CecilSymbolHelper.cs
hunterjm: This is the only change I haven't yet made from this review.  Do you have any suggestions?

src/coverlet.core/Symbols/CecilSymbolHelper.cs
hunterjm: It reduces the number of times we have to return the list if we need to short circuit the method.  Other than that, it's the structure that existed in OpenCover: https://github.com/OpenCover/opencover/blob/master/main/OpenCover.Framework/Symbols/CecilSymbolManager.cs#L351

tonerdo: I'll appreciate if you change it, thanks

src/coverlet.core/Symbols/CecilSymbolHelper.cs
tonerdo: Perhaps put the `try...catch` inside the loop and simply `continue` to ignore the error. I don't see much point throwing an exception because the developer most likely won't know what/how to fix it. IMO better to have some missing coverage info than to have the entire thing fail with no directions on how to fix it. Also if this file was pulled from OpenCover I think it'll be right to attribute it to the project

hunterjm:
All requested changes completed.

Issue #70, opened 2018-05-01
Massive disk usage when running on .NET Core / Ubuntu


Murodese:
One of our build runners is Ubuntu-based currently, building a reasonably large project (~15 projects in a single solution). Using `dotnet test /p:CollectCoverage=true /p:CoverletOutputFormat=opencover`, I'm seeing some files created in /tmp that are fairly enormous (naming scheme `<project>_<hash>`) - one of the files covering our largest project is nearly 90gb - about 180gb for all projects combined. Disabling coverlet, the files don't get created at all, so I'm assuming they're cache files of some kind.

Should they be this big? Our build runner doesn't have that much disk space allocated, so it keeps bringing it down :|

hunterjm:
@Murodese - That would the hits file.  It adds a new line for every sequence point hit, soon to be branch points as well.  Those must be some massive projects.  The hits file for this project is 133MB on my machine, and that's on my branch which adds additional instrumentation for better branch coverage.

@tonerdo - It might be worthwhile looking into reading and writing from a `GZipStream`.  The markers should compress pretty well.

tonerdo:
@hunterjm @Murodese this has already been fixed in #58. New NuGet package with the fix going out in a couple of hours

tonerdo:
On closer look, this looks like more of a disk usage thing than a RAM usage thing. Definitely need to consider your approach @hunterjm

OneCyrus:
yes, the memory problem comes later. but should be fixed in the next release. otherwise you would need 90-180GB memory as well to process this file.

the GZipStream could be a good option to solve this problem. shouldn't be too hard to implement.

hunterjm:
@tonerdo @OneCyrus - Initial take on using GZipStream is on my fork.  If I merge https://github.com/hunterjm/coverlet/pull/1, it will become a part of #69.  Otherwise, I can wait for #69 to be merged and re-target https://github.com/hunterjm/coverlet/pull/1 to master.

hunterjm:
@tonerdo - Agreed.  I just made the changes locally off of the #69 branch, and since the fix touches `Coverage.cs` there would be merge conflicts if I tried to open a separate PR to master.

tonerdo:
New release with disk usage fix - https://www.nuget.org/packages/coverlet.msbuild/2.0.0

PR #88, opened 2018-05-05
GZip Compression on hits files

hunterjm:
Helps reduce some of the issues seen in #70

### DIFF
--- a/src/coverlet.core/CoverageTracker.cs
+++ b/src/coverlet.core/CoverageTracker.cs
@@ -50,7 +50,6 @@ namespace Coverlet.Core
         {
             foreach (var kvp in _markers)
             {
-                _markerFileCount.TryAdd(kvp.Key, 0);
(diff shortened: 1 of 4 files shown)


src/coverlet.core/CoverageTracker.cs
```diff
@@ -37,7 +49,18 @@ public static void MarkExecuted(string path, string marker)
         public static void CurrentDomain_ProcessExit(object sender, EventArgs e)
         {
             foreach (var kvp in _markers)
-                File.AppendAllLines(kvp.Key, kvp.Value);
+            {
+                _markerFileCount.TryAdd(kvp.Key, 0);
```
tonerdo: This seems redundant

hunterjm: It is. Will remove.

hunterjm:
For this project it changed output from ~133 MB to ~13 MB, so a 10x reduction.

Issue #90, opened 2018-05-07
Add html report as an output format


patros:
Is this an option that you would like added @tonerdo? I'd be happy to do some work in this space. I quite like the reports generated by coverage-viewer. It is written in node.js so it would need to be ported over to avoid the node.js dependency.

Edit: And I just double checked the licence for coverage-viewer and it is MIT as well.

patros:
@hunterjm good point about having to maintain the html report generation. With a few good options maintained outside of this project the next question is whether Toni wants to include human readable output formats or to let the user choose their html generator of choice.

tonerdo:
@patros I'm currently leaning towards letting the user choose their html generator of choice. But definitely something to think about

patros:
@tonerdo I'll close this issue. It can always be reopened if you start to lean the other way.

PR #105, opened 2018-05-21
Implementation of HTML Reports

hunterjm:
I know that you mentioned in #90 that you were leaning toward letting people choose their own HTML report generator, but my PR into ReportGenerator was recently merged, and this is a completely optional way of letting people who want to generate both coverage results and HTML reports in one command do so.  If the new, optional, `ReportTypes` property is not specified, nothing changes.

### DIFF
--- a/src/coverlet.msbuild.tasks/CoverageResultTask.cs
+++ b/src/coverlet.msbuild.tasks/CoverageResultTask.cs
@@ -7,6 +7,7 @@ using Coverlet.Core;
 using Coverlet.Core.Reporters;
 using Microsoft.Build.Framework;
 using Microsoft.Build.Utilities;
+using Palmmedia.ReportGenerator.Core;
 
 namespace Coverlet.MSbuild.Tasks
 {
@@ -14,6 +15,7 @@ namespace Coverlet.MSbuild.Tasks
     {
         private string _filename;
         private string _format;
+        private string _reportTypes;
         private int _threshold;
         private string _thresholdType;
 
@@ -29,6 +31,12 @@ namespace Coverlet.MSbuild.Tasks
         {
             get { return _format; }
             set { _format = value; }
+        }
+
+        public string ReportTypes
+        {
+            get { return _reportTypes; }
+            set { _reportTypes = value; }
         }
 
         [Required]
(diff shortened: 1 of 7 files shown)


src/coverlet.msbuild.tasks/CoverageResultTask.cs
```diff
@@ -7,13 +7,15 @@
 using Coverlet.Core.Reporters;
 using Microsoft.Build.Framework;
 using Microsoft.Build.Utilities;
+using Palmmedia.ReportGenerator.Core;
```
coenm: I don't think coverlet should have a dependency on this specific report generator.

hunterjm:
@coenm - I see where you are going with that.  Really this feature is completely optional.  It does add a separate dependency to coverlet, but there are plenty of code coverage tools that include HTML report generation as a part of the base package.  Istanbul for JavaScript, and Cobertura for Java are good examples of this.

I added this functionality more for simplicity than anything else.  For simple projects that don't utilize a build script or custom MSBuild configuration, it seems overkill to have them add an extra step to add one.  I suppose I could make a separate nuget package that hooks into MSBuild and runs after coverlet's `GenerateCoverageResult` task.

@tonerdo, if you believe that would be a better approach, let me know.

tonerdo:
I tend to agree with @coenm, this feature will be good as an add-on but doesn't really fit into the core library. Here a some reasons behind my reservation:

* I imagine a situation where people might have issues with the html generation feature and open them in this repo.
* This will take away visibility from the ReportGenerator project. I prefer Coverlet to focus on just the coverage bit and co-exist with supporting projects as opposed to trying to do everything in itself.

That being said, I understand the work that went into this. I'll leave this PR open and if there's considerable interest for this feature from the community then I'll merge it in

hunterjm:
@tonerdo - no worries.  Wasn't too hard to implement.  I do really like the simplicity of hooking into dotnet test for both, however.  I'm going to explore creating a separate package that can marry the two and create a separate task that executes after `GenerateCoverageResult` to handle report generation.

ppumkin:
0This is a great feature and I agree that this should not be in the Core

Problem is that we cannot consume Coverlet.Core because there is no package. The only way is to download the code and use that.. or make Private packages.

Please check this issue. https://github.com/tonerdo/coverlet/issues/212

MarcoRossignoli:
We close this stale PR for the moment, we'll re-open in future if needed.

Issue #72, opened 2018-05-01
BUG: (Completely) Missing Coverage


nopara73:
Maybe related: https://github.com/tonerdo/coverlet/issues/39

# Behavior.

![image](https://user-images.githubusercontent.com/9156103/39472085-3d11d312-4d71-11e8-8765-549932ddbd8d.png)

# What I do

1. Add coverlet.msbuild NuGet to my test project.
2. Run Tor (needed for my tests to successfully complete.)
3. Run `dotnet test /p:CollectCoverage=true`.  

I also tried creating a new project and everything works fine, however with my current project (https://github.com/zkSNACKs/WalletWasabi/) it silently fails.

nopara73:
Note: Neither the `coverage.xml`, nor the `coverage.json` files are generated.

nopara73:
And travis (both on Linux and OSX) is even stranger:

https://travis-ci.org/zkSNACKs/WalletWasabi/jobs/373441646
https://travis-ci.org/zkSNACKs/WalletWasabi/jobs/373441646

![image](https://user-images.githubusercontent.com/9156103/39472792-b15a7a86-4d75-11e8-9b39-5b61d5e757f1.png)

nopara73:
For the record, here are the commands for only doing one file.

```
dotnet restore && dotnet build && cd WalletWasabi.CodeCoverage
dotnet minicover instrument --workdir . --assemblies ../WalletWasabi.Tests/bin/**/*.dll --sources ../WalletWasabi.Backend/Controllers/ChaumianCoinJoinController.cs
dotnet minicover reset
cd ../WalletWasabi.Tests && dotnet test --no-build
cd ../WalletWasabi.CodeCoverage
dotnet minicover uninstrument --workdir .
dotnet minicover report --workdir . --threshold 50
```

tonerdo:
Thanks for reporting this. I'll take a look at your sample project and see if I can figure it out

nopara73:
@tonerdo Thanks! I put together some build instructions for Windows so to not get stuck:

1. Get .NET Core
2. Download the Tor Experd Bundle (Not the browser.) https://www.torproject.org/download/download
3. Run `tor.exe`
4. Fire up command line:
```
git clone https://github.com/zkSNACKs/WalletWasabi
cd WalletWasabi
dotnet restore && dotnet build
cd WalletWasabi.Tests
dotnet test
```

Notes:

[...]

PR #106, opened 2018-05-21
Logging concept

coenm:
Since version 2.0 I have an issue with a project running on Linux where coverage is not calculated (https://github.com/tonerdo/coverlet/issues/72 and https://github.com/tonerdo/coverlet/issues/107) . Running on Windows doesn't have any problems. My problem is that it is pretty hard to do investigation on Travis and why my project isnt covered. 

Would it be nice if you can enable logging in order to give the user the ability to do a first investigation (and maybe fix and send a PR) before adding an issue.

First of all, I dislike lines like `logger.Log("bla bla");` through all production code at all kind of places. IMO too much of these lines blur the original intention of the code. 

This PR is a starting point of how logging can be inplemented. Hopefully, you all see the value of logging and we can have a discussion about this. 

The logging is done by decorating classes with a `LoggingDecorator`. 

1) https://github.com/coenm/coverlet/commit/f28a7047d2800026c2f8e4ab41e97c16a44a0c04 Added ILogger, ConsoleLogger, and LoggerFactory. No functionality changed.
2) https://github.com/coenm/coverlet/commit/0888e0d7bb3ca48c14dcd0d8ddcd264ddb223e63 Added new interface to existing Instrumenter class. No functionality changed.
3) https://github.com/coenm/coverlet/commit/3cd15cecea760cc7fa93cb18178879fc468e19e9 Creation of Instrumenter is done using static Factory class. No functionality changed.
4) https://github.com/coenm/coverlet/commit/2010ed942504b01211e73fc20eab1ed87ac5095b Added InstrumenterLoggerDecorator to decorate an instance of IInstrumenter. No functionality changed.
5) https://github.com/coenm/coverlet/commit/6f2fc30a012e6be44d923aaf9439f416b12c2ce8 Added functionality to register the decorator to the InstrumenterFactory. EnableLogging class has responsibility to create and add the logging decorator to the InstrumenterFactory. No functionality changed.
6) https://github.com/coenm/coverlet/commit/80aef4b6ee7537cbc95990a072aa30ef24343f1d This step adds a property Verbose to the InstrumentationTask and when this property is true, it enables Logging.
7) https://github.com/coenm/coverlet/commit/b4e66f7adc6970a08ea1e0351f9a634364c2ee3a Seperated classes to own files.

Please note that the original code was extended (decorated) and hardly touched. Only the creation of `Instrumenter` is now done using a factory, and the `Instrumenter` class now implements an interface.

If you like this idea, I'm willing to take this further

BTW, I don't expect this PR to be accepted as is right now as it doesn't add real value at this point.
I just didn't want to put too much effort in this without knowing what the you think about logging and about this concept.

### DIFF
--- a/src/coverlet.core/Coverage.cs
+++ b/src/coverlet.core/Coverage.cs
@@ -37,7 +37,7 @@ namespace Coverlet.Core
                 if (InstrumentationHelper.IsModuleExcluded(module, _filters))
                     continue;
 
-                var instrumenter = new Instrumenter(module, _identifier, _filters, excludedFiles);
+                var instrumenter = InstrumenterFactory.Create(module, _identifier, _filters, excludedFiles);
                 if (instrumenter.CanInstrument())
                 {
                     InstrumentationHelper.BackupOriginalModule(module, _identifier);
--- a/src/coverlet.core/Coverage.cs
+++ b/src/coverlet.core/Coverage.cs
@@ -30,7 +30,7 @@ namespace Coverlet.Core
         {
             string[] modules = InstrumentationHelper.GetCoverableModules(_module);
             string[] excludedFiles =  InstrumentationHelper.GetExcludedFiles(_rules);
-            _filters = _filters?.Where(f => InstrumentationHelper.IsValidFilterExpression(f)).ToArray();
+            _filters = _filters?.Where(InstrumentationHelper.IsValidFilterExpression).ToArray();
 
             foreach (var module in modules)
             {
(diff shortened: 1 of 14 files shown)


coenm:
BTW, I don't expect this PR to be accepted as is right now as it doesn't add real value at this point. I also didn't add any tests.
I just didn't want to put too much effort in this without knowing what the you think about logging and about this concept.

coenm:
@tonerdo Are you interested in a logging feature. Would probably make investigating bugs like #107 easier.

PR #112, opened 2018-06-04
Add support of ExcludeFromCodeCoverage attribute

derhally:
This adds support for excluding items by using the System.Diagnostics.CodeAnalysis.ExcludeFromCodeCoverage attribute or an attribute with the same name in addition to the ExcludeFromCoverage attribute.

Added some unit tests to Instrumenter class.

### DIFF
--- a/src/coverlet.core/Instrumentation/Instrumenter.cs
+++ b/src/coverlet.core/Instrumentation/Instrumenter.cs
@@ -24,14 +24,6 @@ namespace Coverlet.Core.Instrumentation
         private readonly static Lazy<MethodInfo> _markExecutedMethodLoader = new Lazy<MethodInfo>(GetMarkExecutedMethod);
         private InstrumenterResult _result;
 
-        private static readonly string[] _excludeAttributeNames = new[]
-        {
-            nameof(ExcludeFromCoverageAttribute),
-            "ExcludeFromCoverage",
-            nameof(ExcludeFromCodeCoverageAttribute),
-            "ExcludeFromCodeCoverage"
-        };
-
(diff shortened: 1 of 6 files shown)


src/coverlet.core/Instrumentation/Instrumenter.cs
```diff
@@ -269,7 +270,10 @@ private static void ReplaceExceptionHandlerBoundary(ExceptionHandler handler, In
 
         private static bool IsExcludeAttribute(CustomAttribute customAttribute)
         {
-            return customAttribute.AttributeType.Name == nameof(ExcludeFromCoverageAttribute) || customAttribute.AttributeType.Name == "ExcludeFromCoverage";
+            var attributeName = customAttribute.AttributeType.Name;
+
+            return attributeName == nameof(ExcludeFromCoverageAttribute) || attributeName == "ExcludeFromCoverage"
+                || attributeName == nameof(ExcludeFromCodeCoverageAttribute) || attributeName == "ExcludeFromCodeCoverage";
```
tonerdo: It's probably better to have the attribute names in an array and do a simple `array.Any(...)`

src/coverlet.core/Instrumentation/Instrumenter.cs
```diff
@@ -24,6 +24,14 @@ internal class Instrumenter
         private readonly static Lazy<MethodInfo> _markExecutedMethodLoader = new Lazy<MethodInfo>(GetMarkExecutedMethod);
         private InstrumenterResult _result;
 
+        private static readonly string[] _excludeAttributeNames = new[]
```
tonerdo: I honestly don't think we need this as a field. A local variable is just fine

Issue #36, opened 2018-04-10
How can I union coverage results in one file?


Valeriy1991:
Hello! 

I have multiple tests projects - one with unit tests and one with integration tests. How can I calculate code coverage for both projects with coverlet?

dmccaffery:
This is especially important if attempting to upload test results to VSTS, which is expecting a single summary file (and grabs the last match if using a glob pattern).

dmccaffery:
![grumble, grumble](https://user-images.githubusercontent.com/1803684/39091265-73faffe4-45a5-11e8-914f-e4ac5bc32ba8.png)

OneCyrus:
is there an API to hook in to once all tests are complete? i guess currently a single instance of coverlet doesn't know if it's the last time a test is called and would need to consolidate the results for every test project.

Hugh-walsh:
We run several tests in parallel using gitlab CI so would like to take the output files and combine them to give an overall result.  Would be nice if this were possible.

patros:
@aptester's suggestion works well for generating a report. But not a combined xml for VSTS unfortunately :(

skoruba:
@tonerdo What do you think about it? Do you have any suggestions for it? Thx.

sjk07:
i want this so bad! :wink:

tonerdo:
I'm gonna explore how existing tools achieve this and see how best to implement since it's a popular ask. Will go a lot faster if I can get some help 😄

basilfx:
I've started something [over here](https://github.com/basilfx/coverlet/commits/feature/merge_ir), which I'm  currently testing and will PR if ready. This basically allows one to merge the coverage result of one run, with an intermediate result of a previous run.

It's somewhat similar to what OpenCover offers.

Note that this will generate the 'final' report for each intermediate run, so there is some overhead, but should just be as simple as adding `/p:CoverletIntermediateResult=output.json`.

tonerdo:
Awesome @basilfx, thanks a lot for the help

basilfx:
The PR is here: https://github.com/tonerdo/coverlet/pull/123.

tonerdo:
@basilfx can you please point me to the where OpenCover documents its implementation of this feature? Thanks

basilfx:
You can find it here: https://github.com/opencover/opencover/wiki/Usage

Look for -merge output and -mergebyhash.

PR #123, opened 2018-06-18
Add support for merging results

basilfx:
This PR adds support for merging results of multiple projects. It is similar to how OpenCover supports this.

It is as simple as adding `/p:CoverletIntermediateResult=intermediate.json` to the arguments. Just ensure that this file doesn't exist when you would record the test results of multiple projects.

I have created [this](https://github.com/basilfx/Coverlet-Test) test project (you have to fix reference to Coverlet package), which demonstrates how two projects that both cover 50% will yield 100% when merging the intermediate results. I have also verified that the report generated by ReportGenerator looks as expected.

```

Merging with intermediate result...
  Intermediate result written to '../intermediate.json'

Writing report(s)...
  Generating report 'C:\Projects\GitHub\Coverlet-Test\tests\CoverletTests.FooTests\coverage.opencover.xml'
  Generating report 'C:\Projects\GitHub\Coverlet-Test\tests\CoverletTests.FooTests\coverage.cobertura.xml'

+--------------+--------+--------+--------+
| Module       | Line   | Branch | Method |
+--------------+--------+--------+--------+
| CoverletTest | 100%   | 100%   | 100%   |
+--------------+--------+--------+--------+

| Average | 100%   | 100%   | 100%   |
```

References:
#36

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,16 @@ dotnet test /p:CollectCoverage=true /p:Threshold=80 /p:ThresholdType=line
 
 You can specify multiple values for `ThresholdType` by separating them with commas. Valid values include `line`, `branch` and `method`.
 
+### Intermediate Result
+
+For combining the results of multiple projects, it is possible to use an intermediate result by using the `CoverletIntermediateResult` property. Coverage output will be merged with an intermediate result before generating the report(s). Ensure that all test runs point to the same intermediate result file.
+
+```bash
+dotnet test /p:CollectCoverage=true /p:CoverletIntermediateResult=intermediate.json
+```
+
+_Note: When using build automation, ensure that this intermediate result file is removed first. It doesn't make sense to merge with an intermediate result from a different build!_
+
 ### Excluding From Coverage
 
 #### Attributes
--- a/src/coverlet.console/Program.cs
+++ b/src/coverlet.console/Program.cs
@@ -18,6 +18,7 @@ namespace coverlet.console
 
             CommandArgument project = app.Argument("<PROJECT>", "The project to test. Defaults to the current directory.");
             CommandOption config = app.Option("-c|--configuration", "Configuration to use for building the project.", CommandOptionType.SingleValue);
+            CommandOption intermediateResult = app.Option("-i|--coverage-intermediate-result", "The output path of intermediate result (for merging multiple runs).", CommandOptionType.SingleValue);
             CommandOption output = app.Option("-o|--coverage-output", "The output path of the generated coverage report", CommandOptionType.SingleValue);
             CommandOption format = app.Option("-f|--coverage-format", "The format of the coverage report", CommandOptionType.SingleValue);
 
@@ -34,6 +35,9 @@ namespace coverlet.console
                     dotnetTestArgs.Add($"-c {config.Value()}");
 
                 dotnetTestArgs.Add("/p:CollectCoverage=true");
+
+                if (intermediateResult.HasValue())
+                    dotnetTestArgs.Add($"/p:CoverletIntermediateResult={intermediateResult.Value()}");
 
                 if (output.HasValue())
                     dotnetTestArgs.Add($"/p:CoverletOutput={output.Value()}");
--- a/src/coverlet.core/App.config
+++ b/src/coverlet.core/App.config
@@ -1,0 +1,11 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<configuration>
+  <runtime>
+    <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
+      <dependentAssembly>
+        <assemblyIdentity name="Newtonsoft.Json" culture="Neutral" publicKeyToken="30ad4fe6b2a6aeed" />
+        <bindingRedirect oldVersion="0.0.0.0-10.0.0.0" newVersion="9.0.0.0" />
+      </dependentAssembly>
+    </assemblyBinding>
+  </runtime>
+</configuration>
--- a/src/coverlet.core/Coverage.cs
+++ b/src/coverlet.core/Coverage.cs
@@ -66,19 +66,19 @@ namespace Coverlet.Core
                             {
                                 if (methods.TryGetValue(line.Method, out Method method))
                                 {
-                                    documents[doc.Path][line.Class][line.Method].Lines.Add(line.Number, new LineInfo { Hits = line.Hits });
+                                    documents[doc.Path][line.Class][line.Method].Lines.Add(line.Number, new HitInfo { Hits = line.Hits });
                                 }
                                 else
                                 {
                                     documents[doc.Path][line.Class].Add(line.Method, new Method());
-                                    documents[doc.Path][line.Class][line.Method].Lines.Add(line.Number,  new LineInfo { Hits = line.Hits });
+                                    documents[doc.Path][line.Class][line.Method].Lines.Add(line.Number,  new HitInfo { Hits = line.Hits });
                                 }
                             }
                             else
                             {
                                 documents[doc.Path].Add(line.Class, new Methods());
                                 documents[doc.Path][line.Class].Add(line.Method, new Method());
-                                documents[doc.Path][line.Class][line.Method].Lines.Add(line.Number,  new LineInfo { Hits = line.Hits });
+                                documents[doc.Path][line.Class][line.Method].Lines.Add(line.Number,  new HitInfo { Hits = line.Hits });
                             }
                         }
                         else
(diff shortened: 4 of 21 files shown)


basilfx:
Here is a more detail explanation of the changes:

The results are merged `CoverageResultTask.cs`, just before the reports are generated. I've added `Read(string path)` to `IReporter`, to convert a previous result back into a `CoverageResult`. I have only implemented this for the `JsonReporter`, but one could extend this for other reporters, if it makes sense.

To merge the coverage results, I have `CoverageResultExtensions`, which isolates all of the merge operations. However, I needed to rewrite the way branches were recorded. I changed it from

```c#
public class Branches : SortedDictionary<int, List<BranchInfo>> { }
```

into 

```c#
public class Branches : SortedDictionary<(int Number, int Offset, int EndOffset, int Path, uint Ordinal), HitInfo> { }
```

I found that a tuple of (Number, Offset, EndOffset, Path, Ordinal) uniquely identifies the branch (therefore its the key), and that the number of hits is its value.

Unfortunately, the JSON serializer (Jil) does [not support keys](https://github.com/kevin-montrose/Jil/blob/f08b5b3c8b9e9ff152277f5dfab42ef7ac22e768/Jil/Serialize/InlineSerializer.cs#L2910) other than strings or integers. And since it also doesn't support converters, I could not rewrite this without converting the `CoverageResult` into a serializable structure. The easiest option was to move to Newtonsoft.Json.

The story doesn't stop here, as Newtonsoft.Json also has some troubles deserializing a tuple (which happens when the intermediate result is read back into memory). I therefore implemented two custom converters to store line information and branch information. This means that the Coverlet JSON format has changed to this:

```json
...
"System.Boolean CoverletTest.Application::DoSomething(System.Boolean)": {
  "Lines": [
    {
      "Key": {
        "Line": 21
      },
      "Value": {
        "Hits": 2
      }
    }
  ],
  "Branches": [
    {
      "Key": {
        "Number": 11,
        "Offset": 4,
        "EndOffset": 22,
        "Path": 1,
        "Ordinal": 1
      },
      "Value": {
        "Hits": 1
      }
    }
  ]
}
...
```

There are still a few things that I haven't fixed (properly). One of them is the way how the results are stored per assembly. If you take the test project I created, you'll end up with twice `CoverletTest.dll`, where one will reside in `CoverletTests.FooTests\bin\Debug` and the other in `CoverletTests.BarTests\bin\Debug`. This will create two assembly entries in the `CoverageResult`, while technically, they are the same (but just available in two different folders). This results in a result like this:

```

Merging with intermediate result...
  Intermediate result written to '../intermediate.json'

Writing report(s)...
  Generating report 'C:\Projects\GitHub\Coverlet-Test\tests\CoverletTests.BarTests\coverage.opencover.xml'
  Generating report 'C:\Projects\GitHub\Coverlet-Test\tests\CoverletTests.BarTests\coverage.cobertura.xml'

+--------------+--------+--------+--------+
| Module       | Line   | Branch | Method |
+--------------+--------+--------+--------+
| CoverletTest | 61,1%  | 50%    | 66,7%  |
+--------------+--------+--------+--------+
| CoverletTest | 61,1%  | 50%    | 66,7%  |
+--------------+--------+--------+--------+

| Average | 61,1%  | 50%    | 66,7%  |
```

I solved this (for now) by just taking the file name and stripping the full path. A better solution would be preferred, but this will change the resulting JSON again.

sjk07:
@basilfx All this work is awesome! I really want this to happen. Just a few things that came up when thinking this problem though that i wanted to put out there. 

I'm not familiar with the code base at all, so please excuse me for being so naive.

One question i would ask is why are we bothering to output the intermediate results to a file in the first place? I would think the most efficient and straight forward way to make this happen is to have an in memory collection of all tests ran and keep adding the results until all `found` tests have completed. The user could also input a directory / list where all the test projects are located. Then a single file can be output to a default / location of the users liking. 

I would think this would eliminate your serialization issues and knowing all the test projects before hand will allow you to merge all the assemblies

As i said originally, this is probably very naive since I have not taken the time to look though how the code currently works. Just wanted to put it out there

basilfx:
@sjk07 Thanks for the feedback. The reason why I use a intermediate results file is because they are used across multiple processes. From a coverlet point of view, it is not aware if a particular run is the first or the last.

I also don’t know the tests on beforehand. I could use a wildcard to run all tests, or run dotnet test on the solution.

My approach is easy. It doesn’t require scripting (e.g. to discover the test) so it runs very easily in environments such as VSTS.

tonerdo:
@basilfx thanks for taking the time out to look into this. Will take a look once I get the chance. Don't worry about the Travis failure for now, the .NET Core runtime installed on it doesn't support assembly signing. Will move to Appveyor Linux build agents

tonerdo:
Hey @basilfx, please rebase master so that the CI build can kick in. I've transitioned the Linux tests over to Appveyor which already runs all Windows tests

basilfx:
@tonerdo Rebased. It seems to work :-)

tonerdo:
Hey @basilfx, I've taken a look through your code and I do appreciate the effort. I have a couple of questions as well as thoughts on how to make this implementation much simpler.

* Any particular reason you renamed `LineInfo` to `HitsInfo`?
* In terms of the implementation I'm a bit confused why you decided to go with using tuples as the key for branches. Wasn't the existing information in the coverage result json representative enough and the BranchInfo object directly usable.
* In terms of user experience I'd expect something along the lines of `/p:Merge=/path/to/coverage.json` where the file specified is the coverage result output of a previous run from another project
* I don't think there's a need to call merge specifically in the msbuild tasks. A much simpler implementation that comes to mind will be to simply pass the supplied file in the `Merge` property to the `Coverage` class, deserialize it and perform the merge within that class. Merge functionality should be a function of the core library, msbuild tasks should be limited entirely to presentation of results.
* > which demonstrates how two projects that both cover 50% will yield 100% when merging the intermediate results.

  From my knowledge of how merging coverage results work I should expect that the situation above should only be true if each of the two projects cover a _different_ 50% of the assembly under test. Is that the case with your implementation?
* It's ok for Coverlet to only support merging results in its own `json` format.

In summary, I think it'll be simpler and a lot cleaner to just accept a `coverage.json` pass that down to `Coverage` class and have a single merge method within that class that performs the merging. I do have a merge algorithm in mind that should greatly simplify things and remove the duplicate entry issue you're encountering.

Let me know your thoughts

basilfx:
Thanks @tonerdo for looking at the code. Here are my thoughts:

> In terms of the implementation I'm a bit confused why you decided to go with using tuples as the key for branches. Wasn't the existing information in the coverage result json representative enough and the BranchInfo object directly usable.

No, the `BranchInfo` object wasn't directly usable. As far as I understand, every branch is uniquely identified by a combination of values. Therefore you have a mapping of `(Number, Offset, EndOffset, Path, Ordinal) -> Hits`. In the old situation you had `(Number) -> (Offset, EndOffset, Path, Ordinal, Hits)`, which is a bit odd (and wrong, IMHO). It also requires more effort when merging the results, because you have to do multiple lookups.

I think that this should be changed, apart from merging the coverage results. That's why I added it as a separate commit.

I guess it should be easy to convert this tuple into a object, e.g. `BranchPoint`, but that would require more work. This was the easiest.

> Any particular reason you renamed `LineInfo` to `HitsInfo`?

Since `BranchInfo` was stripped down, both `LineInfo` and `BranchInfo` only contained the number of hits. I therefore renamed them to `HitsInfo`.

> In terms of user experience I'd expect something along the lines of `/p:Merge=/path/to/coverage.json` where the file specified is the coverage result output of a previous run from another project

I fully agree, but the point is that Coverlet support multiple output formats. I think it is therefore more confusing for the end-user that it doesn't work whenever JSON format is not selected, so have you any idea's about that?

> I don't think there's a need to call merge specifically in the msbuild tasks. A much simpler implementation that comes to mind will be to simply pass the supplied file in the Merge property to the Coverage class, deserialize it and perform the merge within that class. Merge functionality should be a function of the core library, msbuild tasks should be limited entirely to presentation of results.

Agreed. This was, however, the easiest for me to implement :-)

> From my knowledge of how merging coverage results work I should expect that the situation above should only be true if each of the two projects cover a different 50% of the assembly under test. Is that the case with your implementation?

[Here](https://github.com/basilfx/Coverlet-Test/blob/master/src/CoverletTest/Application.cs) is the code. One test project covers the if-part, the other the else-part.

ErikSchierboom:
Great to see this is being worked on!

tonerdo:
> I fully agree, but the point is that Coverlet support multiple output formats. I think it is therefore more confusing for the end-user that it doesn't work whenever JSON format is not selected, so have you any idea's about that?

Just like OpenCover which has it's primary format, Coverlet also has it's own JSON format. The support for other formats is entirely for compatibility reasons and to make it easy to integrate Coverlet with existing tools.

> Here is the code. One test project covers the if-part, the other the else-part.

Ah ok. That's fine then

> No, the BranchInfo object wasn't directly usable. As far as I understand, every branch is uniquely identified by a combination of values. Therefore you have a mapping of (Number, Offset, EndOffset, Path, Ordinal) -> Hits. In the old situation you had (Number) -> (Offset, EndOffset, Path, Ordinal, Hits), which is a bit odd (and wrong, IMHO). It also requires more effort when merging the results, because you have to do multiple lookups.

I'm not very sold on the new format you came up with. It feels a bit counter-intuitive to have `Key` and `Value` fields. We need to be sure to not impulsively change the json output format as it'll cause thrid party tools that leverage the output to perform certain functionality to break. I'm trying to avoid introducing constant breaking changes. I'll make some modifications to the output format within the next couple of hours that should not be that much of a drastic change from how it is now. Hopefully, that'll be easier to work with.

basilfx:
> Just like OpenCover which has it's primary format, Coverlet also has it's own JSON format. The support for other formats is entirely for compatibility reasons and to make it easy to integrate Coverlet with existing tools.

Point is that whenever someone doesn't output Coverlet JSON, they cannot merge results. So IMHO it is more confusing to the end-user, as opposed to having a `/p:<insert name here>` that will always write the results to that file.

> I'm not very sold on the new format you came up with. It feels a bit counter-intuitive to have Key and Value fields. We need to be sure to not impulsively change the json output format as it'll cause thrid party tools that leverage the output to perform certain functionality to break. I'm trying to avoid introducing constant breaking changes. I'll make some modifications to the output format within the next couple of hours that should not be that much of a drastic change from how it is now. Hopefully, that'll be easier to work with.

It should be very easy to output the old format by changing the serializer. I just wanted to demonstrate that the `Key` composes of multiple numbers, instead just one. That's something you have to take into account I think.

tonerdo:
Apologies for the edit on your last comment. Didn't realize I was not editing my post, didn't even know I could do that!

tonerdo:
> Point is that whenever someone doesn't output Coverlet JSON, they cannot merge results. So IMHO it is more confusing to the end-user, as opposed to having a /p:<insert name here> that will always write the results to that file.

I totally get your point. If a user defaults to opencover as their format of choice they still have to output and extra file to support merging with results from another project. I'd much prefer that extra file to be the primary json coverage result also I don't want to introduce any new properties if I can help it

risogolo:
Guys when this feature will be merged to master? any schedule?

basilfx:
I currently have no time/intention to work on this. I'll close it for now.

PR #128, opened 2018-06-22
Add exe modules to coverage

### LINKED ISSUE #126: Only .dll references are being covered
When targeting .net framework (i.e. net471), the output type of an asp.net core application is .exe, which is then referenced by tests. Coverlet ignores everything except for .dll references: https://github.com/tonerdo/coverlet/blob/3d86243d272e92b96e04d3eaa5942c0c25d85001/src/coverlet.core/Helpers/InstrumentationHelper.cs#L18

MarcoRossignoli:
closes #126 

/cc @tonerdo

### DIFF
--- a/src/coverlet.core/Helpers/InstrumentationHelper.cs
+++ b/src/coverlet.core/Helpers/InstrumentationHelper.cs
@@ -15,7 +15,7 @@ namespace Coverlet.Core.Helpers
     {
         public static string[] GetCoverableModules(string module)
         {
-            IEnumerable<string> modules = Directory.EnumerateFiles(Path.GetDirectoryName(module)).Where(fileName => Path.HasExtension(fileName) && (Path.GetExtension(fileName) == ".exe" || Path.GetExtension(fileName) == ".dll"));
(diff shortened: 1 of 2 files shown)


src/coverlet.core/Helpers/InstrumentationHelper.cs
```diff
@@ -15,7 +15,7 @@ internal static class InstrumentationHelper
     {
         public static string[] GetCoverableModules(string module)
         {
-            IEnumerable<string> modules = Directory.GetFiles(Path.GetDirectoryName(module), "*.dll");
+            IEnumerable<string> modules = Directory.EnumerateFiles(Path.GetDirectoryName(module)).Where(fileName => Path.HasExtension(fileName) && (Path.GetExtension(fileName) == ".exe" || Path.GetExtension(fileName) == ".dll"));
```
tonerdo: Try this:
```csharp
Directory.EnumerateFiles(Path.GetDirectoryName(module)).Where(f => f.EndsWith(".exe") || f.EndsWith(".dll"))
```
It feels more cleaner

MarcoRossignoli: ok

PR #134, opened 2018-06-26
Performance improvements

### LINKED ISSUE #120: Code Coverage Result Calculation takes longer time
Hi,
For our project using Coverlet 2.0 to get the code coverage results, thank you for such a nice Library. 
Between calculating coverage results and generating the report it takes approximately 6 minutes on average. Our average build time has gone up from 3 minutes to 11+ minutes. Is there any way in reducing the Calculating Code Coverage and Report Generation time. With 1.0 we did not have this problem, the overall time for unit tests and code coverage was not more than 3 minutes.

Recent Build Log output for ref:
2018-06-13T13:36:41.7491920Z Step 4/15 : RUN /bin/bash -c "dotnet --version      dotnet sln ./MyProjectName.sln remove docker-compose.dcproj      dotnet restore ./MyProjectName.sln      dotnet publish ./MyProjectName.sln --configuration Release -o ./obj/Docker/publish      dotnet test /p:CollectCoverage=true /p:CoverletOutputFormat=opencover --configuration Release --no-build ./tests/Services/UnitTests/UnitTests.csproj      if [[ -f "/build/tests/Services/UnitTests/coverage.xml" ]]; then cp -f -T /build/tests/Services/UnitTests/coverage.xml /build/src/Services/MyProject/MyProject.API/obj/Docker/publish/API_UnitTests_coverage.xml; fi"
...
2018-06-13T13:38:18.2159990Z Microsoft (R) Test Execution Command Line Tool Version 15.3.0-preview-20170628-02
2018-06-13T13:38:18.2174940Z Copyright (c) Microsoft Corporation.  All rights reserved.
2018-06-13T13:38:18.2181180Z 
2018-06-13T13:38:18.3384270Z Starting test execution, please wait...
2018-06-13T13:38:21.3001400Z [39;49m[xUnit.net 00:00:02.1583989]   Discovering: UnitTests
2018-06-13T13:38:21.4789480Z [39;49m[39;49m[xUnit.net 00:00:02.3576858]   Discovered:  UnitTests
2018-06-13T13:38:21.4856690Z [39;49m[39;49m[xUnit.net 00:00:02.3651608]   Starting:    UnitTests
2018-06-13T13:40:56.2712730Z [39;49m[39;49m[xUnit.net 00:02:37.1488980]   Finished:    UnitTests
2018-06-13T13:40:56.4501810Z [39;49m
2018-06-13T13:40:56.4518150Z [39;49mTotal tests: 610. Passed: 610. Failed: 0. Skipped: 0.
2018-06-13T13:40:56.4534620Z [39;49m[39;49m[32mTest Run Successful.
2018-06-13T13:40:56.4549900Z [39;49m[39;49mTest execution time: 2.6347 Minutes
2018-06-13T13:40:56.5515730Z [39;49m[?1h=
2018-06-13T**13:40:56**.5528900Z Calculating coverage result...
2018-06-13T**13:46:45**.9403570Z Generating report '/build/tests/Services/UnitTests/coverage.xml'
...
2018-06-13T13:46:46.5420020Z -- Report generation completed

petli:
Fixes #120 see comments in that issue on the performance improvements here.

One possibly breaking change here is that it no longer periodically writes files (as there's no real need for it) but that also means that if the process exit callbacks aren't called for some reason, no results at all will be available.  A periodic flush could be added based on number of events, but since such a partial result is not very useful, I think it would be better to look for ways to ensure that the results always are written correctly.

To test the performance change, just checkout commit b3c080c that just contains the new performance test. The changes are in the second commit.

### DIFF
--- a/src/coverlet.core/Coverage.cs
+++ b/src/coverlet.core/Coverage.cs
@@ -55,10 +55,10 @@ namespace Coverlet.Core
             foreach (var result in _results)
             {
                 Documents documents = new Documents();
-                foreach (var doc in result.Documents)
+                foreach (var doc in result.Documents.Values)
                 {
                     // Construct Line Results
-                    foreach (var line in doc.Lines)
+                    foreach (var line in doc.Lines.Values)
                     {
                         if (documents.TryGetValue(doc.Path, out Classes classes))
                         {
@@ -91,7 +91,7 @@ namespace Coverlet.Core
                     }
 
                     // Construct Branch Results
-                    foreach (var branch in doc.Branches)
+                    foreach (var branch in doc.Branches.Values)
                     {
                         if (documents.TryGetValue(doc.Path, out Classes classes))
                         {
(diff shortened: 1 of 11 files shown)


latop2604:
> partial result is not very useful

It can be useful to see in the IDE where the test failed. (dotnet watch + Coverage Gutter)
Visual Studio live unit test as this feature and it helps a lot

tonerdo:
Oh wow! This improvement is much welcome. Will take a deeper look tomorrow

tonerdo:
Also, if you've got time can you just give me a brief overview of the major bottle necks you found?

src/coverlet.core/Coverage.cs
```diff
@@ -147,55 +147,52 @@ private void CalculateCoverage()
...
+                        }
+                        else
+                        {
+                            int end = int.Parse(info[3]);
+                            for (int j = start; j <= end; j++)
                             {
-                                int end = int.Parse(info[3]);
-                                for (int j = start; j <= end; j++)
-                                {
-                                    var line = document.Lines.First(l => l.Number == j);
```
petli: This line was the core of the bottleneck (together with the similar line doing `document.Branches.First` above.  Having the lines in a list means that it finding the correct line for an event will have to search through `n/2` items of the `n` items in the list on average. The number of events in a test can be described as `kn`, where `k` is the number of times the average line is executed during the test.

Putting that together means that processing all events in the test will take `(k/2) * n^2` operations. In complexity theory the factor is not interesting since when `n` gets big, the value of `k` is irrelevant, so the time complexity of this operation is expressed as `O(n^2)` in the big-O notation, i.e. quadratic time complexity.

petli:
@tonerdo I've added a comment to the diff highlighting the lines causing the main bottleneck.

The PR addresses that by using a dictionary to tally the event counts. According to 
https://msdn.microsoft.com/en-us/library/xfhwa508.aspx#Remarks the lookup time complexity of a dictionary lookup is close to `O(1)`, i.e. constant time, but adding an element to the dictionary is worst-case `O(n)` if the dictionary must be extended.  An minor optimisation here might be to initilialize the dictionary to be large enough to fit all events, but I reckon there's already a lot of optimisation going on under the hood in Dictionary to make that a not very huge win.

The second part of this PR moves this tallying of events from the post-processing step into `CoverageTracker` directly. This doesn't change anything in terms of time complexity, but saves time since a lot of events doesn't have to be written to disk and later processed.

Further optimisation possibilities worth investigating could be:

* Instead of using strings to identify the event, increment a counter on each instrumentation and use that number to identify the event.  This saves on hashing and string comparison, reduces memory consumption and should speed up test execution another bit. The post-processing step would translate these event IDs back into the lines or branches hit.  If the number of events can be made known to `CoverageTracker`, it could also allocate an `int[]` of sufficient size and avoid the dictionary lookups completely.  

* Continuing on that and tying into the comment by @latop2604, instead of using a .NET array it might be possible to use a memory-mapped file. This means that hit counts are updated directly on disk without having to write them to a file at the end of it, and would ensure that regardless of how a test ends the hit counts are stored in a file. It would even be possible to produce reports continuously while the test runs, if that would be interesting.

test/coverlet.core.performancetest/coverlet.core.performancetest.csproj
```diff
@@ -0,0 +1,19 @@
...
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="xunit" Version="2.3.1" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.3.1" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\coverlet.testsubject\coverlet.testsubject.csproj" />
```
tonerdo: I think we can move `BigClass.cs` directly into this project.

src/coverlet.tracker/CoverageTracker.cs
```diff
@@ -2,70 +2,61 @@
...
-                    _markerFileCount[path] = _markerFileCount[path] + 1;
+                    fileEvents = new Dictionary<string, int>();
+                    _events.Add(file, fileEvents);
+                }
+
+                if (!fileEvents.TryGetValue(evt, out var count))
+                {
+                    fileEvents.Add(evt, 1);
+                }
+                else if (count < int.MaxValue)
```
tonerdo: What happens if `count` is less than `int.MaxValue`? Shouldn't we try a long?

test/coverlet.core.performancetest/coverlet.core.performancetest.csproj
petli: I might have misunderstood something then, as I thought only the dependencies of the unit tests got instrumented, not the unit test DLL itself. Is that not the case? The purpose of having this test subject DLL was to get BigClass instrumented when running the performance test.

tonerdo: My mistake @petli. Thanks for the clarification

petli: The class do need a comment block explaining why it is where it is, I'll fix that.

PR #141, opened 2018-07-06
Tracking Perf enhancement in multi-threading scenario

tarekgh:
The tracking was taking a glob lock which is a bottleneck for the multi-threaded scenario. This is a simple change avoid the global lock which dramatically improves the performance in such scenarios.

I tested the perf by running coverlet with https://github.com/dotnet/corefx/tree/master/src/System.Collections/tests (after hacking some stuff to make it work with such test). I am seeing the perf improved 3x in this scenario. the test used to take more than 15 minutes to run and now it takes less than 5 minutes. we may look at more improvement if needed.

### DIFF
--- a/src/coverlet.tracker/CoverageTracker.cs
+++ b/src/coverlet.tracker/CoverageTracker.cs
@@ -7,12 +7,15 @@ namespace Coverlet.Tracker
 {
     public static class CoverageTracker
     {
-        private static Dictionary<string, Dictionary<string, int>> _events;
+        private static List<Dictionary<string, Dictionary<string, int>>> _events;
+
+        [ThreadStatic]
+        private static Dictionary<string, Dictionary<string, int>> t_events;
 
         [ExcludeFromCodeCoverage]
         static CoverageTracker()
         {
-            _events = new Dictionary<string, Dictionary<string, int>>();
+            _events = new List<Dictionary<string, Dictionary<string, int>>>();
             AppDomain.CurrentDomain.ProcessExit += new EventHandler(CurrentDomain_ProcessExit);
             AppDomain.CurrentDomain.DomainUnload += new EventHandler(CurrentDomain_ProcessExit);
         }
@@ -20,12 +23,23 @@ namespace Coverlet.Tracker
         [ExcludeFromCodeCoverage]
         public static void MarkExecuted(string file, string evt)
         {
-            lock (_events)
+            if (t_events == null)
             {
-                if (!_events.TryGetValue(file, out var fileEvents))
+                t_events = new Dictionary<string, Dictionary<string, int>>();
+                lock (_events)
+                {
+                    _events.Add(t_events);
+                }
+            }
+
+            // We are taking the lock only for synchronizing with CurrentDomain_ProcessExit event
+            // but the lock should be fast as the thread shouldn't block before getting process exit event
+            lock (t_events)
+            {
+                if (!t_events.TryGetValue(file, out var fileEvents))
                 {
                     fileEvents = new Dictionary<string, int>();
-                    _events.Add(file, fileEvents);
+                    t_events.Add(file, fileEvents);
                 }
 
                 if (!fileEvents.TryGetValue(evt, out var count))
@@ -42,16 +56,30 @@ namespace Coverlet.Tracker
         [ExcludeFromCodeCoverage]
         public static void CurrentDomain_ProcessExit(object sender, EventArgs e)
         {
+            Dictionary<string, bool> encounteredFile = new Dictionary<string, bool>();
+
             lock (_events)
             {
-                foreach (var files in _events)
+                foreach (var localEvents in _events)
                 {
-                    using (var fs = new FileStream(files.Key, FileMode.Create))
-                    using (var sw = new StreamWriter(fs))
+                    lock (localEvents)
                     {
-                        foreach (var evt in files.Value)
+                        foreach (var files in localEvents)
                         {
-                            sw.WriteLine($"{evt.Key},{evt.Value}");
+                            FileMode mode = FileMode.Open;
+                            if (!encounteredFile.TryGetValue(files.Key, out bool b))
+                            {
+                                encounteredFile.Add(files.Key, true);
+                                mode = FileMode.Create;
+                            }
+                            using (var fs = new FileStream(files.Key, mode))
+                            using (var sw = new StreamWriter(fs))
+                            {
+                                foreach (var evt in files.Value)
+                                {
+                                    sw.WriteLine($"{evt.Key},{evt.Value}");
+                                }
+                            }
                         }
                     }
                 }
--- a/src/coverlet.tracker/CoverageTracker.cs
+++ b/src/coverlet.tracker/CoverageTracker.cs
@@ -66,7 +66,7 @@ namespace Coverlet.Tracker
                     {
                         foreach (var files in localEvents)
                         {
-                            FileMode mode = FileMode.Open;
+                            FileMode mode = FileMode.Append;
                             if (!encounteredFile.TryGetValue(files.Key, out bool b))
                             {
                                 encounteredFile.Add(files.Key, true);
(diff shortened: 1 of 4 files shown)


tarekgh:
@tonerdo let me know what you think about this change. Thanks.

CC @pjanotti

petli:
Looks like a good idea, @tarekgh 

You need to make a change here too: https://github.com/tonerdo/coverlet/blob/master/src/coverlet.core/Coverage.cs#L182-L196

When I rewrote hit counting to happen in the CovereTracker, I had the assumption that there's only one row for each event which won't be true with this change. So you'll need to change lines 186 and 194 to add into `branch.Hits` and `line.Hits`.

tarekgh:
Thanks @petli, I'll try to fix this soon.

tarekgh:
@petli looking at https://github.com/tonerdo/coverlet/blob/master/src/coverlet.core/Coverage.cs#L182-L196 it looks my changes here shouldn't affect this part. my changes are doing the optimization before writing the data to the files. my changes don't change the end result written to the files. So, calculating the coverage should work as it used to work. please correct me if I missed anything.

petli:
@tarekgh As I read the change to have one event-counting dictionary per thread and then writing these dictionaries to files at the end of the test, it means that if different threads hit the same line, that line event will be duplicated in the file.  And hence they must be added together in the end.

You could extend the performance test to call BigClass from several threads, to show that it both counts line hits correctly and that it is a performance improvement.

tarekgh:
@petli looking at the CalculateCoverage I thought the duplication of the visited lines shouldn't matter as we'll just override the old same value in ```line.Hits = hits;``` and ``` branch.Hits = hits;```. I may still be missing something but I am trying to understand how everything works. if you still think we shouldn't allow the duplication, I can try to fix that before writing the values to the file in CurrentDomain_ProcessExit or fix it during the CalculateCoverage.

```C#
                        if (!result.Documents.TryGetValue(info[1], out var document))
                        {
                            continue;
                        }

                        int start = int.Parse(info[2]);
                        int hits = int.Parse(info[4]);

                        if (isBranch)
                        {
                            int ordinal = int.Parse(info[3]);
                            var branch = document.Branches[(start, ordinal)];
                            branch.Hits = hits;
                        }
                        else
                        {
                            int end = int.Parse(info[3]);
                            for (int j = start; j <= end; j++)
                            {
                                var line = document.Lines[j];
                                line.Hits = hits;
                            }
                        }
```

src/coverlet.tracker/CoverageTracker.cs
```diff
@@ -42,16 +56,30 @@ public static void MarkExecuted(string file, string evt)
...
                 {
-                    using (var fs = new FileStream(files.Key, FileMode.Create))
-                    using (var sw = new StreamWriter(fs))
+                    lock (localEvents)
                     {
-                        foreach (var evt in files.Value)
+                        foreach (var files in localEvents)
                         {
-                            sw.WriteLine($"{evt.Key},{evt.Value}");
+                            FileMode mode = FileMode.Open;
```
petli: Another thought: will `FileMode.Open` position the file for writing at the end? I can't quite make it out from the documenation at https://docs.microsoft.com/en-us/dotnet/api/system.io.filemode?view=netcore-2.1

If not, this will open the file for writing at the start and overwrite what was written the last time.  It could be better to use `FileMode.Append` instead (see same page above) to ensure that data is always added to the file.

An alternative is to keep the open file streams in `encounteredFile`, but at the price of not eing able to use `using` and the streams must be disposed manually at the end.

petli:
You could investigate this behaviour, together with the one about the file mode, by modifying https://github.com/tonerdo/coverlet/blob/master/test/coverlet.core.performancetest/PerformanceTest.cs to be threaded and see if you get each thread counted in the final report.

src/coverlet.tracker/CoverageTracker.cs
tarekgh: I think you are right here. I'll open the file with the AppendMode.

tarekgh:
@petli I have tested coverlet.core.performancetest and I am seeing it is producing the right count now and everything looks good to me. please have a look and let me know if you have any more comments. Thanks.

tonerdo:
Thanks for the look over @petli, your insight into performance optimizations is much appreciated. @tarekgh thanks for the contribution, will also take a look at the changes. Save for @petli's, nothing seems to jump out at me at the moment so it should most likely get through quickly

> after hacking some stuff to make it work with such test

I'm curious what hacks you had to make to get coverlet to work. Do you mind sharing? I'm trying to ensure coverlet is suitable for as many scenarios as possible

petli:
Looks good to me too, @tonerdo @tarekgh

tonerdo:
@tarekgh thanks for sharing, I do have some thoughts on some of the coverlet specific things you mentioned

* > As coverlet working as msbuild tasks is good in term of having msbuild cache the jitting so it doesn't have to re-jit with every test in our complete code coverage runs which run all tests. although this is good for the performance, it is kind of a little bit complicated to integrate with other build systems like corefx which has some specific way to build and run tests. did you consider providing standalone coverlet tool too? I think this will be very helpful to give the choice for users to either use msbuild targets or use the stand-alone tool.

    PR #143 already has this functionality which distributes coverlet as a xplat .NET Core global tool. The functionality is all there and ready to go. I just need to figure out the best way to update the documentation so that new users don't get confused between the two modes. Will look into setting up a nightly feed so you can have access to the bits and test out early

* > coverlet has Exclude msbuild property which can be used to configure the excludes but corefx work the other way which we have only the list we need to include. we already worked around that by manually filtering the list. I believe this is something can easily be supported in coverlet.

    Adding an `Include` option is a no brainer and that change is quite simple. I'll go ahead and get that in, but one will have to take precedence. Considering the most likely scenario is that the developer will want to exclude, `Include` will be ignored if `Exclude` is also specified.

* > I guess coverlet have packaging issues so I think we'll contact you later regarding that. In short, coverlet built as netstandard 2.0 and the package include the dependent assemblies. it needs to be tested when running it on the full framework scenarios....

    Correct me if I'm wrong but I feel this will no longer be an issue when you switch to the standalone tool.

I appreciate you battle testing coverlet and helping improve it. I'm quite familiar with the `core*` repos - being a contributor myself - and I'm willing to help out with integration on that end.

tarekgh:
> Adding an Include option is a no brainer and that change is quite simple. I'll go ahead and get that in, but one will have to take precedence. Considering the most likely scenario is that the developer will want to exclude, Include will be ignored if Exclude is also specified.

Maybe the best is to allow Include and Exclude work together. I mean, let's coverlet work as it works today with just accommodating Include list if it exists. And allow the option to Exclude everything, something like ```[*]*``` so in our corefx scenario we can use the exclude to exclude everything and use Include to include the list we have. 

> Correct me if I'm wrong but I feel this will no longer be an issue when you switch to the standalone tool.

This will depends how you implement the tools. will you make it as net core or full desktop tool? Also, if you are still supporting the msbuild mode, I guess you'll still need to fix the package for this scenario too.

> I appreciate you battle testing coverlet and helping improve it. I'm quite familiar with the core* repos - being a contributor myself - and I'm willing to help out with integration on that end.

This is great and we appreciate your willing and your work you are doing.

tarekgh:
@tonerdo  I assume you'll be the one can merge the changes as I don't have permission to merge it myself. right?

tonerdo:
* > Maybe the best is to allow Include and Exclude work together. I mean, let's coverlet work as it works today with just accommodating Include list if it exists. And allow the option to Exclude everything, something like [*]* so in our corefx scenario we can use the exclude to exclude everything and use Include to include the list we have.

    Probably isn't necessary to go that far. If just `Include` is specified then everything is excluded by default and only what you specify is included. Meh, I'll try not to overthink it but I'll do my best to make it quite simple.

* > If you are still supporting the msbuild mode, I guess you'll still need to fix the package for this scenario too.

    The global tool will be in a different package, but just reach out if you run into any issues either way.

tonerdo:
> I assume you'll be the one can merge the changes as I don't have permission to merge it myself. right?

Yup, merging now

PR #155, opened 2018-07-17
Improve performance and memory usage during report file generation

alexanderkozlenko:
The PR improves performance and memory usage during report file generation by the following changes:

1. Remove additional allocation of  `MemoryStream`, `List<T>`, `byte[]`, and `string` objects.
2. Change reporters to write content directly to a `StreamWriter` instance.

### DIFF
--- a/src/coverlet.console/Program.cs
+++ b/src/coverlet.console/Program.cs
@@ -77,7 +77,9 @@ namespace Coverlet.Console
 
                     var report = Path.Combine(directory, filename);
                     logger.LogInformation($"  Generating report '{report}'");
-                    File.WriteAllText(report, reporter.Report(result));
+                    using (var fileStream = new FileStream(report, FileMode.Create, FileAccess.Write, FileShare.Read))
+                    using (var streamWriter = new StreamWriter(fileStream, Encoding.UTF8))
+                        reporter.Report(result, streamWriter);
                 }
 
                 var summary = new CoverageSummary();
--- a/src/coverlet.console/Program.cs
+++ b/src/coverlet.console/Program.cs
@@ -77,8 +77,7 @@ namespace Coverlet.Console
 
                     var report = Path.Combine(directory, filename);
                     logger.LogInformation($"  Generating report '{report}'");
-                    using (var fileStream = new FileStream(report, FileMode.Create, FileAccess.Write, FileShare.Read))
-                    using (var streamWriter = new StreamWriter(fileStream, Encoding.UTF8))
+                    using (var streamWriter = File.CreateText(report))
                         reporter.Report(result, streamWriter);
                 }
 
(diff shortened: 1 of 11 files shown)


src/coverlet.console/Program.cs
```diff
@@ -77,7 +77,9 @@ static int Main(string[] args)
 
                     var report = Path.Combine(directory, filename);
                     logger.LogInformation($"  Generating report '{report}'");
-                    File.WriteAllText(report, reporter.Report(result));
+                    using (var fileStream = new FileStream(report, FileMode.Create, FileAccess.Write, FileShare.Read))
```
latop2604: Why not just `var streamWriter = File.CreateText(...)` ?

joopscheer:
This PR is missing Codecov results. Any idea why?

alexanderkozlenko:
@joopscheer, I'm not sure that this is what you are asking about, however, this is code coverage report from AppVeyor for the PR:

```
  +------------------+--------+--------+--------+
  | Module           | Line   | Branch | Method |
  +------------------+--------+--------+--------+
  | coverlet.core    | 91.4%  | 86.3%  | 93.3%  |
  +------------------+--------+--------+--------+
  | coverlet.tracker | 0%     | 0%     | 0%     |
  +------------------+--------+--------+--------+
```
and this is the latest code coverage result for the master branch:
```
  +------------------+--------+--------+--------+
  | Module           | Line   | Branch | Method |
  +------------------+--------+--------+--------+
  | coverlet.core    | 91.4%  | 86.3%  | 93.2%  |
  +------------------+--------+--------+--------+
  | coverlet.tracker | 0%     | 0%     | 0%     |
  +------------------+--------+--------+--------+
```

joopscheer:
@alexanderkozlenko Never mind. Some PR's have four checks. This PR just one. Not sure why. But that's why there isn't a Codecov report comment.

tonerdo:
Hi @alexanderkozlenko, thanks for the contribution. Can you include some stats on the performance improvement. You can use `coverlet.core.performancetest` as a baseline.

tonerdo:
@alexanderkozlenko, that's one way to go. Specifically the `coverlet.tests.performancetest` project

alexanderkozlenko:
Unfortunately, I cannot provide with meaningful performance testing results due to the following reasons:

1. The main change is reducing allocations, that is not reflected in test runner output.
2. The accuracy of the time results is not so good to be considered as a performance benchmark.

I assume the only appropriate way to measure performance changes is to use some performance benchmark framework (like [BenchmarkDotNet](https://benchmarkdotnet.org)). It seems it also requires some additional changes to the source code to allow easy mocking of coverage results.

tonerdo:
Hmmm... perhaps @petli can advise in this situation

petli:
The `coverlet.tests.performancetest` project as it is now is not suitable for testing the report generation, since there a lot of classes, methods and files are necessary. That performance test is only useful for testing the instrumentation and event collection. 

Another test could be added to it that mocks up a big coverage result and then measures the report generation time, like @alexanderkozlenko suggests.

tonerdo:
Hi @alexanderkozlenko, any particular reason you closed this and your other PRs?

Issue #122, opened 2018-06-18
Powershell with /p:ExcludeByFile do not works


hack2root:
Powershell script
By adding command `/p:ExcludeByFile=\"**/MyFun*.cs,**/*Test*.cs,**/*Exception*.cs,**/*Attribute*.cs\"` to 

`& dotnet test --configuration Release --no-build Build.Tests /p:CollectCoverage=true /p:CoverletOutputFormat=opencover`

I manage to break generation of the report file, if I removes `/p:ExcludeByFile=\"...\"` everything works
I tried with no success:
```

/p:ExcludeByFile=\"**/MyFun*.cs,**/*Test*.cs,**/*Exception*.cs,**/*Attribute*.cs\"
/p:ExcludeByFile="**/MyFun*.cs,**/*Test*.cs,**/*Exception*.cs,**/*Attribute*.cs"
/p:ExcludeByFile=\"**\/MyFun*.cs,**\/*Test*.cs,**\/*Exception*.cs,**\/*Attribute*.cs\"
/p:ExcludeByFile=\"**\MyFun*.cs,**\*Test*.cs,**\*Exception*.cs,**\*Attribute*.cs\"
/p:ExcludeByFile="**\MyFun*.cs,**\*Test*.cs,**\*Exception*.cs,**\*Attribute*.cs"
```

`/p:ExcludeByFile` just not works

https://ci.appveyor.com/project/hack2root/build/build/1.0.0.222-master
https://ci.appveyor.com/project/hack2root/build/build/1.0.0.223-master

Why not to use SonarScanner format 

`/d:sonar.coverage.exclusions="**/MyFun*.cs,**/*Test*.cs,**/*Exception*.cs,**/*Attribute*.cs" `

for `/p:ExcludeByFile`, for example:

`/p:ExcludeByFile="**/MyFun*.cs,**/*Test*.cs,**/*Exception*.cs,**/*Attribute*.cs"`

My build script:

```
[...]

PR #157, opened 2018-07-18
Explorative PR to solve bug #122[NO MERGE]

MarcoRossignoli:
This is an explorative PR to understand how to fix bug reported by @hack2root on https://github.com/tonerdo/coverlet/issues/122

After some debugging i think the issue is on `InstrumentationHelper.GetExcludedFiles` implementation.
`InstrumentationHelper.GetExcludedFiles` load a list of cs files to exclude from coverage, but this files are searched only starting from [msbuild starting path](https://github.com/tonerdo/coverlet/blob/master/src/coverlet.core/Helpers/InstrumentationHelper.cs#L199)(` Directory.GetCurrentDirectory()`) for "relative" filters.
I'm able to reproduce issue with simple test like(from CoverletTest folder):
```
dotnet test --configuration Release "...\CoverletTest\CoverletTest.csproj"  /p:CollectCoverage=true /p:CoverletOutputFormat=opencover  /v:m  /p:ExcludeByFile=\"**/*Attribute*.cs\,**/*.cs"
```
and with this directory structure
```
CoverletSampleLib
CoverletTest
bugTest.sln
```

So my idea is to test exclusion directly for every source file processed(maybe could perf better for big projects with lot of source files).
I'm not sure if this behaviour was expected or not, let me know if it makes sense and i can go on with my proposal and cleanup code and add some tests.

/cc @tonerdo  @ido-namely

### DIFF
--- a/src/coverlet.core/Coverage.cs
+++ b/src/coverlet.core/Coverage.cs
@@ -34,7 +34,8 @@ namespace Coverlet.Core
         public void PrepareModules()
         {
             string[] modules = InstrumentationHelper.GetCoverableModules(_module);
-            string[] excludes =  InstrumentationHelper.GetExcludedFiles(_excludes);
+            //var ef = InstrumentationHelper.GetExcludedFiles(_excludes);
+            var excludedFilesHelper = new ExcludedFilesHelper(_excludes);
             _filters = _filters?.Where(f => InstrumentationHelper.IsValidFilterExpression(f)).ToArray();
 
             foreach (var module in modules)
@@ -42,7 +43,7 @@ namespace Coverlet.Core
                 if (InstrumentationHelper.IsModuleExcluded(module, _filters))
                     continue;
 
-                var instrumenter = new Instrumenter(module, _identifier, _filters, excludes);
+                var instrumenter = new Instrumenter(module, _identifier, _filters, excludedFilesHelper);
                 if (instrumenter.CanInstrument())
                 {
                     InstrumentationHelper.BackupOriginalModule(module, _identifier);
--- a/src/coverlet.core/Helpers/InstrumentationHelper.cs
+++ b/src/coverlet.core/Helpers/InstrumentationHelper.cs
@@ -246,5 +246,50 @@ namespace Coverlet.Core.Helpers
             }
         }
     }
+
+    internal class ExcludedFilesHelper
+    {
+        const string RELATIVE_KEY = nameof(RELATIVE_KEY);
+        Dictionary<string, Matcher> _matcherDict;
+        public ExcludedFilesHelper(string[] excludes)
+        {
+            if (excludes != null && excludes.Length > 0)
+            {
+                _matcherDict = new Dictionary<string, Matcher>() { { RELATIVE_KEY, new Matcher() } };
+                foreach (var excludeRule in excludes)
+                {
+                    if (Path.IsPathRooted(excludeRule))
+                    {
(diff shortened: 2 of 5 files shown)


src/coverlet.msbuild.tasks/InstrumentationTask.cs
```diff
@@ -40,6 +40,7 @@ public override bool Execute()
         {
             try
             {
+                //System.Diagnostics.Debugger.Launch();
```
MarcoRossignoli: OOT @tonerdo  do you have other way to debug tasks?
We could have some "guide" to explain how to "debug" real msbuild scenario. I think your project will have more contributors, there are few guide on line and maybe a step-by-step guide could be useful, i can help if you think is useful, or build phase could create a simple solution on bin folder ready for debugging.

joaosilvareed:
I'm having the same issue, is it possible to commit this fix?

tonerdo:
Gonna take some time out this weekend to review this and your other PR. I've been quite busy with my day job lately and haven't really been available here

MarcoRossignoli:
@tonerdo don't worry! This is only experimental(i'll need to resolve conflict)...let me know if i can be more useful than "ping" :). I saw Ms guys on coverlet  i'm very happy about it!  Congratulations!

src/coverlet.core/Coverage.cs
```diff
@@ -34,15 +34,16 @@ public Coverage(string module, string[] filters, string[] excludes)
...
+            var excludedFilesHelper = new ExcludedFilesHelper(_excludes);
             _filters = _filters?.Where(f => InstrumentationHelper.IsValidFilterExpression(f)).ToArray();
 
             foreach (var module in modules)
             {
                 if (InstrumentationHelper.IsModuleExcluded(module, _filters))
                     continue;
 
-                var instrumenter = new Instrumenter(module, _identifier, _filters, excludes);
+                var instrumenter = new Instrumenter(module, _identifier, _filters, excludedFilesHelper);
```
tonerdo: I would prefer that `_excludes` be passed in here as is and leave it to the `Instrumenter` class to instantiate `ExcludedFilesHelper`

src/coverlet.core/Helpers/InstrumentationHelper.cs
```diff
@@ -246,5 +246,50 @@ private static bool IsAssembly(string filePath)
             }
         }
     }
+
+    internal class ExcludedFilesHelper
+    {
+        const string RELATIVE_KEY = nameof(RELATIVE_KEY);
+        Dictionary<string, Matcher> _matcherDict;
+        public ExcludedFilesHelper(string[] excludes)
```
tonerdo: I'm not familiar with most of the filesystem apis used in this class. Will play around with them so I can better review this portion

src/coverlet.core/Coverage.cs
MarcoRossignoli: ok when you'll finish review i'll do all amendments, thank you for looking into this.

MarcoRossignoli:
@tonerdo could I close this PR or do you think is useful again?

tonerdo:
@MarcoRossignoli let's keep it open. It's a good reference point and maybe part of it will be incorporated in the future

yowl:
I'm a bit confused with the state of this.  Does this also affect coverlet.exe and is the workaround to use absolute paths?

MarcoRossignoli:
Close because I'll open new one updated.

PR #164, opened 2018-07-27
Add include option

### LINKED ISSUE #163: Add Include option
Per discussion https://github.com/tonerdo/coverlet/pull/141#issuecomment-403436626 

> Adding an Include option is a no brainer and that change is quite simple. I'll go ahead and get that in, but one will have to take precedence. Considering the most likely scenario is that the developer will want to exclude, Include will be ignored if Exclude is also specified.

pjanotti:
Fixes #163. 

Assuming that users typically won't be using both, if they do exclude filter takes precedence. Notice the defaults:

* Default exclude: nothing is excluded
* Default include: everything is included

I felt tempted to avoid some allocations on the filters but opted to keep the code clear as it is, e.g.: I could have avoid some string allocations by batching the creation of substrings etc. Perhaps we can look at this later as separate optimization.

For the tests the include filters are almost symmetrical to the exclude ones so I put some of them under the same test method.

### DIFF
--- a/src/coverlet.console/Program.cs
+++ b/src/coverlet.console/Program.cs
@@ -31,8 +31,9 @@ namespace Coverlet.Console
             CommandOption formats = app.Option("-f|--format", "Format of the generated coverage report.", CommandOptionType.MultipleValue);
             CommandOption threshold = app.Option("--threshold", "Exits with error if the coverage % is below value.", CommandOptionType.SingleValue);
             CommandOption thresholdTypes = app.Option("--threshold-type", "Coverage type to apply the threshold to.", CommandOptionType.MultipleValue);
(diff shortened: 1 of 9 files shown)


tonerdo:
Thanks for the contribution @pjanotti, will take a look later today

> I felt tempted to avoid some allocations on the filters but opted to keep the code clear as it is, e.g.: I could have avoid some string allocations by batching the creation of substrings etc. Perhaps we can look at this later as separate optimization.

I'm definitely open to whatever perf optimizations you have in mind

pjanotti:
Thanks @tonerdo. The bit about the optimization is a bit "premature" in the sense that I see that we could avoid some allocations but that shouldn't be the issue with better ROI. Hopefully I will have some time to find better targets by collecting some performance profiles.

PR #172, opened 2018-08-13
Performance improvement during run of instrumented code

pjanotti:
This PR has a set of small set of low risk optimizations. Coverlet can do much better than this PR by using indexes when calling the tracker code instead of relying on strings and dictionaries, however, that is a much more risky change and needs to be carefully tested and validated. I'm putting this PR up so we can ensure that on next release at least these performance gains are in place, hopefully, me or somebody has a chance to go for this more complete optimization.

The performance effects were measured against the test project System.Text.Encodings.Web.Tests in dotnet/corefx repo. This project was selected for optimization because it takes around 6 seconds to run it without instrumentation but was taking almost 300 seconds using coverlet 2.1.1. Recent changes to fix BadImageFormatException brought this down already to ~170 seconds in release 2.2.1. This PR brings it down a bit further to down to ~115 seconds.  

| Version | RunTestsForProject (ms) | % reduction from previous | From 2.1.1 |
| ------- | -----------------------:| -------------------------:| ----------:|
| 2.1.1 | 295,168 | N/A | N/A |
| 2.2.1 | 168,524 | 42.91% | 42.91% |
| This PR | 113,427 | 32.69% | 61.57% |

The main performance gain comes from a custom comparer for the strings used in the code to track execution. Since these strings typically differ only on the end (e.g.: prefix for temp folder of hits file is always the same), using a comparer that generates hashes based only on the final chars of the strings saves considerable amount of work.

Some of the changes affect `InstrumentationTask` and `CoverageResultTask`. However, for the targeted project the times of `InstrumentationTask` (~1,250 msec) and `CoverageResultTask` (less than 200 msec) were too small in relation to the test execution time and as such I didn't focus on improving them, just ensuring no regression on them. The changes regarding these two tasks were expected to have effect only if a large number of source or hit files were being used by the targeted project.

### DIFF
--- a/src/coverlet.core/Instrumentation/Instrumenter.cs
+++ b/src/coverlet.core/Instrumentation/Instrumenter.cs
@@ -175,16 +175,11 @@ namespace Coverlet.Core.Instrumentation
 
         private Instruction AddInstrumentationCode(MethodDefinition method, ILProcessor processor, Instruction instruction, SequencePoint sequencePoint)
         {
-            int documentIndex = 0;
             if (!_result.Documents.TryGetValue(sequencePoint.Document.Url, out var document))
             {
                 document = new Document { Path = sequencePoint.Document.Url };
-                documentIndex = _result.Documents.Count;
+                document.Index = _result.Documents.Count;
                 _result.Documents.Add(document.Path, document);
-            }
-            else
-            {
-                documentIndex = _result.Documents.Keys.ToList().IndexOf(document.Path);
             }
 
             for (int i = sequencePoint.StartLine; i <= sequencePoint.EndLine; i++)
@@ -193,7 +188,7 @@ namespace Coverlet.Core.Instrumentation
                     document.Lines.Add(i, new Line { Number = i, Class = method.DeclaringType.FullName, Method = method.FullName });
             }
 
-            string marker = $"L,{documentIndex},{sequencePoint.StartLine},{sequencePoint.EndLine}";
+            string marker = $"L,{document.Index},{sequencePoint.StartLine},{sequencePoint.EndLine}";
 
             var pathInstr = Instruction.Create(OpCodes.Ldstr, _result.HitsFilePath);
             var markInstr = Instruction.Create(OpCodes.Ldstr, marker);
@@ -208,16 +203,11 @@ namespace Coverlet.Core.Instrumentation
 
         private Instruction AddInstrumentationCode(MethodDefinition method, ILProcessor processor, Instruction instruction, BranchPoint branchPoint)
         {
-            int documentIndex = 0;
             if (!_result.Documents.TryGetValue(branchPoint.Document, out var document))
             {
                 document = new Document { Path = branchPoint.Document };
-                documentIndex = _result.Documents.Count;
+                document.Index = _result.Documents.Count;
                 _result.Documents.Add(document.Path, document);
-            }
-            else
-            {
-                documentIndex = _result.Documents.Keys.ToList().IndexOf(document.Path);
             }
 
             var key = (branchPoint.StartLine, (int)branchPoint.Ordinal);
@@ -235,7 +225,7 @@ namespace Coverlet.Core.Instrumentation
                     }
                 );
 
-            string marker = $"B,{documentIndex},{branchPoint.StartLine},{branchPoint.Ordinal}";
+            string marker = $"B,{document.Index},{branchPoint.StartLine},{branchPoint.Ordinal}";
 
             var pathInstr = Instruction.Create(OpCodes.Ldstr, _result.HitsFilePath);
             var markInstr = Instruction.Create(OpCodes.Ldstr, marker);
--- a/src/coverlet.tracker/CoverageTracker.cs
+++ b/src/coverlet.tracker/CoverageTracker.cs
@@ -8,6 +8,7 @@ namespace Coverlet.Tracker
     public static class CoverageTracker
     {
         private static List<Dictionary<string, Dictionary<string, int>>> _events;
+        private static readonly StringHashSuffixComparer _stringHashSuffixComparer = new StringHashSuffixComparer();
 
         [ThreadStatic]
         private static Dictionary<string, Dictionary<string, int>> t_events;
@@ -25,7 +26,7 @@ namespace Coverlet.Tracker
         {
             if (t_events == null)
             {
-                t_events = new Dictionary<string, Dictionary<string, int>>();
+                t_events = new Dictionary<string, Dictionary<string, int>>(_stringHashSuffixComparer);
                 lock (_events)
                 {
                     _events.Add(t_events);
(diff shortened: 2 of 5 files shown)


src/coverlet.core/Instrumentation/Instrumenter.cs
```diff
@@ -175,25 +175,20 @@ private void InstrumentIL(MethodDefinition method)
 
         private Instruction AddInstrumentationCode(MethodDefinition method, ILProcessor processor, Instruction instruction, SequencePoint sequencePoint)
         {
-            int documentIndex = 0;
             if (!_result.Documents.TryGetValue(sequencePoint.Document.Url, out var document))
             {
                 document = new Document { Path = sequencePoint.Document.Url };
-                documentIndex = _result.Documents.Count;
+                document.Index = _result.Documents.Count;
```
tonerdo: Oh shit, this is so smart 😄

src/coverlet.tracker/CoverageTracker.cs
```diff
@@ -25,7 +26,7 @@ public static void MarkExecuted(string file, string evt)
         {
             if (t_events == null)
             {
-                t_events = new Dictionary<string, Dictionary<string, int>>();
+                t_events = new Dictionary<string, Dictionary<string, int>>(_stringHashSuffixComparer);
```
tonerdo: You know, now that I think of it there's really no reason to store the path to the hits file as a key in a dictionary. By design, the hits file for an assembly remains the same for an entire test session. Also, the `coverage.tracker` assembly is usually loaded by the runtime in different contexts for each assembly that depends on it so there's no chance of multiple instrumented assemblies supplying conflicting information.

I believe this approach was added when we had multiple separate gzipped hits files because some users were experiencing considerable disk usage due to large hit files. That's no longer the case.

In this case we can simply make the hits file a field and assign to it once (if it evaluates to `null`) and have a standalone `Dictionary<string, int>` to store the marker information.

Let me know what you think

tarekgh:
@tonerdo I thought the same when I worked previously on the perf issue. I didn't spend time during then to investigate more but what you have said align what I was thinking too.

src/coverlet.tracker/CoverageTracker.cs
pjanotti: It sounds correct to me. I'll give a try during the day.

pjanotti: This requires some risky changes: the same thread can receive events from different modules. In other words while the hits file is the same for a given assembly the same thread may receive markers from different assemblies. I'm going to try some minimal changes to optimize that but perhaps we should go for the safe and simple optimization in the short run, while we work on a more profound one.

tonerdo: I don't think it's currently possible for the `coverlet.tracker` assembly to receive markers from different assemblies. Here's an illustration of how I believe things are setup in a multiple assembly scenario.

1. `SomeAssembly.Tests.dll` depends on `SomeAssembly1.dll` and `SomeAssembly2.dll` which are both instrument-able
2. `SomeAssembly1.dll` and `SomeAssembly2.dll` are instrumented by coverlet and the `coverlet.tracker.dll` dependency is copied to the same folder the two assemblies occupy.
3. The test project `SomeAssembly.Tests` is run and during that process, methods in `SomeAssembly1.dll` are called which contain instrumentation instructions that call out to `coverlet.tracker.dll`
4. During that same process, instrumented methods from `SomeAssembly2.dll` also call out to `coverlet.tracker.dll`

What happens in this case is that both `SomeAssembly1.dll` and `SomeAssembly2.dll` have `coverlet.tracker.dll` loaded in different contexts. Whatever changes `SomeAssembly1.dll` makes to static variables in its own copy of `coverlet.tracker.dll` doesn't interfere with how `SomeAssembly2.dll` interacts with its own equivalent copy. `SomeAssembly1.dll` and `SomeAssembly2.dll` depend independently on `coverlet.tracker.dll` the same way two different projects can depend on the same version of `Newtonsoft.Json` and still not get in each other's way even if their binaries all end up in the same folder.

See the original naive implementation of the `CoverageTracker` class https://github.com/tonerdo/coverlet/blob/1d92ee66a1bb68629bf31022a300e57c36ce535f/src/coverlet.core/CoverageTracker.cs. It worked perfectly, it just wasn't performant or thread safe

pjanotti: I did a test to be sure: check https://github.com/tonerdo/coverlet/compare/master...pjanotti:test.multithread?expand=1 It does hit the exception when running the performance test. The thread static is the same in a single app domain (that even doesn't really exist on core).

tonerdo: Does this happen only in a multi-threaded scenario?

pjanotti: No, single thread too, you just need different modules on the same thread.

pjanotti: I did an experiment to generate a "tracker type" per module and get rid of the dictionaries, it seems to be going well. It needs more testing to ensure correctness and that it doesn't break some scenarios that are working now. If you want to take a peak before I do some clean-up/polishing/fine tunning take a look [here]https://github.com/tonerdo/coverlet/compare/master...pjanotti:tracker.type.per.module?expand=1(). There is a bit of trade-off and it probably it will use more memory in projects with lower coverage numbers but for the project that I tested here it consistently runs the whole test in ~19 seconds (about 6 times faster than the optimization proposed on this PR).

tonerdo: Took a quick peek. Seems like a good crop of changes. Will merge this in now since I'm fine with it, thanks for the change and explanations

pjanotti: Thanks @tonerdo

tarekgh:
@pjanotti  I have added a comment but I am not sure if you answered it my question. why we introduced StringHashSuffixComparer  and didn't use StringComparer.Ordinal?

pjanotti:
Because the prefix of the strings is very similar and typically long (think paths for sources and modules) and it is only the end of the strings that actually differentiate the hashes. The next optimization is to get rid of them entirely (see the WIP mentioned above).

tarekgh:
The hashing method can produce deceiving results and we may be lucky so far.
strings like:

"a" and "aaaaaaaaa" will produce the same hash code. if we are planning to change this anyway in the near future, then that is ok. otherwise, this may cause unexpected issues and will be hard to investigate.

pjanotti:
> "a" and "aaaaaaaaa" will produce the same hash code.

Hashes in this case will be different. The "known" case of hash collision introduced by this PR is if the strings have the same length and the last 8 chars are the same, of course there could be the normal "statistical collisions". That said if there are a very high number of the collisions the performance will suffer but correctness won't be affected.

PR #181, opened 2018-08-22
Perf: create a tracker per module and use indexes to improve performance

pjanotti:
This change came from the discusson on PR #172. 

This change brings the performance for System.Text.Encondings.Web.Tests in CoreFx from around 113 seconds to around 19 seconds. The coverlet performance test project now runs for 20,000 iterations in about the same time that it previously was running for 200 iterations, now at 200 iterations the time is dominated by instrumentation and report generation.

The idea of the change is that by creating a distinct static tracker type per module the dictionaries used before to track execution can be replaced with an array to represent the hit locations. This array is written to the disk in binary format, only the number of hits per index, avoiding the need to parse to process the hit files. The trade-off is that this should represent a bit more consumption of memory for projects with very low coverage numbers.


pjanotti:
@tonerdo I know that you just merged #172 but I got this PR ready :grinning: 

I will be slow to answer any Q/comments for a few days, so no rush with this one. It will be great if other contributors could spin this PR against their projects just to be sure that there isn't some corner case missed here.

/cc @tarekgh @petli

petli:
Testing this PR against one of the bigger unit test projects in the code I'm working on at the day job the performance improvements are very good:

Test type | Test execution time | Total time to run dotnet test --no-build
----------|--------------------|--------------
Without coverage | 30s | 31s
coverlet 2.2.1   | 133s | 144s
PR changes       | 34s | 37s

Impressive, @pjanotti!

test/coverlet.core.performancetest/PerformanceTest.cs
```diff
@@ -14,8 +14,9 @@ namespace coverlet.core.performancetest
     /// </summary>
     public class PerformanceTest
     {
-        [Theory(Skip = "Only enabled when explicitly testing performance.")]
-        [InlineData(150)]
+        [Theory(/*Skip = "Only enabled when explicitly testing performance."*/)]
```
petli: Should the test be enabled by default now, or remain Skip in the repo code?

src/coverlet.core/Instrumentation/Instrumenter.cs
```diff
@@ -67,20 +70,119 @@ private void InstrumentModule()
                 using (var module = ModuleDefinition.ReadModule(stream, parameters))
                 {
                     var types = module.GetTypes();
+                    AddCustomModuleTrackerToModule(module);
+
                     foreach (TypeDefinition type in types)
                     {
                         var actualType = type.DeclaringType ?? type;
                         if (!actualType.CustomAttributes.Any(IsExcludeAttribute)
+                            && actualType.Namespace != "Coverlet.Core.Instrumentation.Tracker"
```
tonerdo: Why not add an `ExcludeFromCodeCoverage` attribute to all the methods in that class?

pjanotti:
@tonerdo and @petli thanks for reviewing this PR. I will address the comments soon.

src/coverlet.core/Instrumentation/Instrumenter.cs
pjanotti: I see that you already removed it 👍 Thanks, it was not needed as you noticed

test/coverlet.core.performancetest/PerformanceTest.cs
pjanotti: The perf test is not run by default so we don't need the skip in place (avoids us changing the source just to test perf). That said I should have removed the skip instead of commenting it, I'll clean it up on my next change.

PR #195, opened 2018-09-21
Include ObsoleteAttribute in Attributes to exclude fixes #175

### LINKED ISSUE #175: An option to ignore deprecated types/methods
Hi there, amazing project!

I was wondering if you would consider adding an option to ignore types/methods that are marked with `ObsoleteAttribute`. This way the functionality that is no longer maintained and is already discouraged to use wouldn't affect the coverage results.

amweiss:
I can try to add another config property to disable this, but it seems logical to exclude obsolete things from coverage.

### DIFF
--- a/src/coverlet.core/Instrumentation/Instrumenter.cs
+++ b/src/coverlet.core/Instrumentation/Instrumenter.cs
@@ -388,7 +388,9 @@ namespace Coverlet.Core.Instrumentation
                 nameof(ExcludeFromCoverageAttribute),
                 "ExcludeFromCoverage",
                 nameof(ExcludeFromCodeCoverageAttribute),
-                "ExcludeFromCodeCoverage"
+                "ExcludeFromCodeCoverage",
+                nameof(ObsoleteAttribute),
+                "Obsolete"
             };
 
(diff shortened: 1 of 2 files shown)


tonerdo:
@amweiss Thanks for the contribution! The code is fine, I'm just a bit skeptical whether we should automatically exclude [Obsolete] attributes

amweiss:
Would you be comfortable with some kind of option to do it? That would require a bit more added to the logic or a way to push a config option that far down.

tonerdo:
Hmmm...an option just for this seems a bit too much. Lets try this, I'll keep this PR open and if I get a lot of requests for this feature I'll merge it right in

tonerdo:
@amweiss what do you think about #232, it seems like a more blanket solution

amweiss:
Yeah, that would work for me @tonerdo

tonerdo:
Ok. Do you wanna take a stab at it? @amweiss

amweiss:
Yeah @tonerdo, I'll see what I can do. Just need to review some of the style and patterns a bit in the code base to make sure I will be consistent.

Issue #203, opened 2018-09-28
/p:MergeWith does not merge but adds


pape77:
I'm testing using coverlet over a solution that has 3 test projects which i want to combine into a single .opencover.xml file

Since there's currently no direct solution for this. What I tried to do is generate the coverage for one of the test projects, output its result into an "converage.json" file and then apply a MergeWith param in the following dotnet test commands to merge into that same json file and set the output format to opencover

Problem is, that since the first project(ProjectName.Api) includes other two as references (ProjectName.Application, ProjectName.Data), the report is generated as following:
```
Test run for /build/test/ProjectName.Api.Tests/bin/Debug/netcoreapp2.0/ProjectName.Api.Tests.dll(.NETCoreApp,Version=v2.0)

Results File: /results/_8e5df6e97674_2018-09-28_11_10_36.trx

Total tests: 15. Passed: 15. Failed: 0. Skipped: 0.
Test Run Successful.
Test execution time: 2.4645 Seconds

Calculating coverage result...
  Generating report '/results/coverage.json'

+----------------------+--------+--------+--------+
| Module               | Line   | Branch | Method |
+----------------------+--------+--------+--------+
|ProjectName.Api         | 24.4%  | 28.6%  | 33.3%  |
+----------------------+--------+--------+--------+
| ProjectName.Data        | 0%     | 0%     | 0%     |
+----------------------+--------+--------+--------+
|ProjectName.Application | 0%     | 0%     | 0%     |
+----------------------+--------+--------+--------+
```
So it includes two other coverages for the referenced projects with 0% coverage.
This wouldn't be a problem if /p:MergeWith param actually did a proper merge. 

But when I run the next dotnet test (over the ProjectName.Application project) with the MergeWith param pointing to this recently generated coverage.json file, I get this:
```
Test run for /build/test/ProjectName.Application.Tests/bin/Debug/netcoreapp2.0/ProjectName.Application.Tests.dll(.NETCoreApp,Version=v2.0)

Starting test execution, please wait...
Results File: /results/_5d6c3e7d1c85_2018-09-28_11_10_47.trx

Total tests: 4. Passed: 4. Failed: 0. Skipped: 0.
Test Run Successful.
Test execution time: 1.7143 Seconds

Calculating coverage result...
  Generating report '/results/coverage.json'

+----------------------+--------+--------+--------+
| Module               | Line   | Branch | Method |
+----------------------+--------+--------+--------+
[...]

PR #204, opened 2018-10-01
Merges by fileName (dll's name) instead of the complete path.

pape77:
Merges by fileName (dll's name) instead of the complete path. So the same dlls are merged regardless of which test project references
them, when having multiple test projects.

### DIFF
--- a/src/coverlet.core/CoverageResult.cs
+++ b/src/coverlet.core/CoverageResult.cs
@@ -44,7 +44,7 @@ namespace Coverlet.Core
         {
             foreach (var module in modules)
             {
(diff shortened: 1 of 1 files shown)


pape77:
For more details refer to https://github.com/tonerdo/coverlet/issues/203

pedrofmanuel-sky:
Hi, any idea when this gets into a release?

Thanks

pape77:
@pedrofmanuel-sky Check #211 where he said he will test it and then create a release :)

pedrofmanuel-sky:
:) didn't saw that! Great! thanks @pape77

PR #209, opened 2018-10-04
Add System.Private.CoreLib support

pjanotti:
This is a special case needed to make coverlet the official code coverage tool on CoreFx repo (see https://github.com/dotnet/corefx/issues/31281).

The change didn't cause any noticeable performance penalty and basically has a custom "metada importer processor" to deal with System.Private.CoreLib.

### DIFF
--- a/src/coverlet.core/Instrumentation/Instrumenter.cs
+++ b/src/coverlet.core/Instrumentation/Instrumenter.cs
@@ -66,6 +66,10 @@ namespace Coverlet.Core.Instrumentation
             {
                 resolver.AddSearchDirectory(Path.GetDirectoryName(_module));
                 var parameters = new ReaderParameters { ReadSymbols = true, AssemblyResolver = resolver };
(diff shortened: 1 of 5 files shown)


tonerdo:
@pjanotti don't worry about the coverage result right now. I don't enforce a threshold and 94% is still pretty good! Will review this PR over the weekend

ViktorHofer:
@tonerdo have you had time to look into the changes? We really would like to use them in corefx as soon as possible :)

pjanotti:
Good question @tonerdo: the issue is that System.Private.CoreLib can't reference any external assembly and the tracker template is injected, intentionally, by referencing netstandard. This allows it (at least in theory) to instrument assemblies running against any runtime that supports netstandard. All the code used by the tracker template is part of System.Private.CoreLib so what this change does is that it ensures that no actual reference is made via netstandard.

Unfortunately I didn't see a way for this special case.

tonerdo:
Thanks for this @pjanotti, will make a new release tomorrow and notify you once that happens

PR #211, opened 2018-10-09
Fix/merge modules

pape77:
Fixes Merge #204 !

@tonerdo  So after playing a bit more around with my last change (which is already merged #204) , I realized I made a fuck-up because despite i was comparing by filename, you were already storing the keys for the ```Modules``` dictionary as complete paths in ```Coverage.cs```. 

This caused the comparison I modified in #204 to asume the keys for ```modules``` and ```Modules``` were the same. But i didn't change the values of these keys to be only the file name too, so they still had the complete path on it (and therefore, were different) . 
As it tries to access  the dictionary with the wrong key (asuming they were the same in both ```modules``` and ```Modules```) it crashes. 
If you see the different commits maybe it helps to understand what I was aiming to achieve.

The solution was much simpler. Now I just store the module keys with ```Path.GetFileName``` in ```Coverage.cs```, so they can be correctly matched in ```CoverageResult.cs``` as it was done before my change.

I already tested it locally and worked for me for the example on issue #203 

Hope this is clear enough!

Here's the output. Please merge!

```
  Generating report '\results\coverage.opencover.xml'

+----------------------+--------+--------+--------+
| Module               | Line   | Branch | Method |
+----------------------+--------+--------+--------+
| Rtl.ProjectName.Api         | 59.4%  | 66.7%  | 66.7%  |
+----------------------+--------+--------+--------+
| Rtl.ProjectName.Application | 100%   | 60%    | 100%   |
+----------------------+--------+--------+--------+
| Rtl.ProjectName.Data        | 24.8%  | 38.5%  | 54.5%  |
+----------------------+--------+--------+--------+

```

### DIFF
--- a/src/coverlet.core/Coverage.cs
+++ b/src/coverlet.core/Coverage.cs
@@ -145,7 +145,7 @@ namespace Coverlet.Core
                     }
                 }
 
-                modules.Add(result.ModulePath, documents);
+                modules.Add(Path.GetFileName(result.ModulePath), documents);
                 InstrumentationHelper.RestoreOriginalModule(result.ModulePath, _identifier);
             }
 
--- a/src/coverlet.core/CoverageResult.cs
+++ b/src/coverlet.core/CoverageResult.cs
@@ -45,53 +45,52 @@ namespace Coverlet.Core
         {
             foreach (var module in modules)
             {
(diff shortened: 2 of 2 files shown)


pape77:
@tonerdo Great! Please let me know when you do as I really need this!
Feel free to test it of course. I'm pretty confident it will work :)

Issue #192, opened 2018-09-18
Include non-referenced assemblies


jherby2k:
My library uses MEF to load "extensions" that are located in a separate directory. I'd like these assemblies instrumented as well. Is there a way to instrument assemblies that aren't project references?

jherby2k:
Taking a look at the code, it seems to just instrument assemblies in the same directory as the test project, minus exclusions. Would you be open to a pull request that allows one to Include additional assemblies? I’m going to start working on it regardless.

tonerdo:
@jherby2k, I'd be open to that contribution. Thanks!

tswiftma:
Yes I need the MEF capability also! I'm currently using OpenCover but need a solution that runs on linux like Coverlet. I would be happy to help test/compare coverage results.

jherby2k:
@tswiftma I have a nuget package for testing this feature at https://www.myget.org/feed/audioworks/package/nuget/coverlet.msbuild, if you want to try it. Do you keep all your MEF components in a single directory, or in subdirectories?

Use either /p:IncludeDirectories="<extension_path>" or add a "\*" to the end if you want to search subdirectories.

tswiftma:
@jherby2k, @tonerdo,  sorry for the slow response I was flat out at work this week. I had to go back to the code and check, we actually use dependency injection. But your change allowed coverlet to see all my dependencies. I ran coverlet vs open cover (both with the same exclusions) with ReportGenerator and the results were virtually identical:

![coverlet](https://user-images.githubusercontent.com/11710469/47228717-e460c300-d393-11e8-9a62-2f36f67c1137.png)

I will dig into the small differences a bit more but this is a big win that I can run coverage directly with dotnet test on linux!

ViktorHofer:
@tonerdo this issue can now be closed.

PR #213, opened 2018-10-12
Add ability to instrument non-referenced assemblies

jherby2k:
As discussed in #192, I want to be able to instrument assemblies that aren't project references (loaded at runtime). I actually have 2 use cases:

1) Instrumenting a PowerShell core binary module (and its references), which gets loaded via System.Management.Automation.Runspaces.InitialSessionState.ImportPSModule.
2) Instrumenting "extensions" discovered at runtime using MEF (System.Composition.Hosting), and loaded via Assembly.LoadFrom.

To accomplish this, I added an "IncludeDirectories" parameter which allows additional paths to be searched for modules. It supports relative and absolute paths, and a wildcard (*) for including one level of subdirectories (useful for MEF extensions).

One side effect is the possibility that one or more assemblies exists in multiple locations. I dealt with this by including a hash of the file path in each backup file name, and then merging the results for all files with the same name after instrumentation.

Example that includes both a bunch of extension subdirectories and the PowerShell.Commands module directory (both relative paths from the project output directory):

`dotnet test --no-build /p:CollectCoverage=true /p:IncludeDirectories="Extensions\*,PowerShell.Commands"`


jherby2k:
Uhm, those two failing tests seem really spurious - I saw them fail once on my (Windows) desktop too. Not sure it has anything to do with my PR.

jherby2k:
Looks like PR #204 also deals with the merging of files with the same name and different paths.

pape77:
@jhereby2k check #211 i changed it to merge by file name outside the merge method. Adding modules by filename not path anymore in coverage.cs!

jherby2k:
@pape77 That's a nicer solution, so I've reverted the changes I made to CoverageResult.cs in my PR.

pape77:
@jherby2k  nice :) Have you retested with this?

jherby2k:
Of course! It works fine.

jherby2k:
I'm happy to update the readme as well, I just want to confirm this looks good first.

jherby2k:
@pape77 on further testing it does not work fine :(. It merges 2 copies of the same file no problem, but it can't instrument 2 different files with the same name at the same time. So I've reverted to my method, which uses the full path as key and then merges them together at the end. A bit more complicated, but more robust too.

pape77:
@jhereby2k I couldn't quite follow. In the example of my pr i do a dotnet test on a solution that has several test projects.
The scenario there is more than one proj referencing the same dll from different paths. It adds the key by filename the first time and does compare without problems on the following merges with the same filenames bit different paths.
Can you illustrate a bit more when does it not work? Is it just for clacenarios in which you include your extra param to include more files to be merged?

jherby2k:
It’s specific to my PR’s changes. With the IncludeDirectories option, you might end up with two copies of the same file during the same run. Really only one gets loaded into memory, but since you don’t know which will get loaded first, they need to be instrumented separately (so two different keys in the dictionary) and then merged before displaying the results. So my PR always does a merge at the end, and merges any files with the same name.

tonerdo:
@jherby2k kindly fix the merge conflicts so I can review. Thanks!

tonerdo:
Still have some questions regarding merging results but will have a clearer picture once I can make a more wholesome review. Apologies for letting this slide for too long

jherby2k:
@tonerdo No problem! Conflicts resolved. Sorry about all the commits - I had some issues with line endings in Coverage.cs.

tonerdo:
@jherby2k not to worry. I'm trying to wrap my head around your merging logic and trying to figure out if it's indeed needed. Will take a much deeper look after that

jherby2k:
If one of my libraries is a dependency of multiple add-ins/ extensions, it will appear in multiple folders at once, right? And each copy gets instrumented during testing. The merging combines them into a single result. If I don’t merge the result, the dependency gets listed multiple times.

Hope that makes sense.

tonerdo:
@jherby2k just a heads up that this is still on my radar

ViktorHofer:
@jherby2k Please see my PR https://github.com/tonerdo/coverlet/pull/253 which deals with the same issue. Unfortunately I overlooked your PR. If you agree I will incorporate the changes I made in the InstrumentationHelper class into your PR. There's only one problem.

I wanted to avoid having duplicate assembly names at all as that could leave to different issues. My solution is to prioritize the assembly probing:
1. The test assembly's directory.
2. The additional passed in directories as they are passed in.

That allows us in corefx i.e. to use the IL version of System.Private.CoreLib.dll. Means, I need a way to load assembly a from directory A but not load assembly a from directory B.

ViktorHofer:
@jherby2k this feature is fairly time sensitive for us in corefx. It would be great if we could discuss it together with @tonerdo on a channel to get it in as soon as possible. We could either use Gitter, Teams or Skype. Does that work for both of you?

jherby2k:
@ViktorHofer 

> I wanted to avoid having duplicate assembly names at all as that could leave to different issues. My > solution is to prioritize the assembly probing:
>  1. The test assembly's directory.
>  1. The additional passed in directories as they are passed in.

I'm cool with selecting only one in principal, but in practice the above order isn't what I want. I have a binary PowerShell module that depends on and returns types defined in "common.dll". The module exists in its own directory. The test assembly also references "common.dll" because it tests the results returned from that module. But I want the one in the module directory instrumented, not the one in the test directory.

jherby2k:
How about a MergeIdenticalFileNames switch?

ViktorHofer:
Having assemblies with the same assembly and module name isn't support well as it comes with different issues outside of the GAC world. I strongly recommend not going that path...

ViktorHofer:
Opened https://github.com/jherby2k/coverlet/pull/1

jherby2k:
Poorly named option then - not what I intended. How about MergeIdenticalAssemblies? Maybe with an additional check to see if the assemblies really are identical.

ViktorHofer:
> Poorly named option then - not what I intended. How about MergeIdenticalAssemblies? With an additional check to see if the assemblies really are identical.

Can you please explain what identical assemblies are and what they are for? Why do I need to instrument identical assemblies multiple times?

jherby2k:
you don't. As you said, you just need to instrument one of them. But your intended probe order is the opposite of mine. That's why I just instrument them both and then merge the results. Then it doesn't matter which one is actually loaded.

ViktorHofer:
But instrumenting both takes a considerable amount of time and doesn't seem right. If your issue is the priority, why don't we introduce an option to control it?

ViktorHofer:
Sure, let me know if I can help somehow. And sorry to be so pushy to get this feature in but we depend on it in corefx ⚡️

ViktorHofer:
@jherby2k could you accept my PR in the meantime so that @tonerdo can review your PR with my changes? That would be really great. I also noticed that with this work this issue will also be fixed: https://github.com/tonerdo/coverlet/issues/246

jherby2k:
@ViktorHofer Sorry about the unnecessary back and forth - looks like your solution gives the results I want regardless. I've merged your PR and added one extra commit to revert my now-unnecessary backup file naming algorithm.

ViktorHofer:
Your PR seems wrong as it contains all the commits that happened since you opened yours. You probably want to fix that by rebasing or merging or just take my PR where I already did that: https://github.com/tonerdo/coverlet/pull/253

jherby2k:
Blargh, I give up. I'll add my one change to #253.

PR #220, opened 2018-10-18
Merging should not fail if file doesn't exist

vlef:
Changed merging behavior: merges the result of the current test coverage with the specified file only if the file exists.

### DIFF
--- a/src/coverlet.core/Coverage.cs
+++ b/src/coverlet.core/Coverage.cs
@@ -151,7 +151,7 @@ namespace Coverlet.Core
             }
 
             var coverageResult = new CoverageResult { Identifier = _identifier, Modules = modules };
-            if (!string.IsNullOrEmpty(_mergeWith) && !string.IsNullOrWhiteSpace(_mergeWith))
+            if (!string.IsNullOrEmpty(_mergeWith) && !string.IsNullOrWhiteSpace(_mergeWith) && File.Exists(_mergeWith))
             {
                 string json = File.ReadAllText(_mergeWith);
                 coverageResult.Merge(JsonConvert.DeserializeObject<Modules>(json));


vlef:
As discussed on #203 

This tiny change *should* enable the use of a single "dotnet test" command to generate coverage for multiple test projects, aggregate all the results in one file, available in other formats (if requested).

Using VSTS/Azure Devops, this would be the way to have it work properly:

dotnet test \*\*\/\*[Tt]ests/\*.csproj
	/p:CollectCoverage=true
	/p:MergeWith=myfolder/coverage.json
	/p:CoverletOutput=myfolder/
	/p:CoverletOutputFormat=json,opencover,cobertura

Caveats: 

1. When requesting multiple formats in addition to JSON, those other files are generated and overwritten multiples times, for no good reason (other than the execution context not "knowing" which merge is the last one).

2. In my scenario, my build definition depends on coverlet's default naming convention (coverage.json)

Correct me if I'm wrong in my analysis. I'll check to see if I can figure out how to test this, but I probably won't have enough time. If I'm right, this should be a way to really enhance coverlet's usage from a user's standpoint.

pape77:
Nope. I tested it and although it does create a new file if not exists, it creates a new file in each run without merging...

What i run:
```dotnet test .\Rtl.ProjectName.sln -l trx -r /results /p:CollectCoverage=true /p:Cove
rletOutputFormat=opencover /p:CoverletOutput=/results/coverage /p:MergeWith='/results/coverage.json'```

For my example in #203 

 @tonerdo

**Comment by pape77** [comment]
Nope. I tested it and although it does create a new file if not exists, it creates a new file in each run without merging...

What i run:
```
dotnet test .\ProjectName.sln -l trx -r /results /p:CollectCoverage=true /p:Cove
rletOutputFormat=opencover /p:CoverletOutput=/results/coverage /p:MergeWith='/results/coverage.json'
```

For my example in #203 

 @tonerdo  @vlef

**Comment by vlef** [comment]
It should work, if you request both json and your other formats (those you actually want), separated by a coma.

/p:CoverletOutputFormat=json,opencover,cobertura

It's gonna merge with the json file, and also generate/overwrite the other formats in the same folder as the json. You really have to target a folder as the ouput though, can't pick a specific file or it's not gonna work either.

**Comment by pape77** [comment]
@vlef ah good point! Haven't thought about adding both output types as param.
Will try it next time next week. Hope it works.
Thanks!

**Comment by pape77** [comment]
@vlef couldn't resist and I just tested it. Works! :D 
In my case i had to add the ```%2c``` as to be interpreted as a comma, as stated in the project's wiki

```
dotnet test .\ProjectName.sln -l trx -r /results /p:CollectCoverage=true /p:CoverletOutputFormat='json%2copencover' /p:CoverletOutput=/results/coverage /p:MergeWith='/results/coverage.json'
```

And I got the file nicely merged in both formats :D

@tonerdo hope after this merge comes some testing on your side followed by a release! 
As we are quite a few asking for this.

Cheers!

Issue #232, opened 2018-11-01
Add support for exclude by attribute


nordinrahman:
As of now, _coverlet_ could exclude any classes/methods/properties decorated with _ExcludeFromCodeCoverage_ attribute. However, it does not applicable for code annotated by other attributes like _GeneratedCodeAttribute_ nor _CompilerGeneratedAttribute_. In short, it does not allow user to exclude with arbitrary attribute. This is one of most common feature for other coverage tool (e.g. _dotcover_) that I usually use.

There are already demands to this feature, as listed below:

* #194 - CompilerGeneratedAttribute
* #125 - GeneratedCodeAttribute

The feature will have the following criteria:

1. Add additional argument parameter `--exclude-by-attribute` for _CLI_ tool
2. Add additional _MSBUILD_ property `ExcludeByAttribute`
3. Existing implementation to exclude code annotated by `ExcludeByCodeCoverage` remains as is
4. Exclusion list could be arbitrary specified in multiple, delimited by `,`, just like existing `Exclude` list.
5. Any code, i.e.: _classes_, _properties_, and _methods_ annotated with listed attributes, will be ignored from coverage.
6. The behavior of criteria 5 above, will be similar to how existing `ExcludeByCodeCoverage` works.

TQ.

tonerdo:
Hi @nordinrahman, I'm inclined to agree with you on this. Will add it to my personal backlog but I'm open to PRs on this

amweiss:
I can start taking a stab at this if no one else is since it also can cover my desire to see `ObsoleteAttribute` excluded.

nordinrahman:
@amweiss , yes I reviewed the code, and it looks good to me.

Initially, I intend to add more unit test to cover the business logic introduced, but it's kinda of a challenge (eg changes on console app).

I'm yet to run the build against sample project, for manual testing - I will update here once I'm done.

Thanks, @amweiss . I wonder what @tonerdo 's opinion on this?

amweiss:
@nordinrahman I agree, I did the minimum unit test to make sure the attribute excluded the code from coverage. I do have a few stylistic and coverage changes I'll send over on a subsequent PR if this one is getting merged.

amweiss:
Perhaps apparent to other people, but not me, I think it's only going to work as part of a build the way I did it, not in the CLI afterward since it runs at the instrumenter level. So technically it won't meet the needs of this.

amweiss:
[...]

PR #233, opened 2018-11-05
feat: Add ExcludeByAttribute option (#232)

amweiss:
I think this is generally working, but I'm looking for a better way to automate the test for the MSBUILD task and CLI tool to make sure I didn't miss something.

If someone else can pull this and test it for their usage, I'll feel better about it and remove the [WIP] and say it's ready to merge.

Otherwise, there's also a whole bunch of random style and whitespace changes I'd like to make but don't want to mess up this PR for that and will do it in a later one. I tried to keep everything in the same pattern as the existing things, but it's not exactly how I would normally do it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,7 +156,7 @@ You can ignore a method or an entire class from code coverage by creating and ap
 You can also ignore additional attributes by using the `ExcludeByAttribute` property (short name or full name supported):
 
 ```bash
-coverlet <ASSEMBLY> --target <TARGET> --targetargs <TARGETARGS> --exclude-by-attributes "Obsolete,GeneratedCodeAttribute,CompilerGeneratedAttribute"
+coverlet <ASSEMBLY> --target <TARGET> --targetargs <TARGETARGS> --exclude-by-attribute "Obsolete,GeneratedCodeAttribute,CompilerGeneratedAttribute"
 ```
 
 ##### Source Files
--- a/README.md
+++ b/README.md
@@ -153,6 +153,12 @@ coverlet <ASSEMBLY> --target <TARGET> --targetargs <TARGETARGS> --threshold 80 -
 
 You can ignore a method or an entire class from code coverage by creating and applying the `ExcludeFromCodeCoverage` attribute present in the `System.Diagnostics.CodeAnalysis` namespace.
 
+You can also ignore additional attributes by using the `ExcludeByAttribute` property (short name or full name supported):
+
+```bash
+coverlet <ASSEMBLY> --target <TARGET> --targetargs <TARGETARGS> --exclude-by-attributes "Obsolete,GeneratedCodeAttribute,CompilerGeneratedAttribute"
+```
+
 ##### Source Files
 
 You can also ignore specific source files from code coverage using the `--exclude-by-file` option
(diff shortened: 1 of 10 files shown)


src/coverlet.core/Instrumentation/Instrumenter.cs
```diff
@@ -392,18 +395,22 @@ private static void ReplaceExceptionHandlerBoundary(ExceptionHandler handler, In
...
         {
-            var excludeAttributeNames = new[]
+            // The default custom attributes used to exclude from coverage.
+            IEnumerable<string> excludeAttributeNames = new List<string>()
             {
                 nameof(ExcludeFromCoverageAttribute),
-                "ExcludeFromCoverage",
-                nameof(ExcludeFromCodeCoverageAttribute),
-                "ExcludeFromCodeCoverage"
+                nameof(ExcludeFromCodeCoverageAttribute)
```
tonerdo: This is a backward incompatible change. We should make specifying the `Attribute` suffix optional

src/coverlet.msbuild.tasks/InstrumentationTask.cs
```diff
@@ -50,15 +51,22 @@ public string MergeWith
             set { _mergeWith = value; }
         }
 
+        public string ExcludeByAttribute
```
tonerdo: You need to update the `.props` and `targets` msbuild files as well

README.md
```diff
@@ -281,6 +281,12 @@ You can specify multiple values for `ThresholdType` by separating them with comm
 
 You can ignore a method or an entire class from code coverage by creating and applying the `ExcludeFromCodeCoverage` attribute present in the `System.Diagnostics.CodeAnalysis` namespace.
 
+You can also ignore additional attributes by using the `ExcludeByAttribute` property:
+
+```bash
+dotnet test /p:CollectCoverage=true /p:ExcludeByAttribute="ObsoleteAttribute,GeneratedCodeAttribute,CompilerGeneratedAttribute"
+```
```
tonerdo: You should mention that the `Attribute` suffix is optional. See my other comment

src/coverlet.core/Instrumentation/Instrumenter.cs
amweiss: Is this actually incompatible? From what I saw of the usage even if you use the short name attribute like `[ExcludeFromCoverage]` the attribute name still comes across with the full name `ExcludeFromCoverageAttribute`. Were you trying to account for people writing custom attributes to handle excluding from coverage that were named the same as your attributes?

README.md
amweiss: I had a sight fear since it wasn't using the fully qualified name that it could cause issues, but I can make the it try to do the short name resolution like the compiler as well.

src/coverlet.core/Instrumentation/Instrumenter.cs
tonerdo: It's more of a user experience thing. Since the compiler doesn't enforce the `Attribute` suffix I don't want us to as well. I want the user to be able to specify `Obsolete` and still be fine. You can simply append `Attribute` to the name if it's not specified

amweiss:
I think I addressed the items in your comments @tonerdo

src/coverlet.msbuild/coverlet.msbuild.targets
```diff
@@ -20,6 +21,7 @@
       Exclude="$(Exclude)"
       ExcludeByFile="$(ExcludeByFile)"
       MergeWith="$(MergeWith)"
+      ExcludeByAttributes="$(ExcludeByAttributes)"
```
tonerdo: Let's keep this and all other newly added options singular

tonerdo:
@amweiss This is ready to go, modulo one more comment

amweiss:
@tonerdo Change made.

tonerdo:
This is ready to go. Will merge once I'm ready to make a new release

PR #236, opened 2018-11-07
ExcludeByFile add traversal support when full path specified.

SlowLogicBoy:
Fixes 
`System.ArgumentException: ".." can be only added at the beginning of the pattern.`
when specifying full path with path traversal:
`<ExcludeByFile>$(MSBuildThisFileDirectory)../**/*.Designer.cs</ExcludeByFile>`


SlowLogicBoy:
`D:\Some\Path\..\..\..\SomeOther\Path`: when traversing more then there is path. The only gotcha that bothers me, since I didn't check for that.

PR #241, opened 2018-11-16
Change hits file instrumentation to be memory mapped

### LINKED ISSUE #210: Spurious 'Unable to read beyond end of stream' in CI while collecting results
https://travis-ci.org/tgstation/tgstation-server/jobs/438449753#L1156

Not sure what causes this it rarely happens and seemingly at random. Latest coverlet.console

Cyberboss:
Fixes #210 

Added benefit of getting kernel level cleanup and atomic-ity when incrementing hit counters

### DIFF
--- a/src/coverlet.core/Instrumentation/ModuleTrackerTemplate.cs
+++ b/src/coverlet.core/Instrumentation/ModuleTrackerTemplate.cs
@@ -21,7 +21,7 @@ namespace Coverlet.Core.Instrumentation
         public static string DefaultHitsFilePath;
         public static int DefaultHitsArraySize;
 
-        public static ModuleTrackerTemplate Singleton { get; }
+        private static readonly ModuleTrackerTemplate singleton;
 
         private const string MemoryMappedFileNamePostfix = ".coverlet_memory_mapped";
 
@@ -36,16 +36,16 @@ namespace Coverlet.Core.Instrumentation
             // the module.
 
             if (DefaultHitsFilePath != null)
-                Singleton = new ModuleTrackerTemplate(DefaultHitsFilePath, DefaultHitsArraySize);
+                singleton = new ModuleTrackerTemplate(DefaultHitsFilePath, DefaultHitsArraySize);
 
             //we always keep the view accessor around without disposing it. So, regardless of what happens, be it the module unloading or the process terminating, it will close properly when the finalizers are run and, failing that, when the kernel does process cleanup
         }
 
         public static void RecordHit(int hitLocationIndex)
         {
-            if (Singleton == null)
+            if (singleton == null)
                 throw new InvalidOperationException("Singleton not initialized!");
-            Singleton.InstanceRecordHit(hitLocationIndex);
+            singleton.InstanceRecordHit(hitLocationIndex);
         }
 
         public ModuleTrackerTemplate(string hitsFilePath, int hitsArraySize)
--- a/src/coverlet.core/Instrumentation/ModuleTrackerTemplate.cs
+++ b/src/coverlet.core/Instrumentation/ModuleTrackerTemplate.cs
@@ -18,10 +18,10 @@ namespace Coverlet.Core.Instrumentation
     [ExcludeFromCodeCoverage]
     public static class ModuleTrackerTemplate
     {
+        private const string MemoryMappedFileNamePostfix = ".coverlet_memory_mapped";
+
         public static string hitsFilePath;
         public static int hitsArraySize;
-        
-        private const string MemoryMappedFileNamePostfix = ".coverlet_memory_mapped";
 
         public static MemoryMappedFile memoryMappedFile;
         public static MemoryMappedViewAccessor memoryMappedViewAccessor;
@@ -70,12 +70,6 @@ namespace Coverlet.Core.Instrumentation
             //although the view accessor will keep the mapped file open, we need to not dispose the actual MMF handle
             //doing so will cause the calls to MemoryMappedFile.OpenExisting above to fail
             memoryMappedViewAccessor = memoryMappedFile.CreateViewAccessor();
-        }
-
-        public static void Dispose()
-        {
-            memoryMappedViewAccessor?.Dispose();
-            memoryMappedFile?.Dispose();
         }
 
         public static void RecordHit(int hitLocationIndex)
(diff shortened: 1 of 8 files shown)


WiredUK:
Nice to see someone take this on, I'll try this out when I'm back at work on Monday. Thanks!

src/coverlet.core/Instrumentation/ModuleTrackerTemplate.cs
```diff
@@ -15,130 +16,133 @@ namespace Coverlet.Core.Instrumentation
...
-                _threads.Clear();
-            }
+            this.hitsArraySize = hitsArraySize;
 
-            // The same module can be unloaded multiple times in the same process via different app domains.
-            // Use a global mutex to ensure no concurrent access.
-            using (var mutex = new Mutex(true, Path.GetFileNameWithoutExtension(HitsFilePath) + "_Mutex", out bool createdNew))
+            //now HitsFilePath and HitsArraySize should be populated
+
+            //first, try to create the hits file if it doesn't exist already
```
petli: Could you use https://docs.microsoft.com/en-us/dotnet/api/system.io.memorymappedfiles.memorymappedfile.createnew?view=netcore-2.1#System_IO_MemoryMappedFiles_MemoryMappedFile_CreateNew_System_String_System_Int64_ instead? The file doesn't actually have to contain all zeros, the OS will create zeroed memory pages as necessary.

Cyberboss:
Think I nailed it now

WiredUK:
> Think I nailed it now

Anecdotally at least, this works for me. Ran it a couple of times and coverage matched perfectly.

petli:
I've gave this a spin on my code base, and it seems to count hits correctly. But it is significantly slower than version 2.3.2, where two test projects go from about 50s to respectively 90s and 135s with this change. 

I think this might be related to what @pjanotti discussed in #181 about multithreaded tests.  Even if Interlocking is less bad than using mutexes, it still adds a synchronization between the CPU cores which slows down tests. It would be interesting if @pjanotti could see what the performance is on those projects, since they seemed to be even more penalised by synchronization between parallel unit tests than the projects I test here.

This could perhaps be improved by something similar to #181 keeping one hit count memory mapped file for each CPU core in a thread-static variable and increment it without Interlocking.  But then the Coverage class must find all these files and add upp the hit counts from them.

pjanotti:
Hi @petli, yes I tried an initial perf improvement with Interlocked and for heavy concurrency scenarios it was really bad, so had to fallback on hit counting per thread. There were some projects on the CoreFx repo that were especially bad (I can't recall exactly which ones, but the names may live in some of my previous comments). I would recommend being careful about adding interlocked in the normal hit path count if considering performance for these concurrent scenarios is relevant. Perhaps an option to use the memory mapped file?

(/cc @danmosemsft @ViktorHofer)

petli:
I realised the performance test in coverlet can easily be adjusted to test the threaded behaviour.  This really exercises the code, since the contention gets worse if there's lots of threads all hitting the same memory pages and here's there's twenty tests all looping intensely over the same code:

    public class PerformanceTest
    {
        [Theory]
        [InlineData(20_000)]
        [InlineData(20_001)]
        [InlineData(20_002)]
        [InlineData(20_003)]
        [InlineData(20_004)]
        [InlineData(20_005)]
        [InlineData(20_006)]
        [InlineData(20_007)]
        [InlineData(20_008)]
        [InlineData(20_009)]
        [InlineData(20_010)]
        [InlineData(20_011)]
        [InlineData(20_012)]
        [InlineData(20_013)]
        [InlineData(20_014)]
        [InlineData(20_015)]
        [InlineData(20_016)]
        [InlineData(20_017)]
        [InlineData(20_018)]
        [InlineData(20_019)]
        public void TestPerformance(int iterations)

Running those 20 parallel tests on 2.3.2 took 30s on my dev machine, with the Interlocking usage in this PR it takes... \*twiddles thumbs\* ...16 minutes.  

So before this can be merged it really needs to be changed to use ThreadStatic to keep track of one memory mapped file per thread. :/

Cyberboss:
Alright, taken care of

petli:
I was wrong above that the test case needed additional `InlineData` to stress-test the threaded behaviour, since it already creates a lot of tasks with `Task.Run()`.  There's also a subtle bug lurking in the test which Resharper told me about and it need to change to this to give more reliable numbers:

            for (var i = 0; i < iterations; i++)
            {
                var j = i;
                tasks.Add(Task.Run(() => big.Do(j)));
            }

(I'll do a pull request on this.)

This led me to do an incorrect analysis of the test run I did before the change to thread-local files, the one I wrote about above taking 16 minutes.  It does not seem to be a problem with `Interlocking`, after all, hit counting itself is just much slower using memory mapped files rather than an `int[]`. 

Running just a single run of the performance test with 20000 iterations takes 3s with the old `int[]` counting-then-writing-to-files, 50 seconds with the pointer incrementing (either the original `Interlocking.Increment` or doing `(*hitLocationArrayOffset)++` the one-file-per-thread change). Using Read+Write instead of pointers takes 1.8 minutes.  (The test I ran with 20 `InlineData` didn't expose any additional contention, that's just 50 seconds * 20 tests = 16 minutes.)

@Cyberboss, have you seen this too or is this just something on my machine? Sorry for leading you down the wrong track here...

Cyberboss:
I have yet to look into how to run the performance test but yeah, my own MMF experiments don't bode well with random access incrementing like this. A shared memory implementation (non-file backed) should be infinitely better, I'll see if I can get that working.

WiredUK:
@Cyberboss Any progress on the perf testing? Keen to get this merged so we can fix our CI pipeline.

Cyberboss:
I'll see if I can't find time to readdress this sometime this week, end of year has gotten pretty busy.

petli:
@WiredUK @Cyberboss I've played around with this today, but just changing to a pure memory-based file didn't help much. Instead I've tried a hybrid solution where hit counting is done directly in int arrays as before, but passing the counts to the coverage analysis is done via mapped memory. That can hopefully be quicker than writing to a file and thus have time to finish running in `UnloadModule` even on loaded build servers before it the process is killed.

I have to clean up the code a bit, but will send a PR tomorrow with this.

petli:
Finally pushed #276.

PR #247, opened 2018-11-22
Allow relative output path for console runner

ViktorHofer:
Allows supplying a relative output parameter without a directory name, i.e. `coverage.xml`. Before you had to either pass in a full or relative path, i.e. `.\coverage.xml`.

cc @tonerdo


ViktorHofer:
sure. done.

PR #259, opened 2018-12-08
Add support for assemblies with embedded PDBs

tonerdo:


pape77:
Is it an idea that this is an option to enable or not? Now there's things like "xunit" dlls getting instrumented, or others which are used in tests maybe and we don't care at all for coverage

PR #266, opened 2018-12-16
add threshold checks for overall coverage

### LINKED ISSUE #245: Add option to set threshold only for total coverage
Add option to set threshold only for total coverage.
So run will be succeeded when total coverage more then specified threshold, even if some of projects have coverage 0, or below specified threshold.
Something like 
dotnet test /p:CollectCoverage=true /p:TotalThreshold=80

maikschoeneich:
with this change it is possible to fail the coverage analysis where not a single module is checked against the threshold, but the whole project / solution. 

fixes #245

### DIFF
--- a/README.md
+++ b/README.md
@@ -306,7 +306,7 @@ The above command will automatically fail the build if the line, branch or metho
 dotnet test /p:CollectCoverage=true /p:Threshold=80 /p:ThresholdType=line
 ```
 
-You can specify multiple values for `ThresholdType` by separating them with commas. Valid values include `line`, `branch` and `method`.
+You can specify multiple values for `ThresholdType` by separating them with commas. Valid values include `line`, `branch` and `method` as well as `total-line`, `total-branch` and `total-method`.
 
 #### Excluding From Coverage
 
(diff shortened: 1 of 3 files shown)


tonerdo:
See https://github.com/tonerdo/coverlet/issues/245#issuecomment-447683831 for my thoughts on this. Also, it might make sense to move the threshold functionality into the core library (coverlet.core) since it's shaping up to be an integral part of the overall experience

maikschoeneich:
That's the same I thought while working on it. But I just wanted to make the functionality available as fast as possible because it would indeed benefit a project I am currently working on.

Maybe we can attack the core integration with a follow-up issue?

tonerdo:
I'm all for doing it in a follow-up issue. The only problem is releasing this feature and then going back on it to re-implement my `ThresholdStat` idea will change the public facing APIs and cause backwards incompatible changes.

If you need this feature urgently, I can go ahead to implement is quickly or would you prefer to take a stab at it?

maikschoeneich:
It is not that urgent, and as far it is your idea, you should be the person to implement it in order not to miss any topics you also want to address.

tonerdo:
Addressed in #271 and available in latest releases

PR #276, opened 2018-12-21
Writing hit counts to a shared memory mapped area instead of to a file.

petli:
This might speed up UnloadModule enough that it will reliably execute within the
short time that ProcessUnload allows it, even on CI servers with load.

This is heavily based on work done by @Cyberboss, that unfortunately
showed that  memory mapped files were to slow to use directly in RecordHit.

**Note:** I've only had time to test this code on Windows, and based on #241 there might be some differences in the memory map handling on Linux. That must be tested before this can be merged.


codemzs:
@tonerdo why is this PR not merged?

petli:
Sorry @codemzs, I think this needs more discussion before merging as I'm dubious it helps much.  I'm going to try another approach, so in the meantime I'll close the PR.

petli:
@codemzs @tonerdo  Boy, was my hunch wrong (again!). I added measurements to current coverlet, this branch, and another file -writing solution I wanted to try (details below).  Writing hit counts to file, either solution, takes around 2000-3000 ticks, while the memory map writing takes around 100 ticks.

Caveats here is the usual: it is difficult to compare the results on a developer laptop with the situation on a loaded build machine.  If that has less memory paging could end up causing the mmap solution to be slow too, perhaps slower than the less memory-intense operation of writing to files, but I retract my close and reopens this.

(
For reference, I thought it might be quicker to write the entire hits array as a buffer using the new spans, but it didn't make any significant difference (and in some cases were slower):

    var intBuffer = new ReadOnlyMemory<int>(HitsArray);
    unsafe
    {
        var pinnedBuffer = intBuffer.Pin();
        var byteSpan = new ReadOnlySpan<byte>(pinnedBuffer.Pointer, HitsArray.Length * sizeof(int));
        bw.Write(byteSpan);
    }
)

sharwell:
@petli What about using [`Marshal.Copy`](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.marshal.copy?view=netframework-4.7.2#System_Runtime_InteropServices_Marshal_Copy_System_Int32___System_Int32_System_IntPtr_System_Int32_)?

codemzs:
@petli @tonerdo @cyberboss Can we please get some traction on this PR?

src/coverlet.core/Instrumentation/ModuleTrackerTemplate.cs
```diff
@@ -69,77 +73,90 @@ public static void UnloadModule(object sender, EventArgs e)
...
+                                    }
+
+                                    if (count > 0)
+                                    {
+                                        // There's a header of one int before the hit counts
+                                        var hitLocationArrayOffset = intPointer + i + Coverage.HitsResultHeaderSize;
+
+                                        // No need to use Interlocked here since the mutex ensures only one thread updates 
+                                        // the shared memory map.
+                                        *hitLocationArrayOffset += count;
```
sharwell: ❔ Is the file guaranteed to be zero-filled when it is first created?

test/coverlet.core.tests/Instrumentation/ModuleTrackerTemplateTests.cs
```diff
@@ -9,62 +12,83 @@ namespace Coverlet.Core.Tests.Instrumentation
 {
     public class ModuleTrackerTemplateTestsFixture : IDisposable
     {
+        // Prevent parallel execution of tests using the ModuleTrackerTemplate static class
```
sharwell: ❔ Why not mark the tests as being in a collection that doesn't allow parallel execution? The semaphore works, but it's not the standard xUnit approach so it suffers from decreased clarity.

src/coverlet.core/Instrumentation/ModuleTrackerTemplate.cs
petli: Yes, it's either a pure memory map (on Windows) or a new and thus empty file (on Unixlike), and in both cases zero pages are allocated for the map.

test/coverlet.core.tests/Instrumentation/ModuleTrackerTemplateTests.cs
petli: Brilliant, I didn't know about that feature. I'll change to use that.

petli:
> @petli What about using [`Marshal.Copy`](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.marshal.copy?view=netframework-4.7.2#System_Runtime_InteropServices_Marshal_Copy_System_Int32___System_Int32_System_IntPtr_System_Int32_)?

@sharwell Does that help getting the bytes into the file? That seems to be more about copying an array into memory, but the Spans solve that without copying.

codemzs:
@petli @tonerdo Is there an ETA when this fix can go in and a new nuget be released? We are trying to integrate coverlet in [Microsoft's ML package](https://github.com/dotnet/machinelearning) and currently our CI builds are failing due to [bug ](https://github.com/tonerdo/coverlet/issues/210) this [PR](https://github.com/dotnet/machinelearning/pull/2116) fixes. I will highly appreciate an ETA to determine if we should use an alternative solution since we are under time constraint. Also, if there is a temporary workaround you can suggest that also be appreciated. 

CC: @shauheen

test/coverlet.core.tests/Instrumentation/ModuleTrackerTemplateTests.cs
petli: Fixed

test/coverlet.core.tests/CoverageTests.cs
```diff
@@ -12,6 +12,7 @@
 
 namespace Coverlet.Core.Tests
 {
+    [Collection(nameof(ModuleTrackerTemplate))]
```
sharwell: 📝  I don't see the matching definition, which woul be a class with this:

```
[CollectionDefinition(nameof(ModuleTrackerTemplate))]
```

src/coverlet.template/ModuleTrackerTemplate.cs
```diff
@@ -50,61 +56,76 @@ public static void UnloadModule(object sender, EventArgs e)
...
+                {
+                    // Avoid double counting if unloaded multiple times (if that can happen?)
+                    HitsArray = new int[HitsArray.Length];
+                    
+                    mutex.ReleaseMutex();
+                    memoryMap?.Dispose();
+                }
             }
+
+            s_isTracking = false;
```
petli: Here the no-longer threadstatic field is set to false, in the assumption that UnloadModule is not called in parallel in the same appdomain. This is necessary for the multiple-call test case, but if that assumption is not true (cf #291 which changed it from treadstatic) another solution could be found.

petli:
@tonerdo  @codemzs @sharwell Merged with the other template-related changes, so this PR is again ready for merging.

src/coverlet.template/ModuleTrackerTemplate.cs
petli: Nevermind, @sharwell removes this field entirely in #297... So that should go in to master first, then that change needs to be merged into this PR before it is (again) ready to merge.

codemzs: @petli His change is already merged in master.

sharwell:
I think we should merge #297 first since that is a substantial correctness change for the dotnet/coreclr team, and then update this pull request to focus on the performance improvements.

petli:
> 
> 
> I think we should merge #297 first since that is a substantial correctness change for the dotnet/coreclr team, and then update this pull request to focus on the performance improvements.
> 
> @petli Sorry for all the churn I know that can be frustrating.

I agree on merging #297 first, that order will be easier to sort out codewise too. cc @tonerdo

petli:
@tonerdo This PR is now merged with the latest PRs from @sharwell

tonerdo:
@petli quick sanity check that the merge conflict resolution didn't affect the functionality of both this PR and #297

sharwell:
@tonerdo I verified that it did not impact the intent of my changes that were recently merged

codemzs:
@tonerdo Do we have an ETA for the new nuget with this change?

tonerdo:
@codemzs you'd have an ETA within 24 hours

codemzs:
Thanks, @tonerdo ! This is good news, looking forward to getting the new nuget and integrating coverlet with dotnet/machinelearning. 

CC: @shauheen

ViktorHofer:
Unfortunately this broke System.Private.CoreLib coverage again as MemoryMappedFile isn't available in it: https://github.com/dotnet/corefx/pull/34641#issuecomment-455181512.

ViktorHofer:
@petli you were looped in in the discussion in corefx where Jan pointed out that the early termination is not triggered by the CLR. I created https://github.com/Microsoft/vstest/issues/1900 as the vstest framework is the one who terminates the process before the OnProcessExit (+ ProcessExit events) call is completed. 

I think we should revert this change and follow Jan's recommendation and just do sequential IO transfer. This will unblock the currently failing System.Private.CoreLib coverage measurements.

petli:
I agree, no point complicating things with memory maps when they won't always help.

ViktorHofer:
Thanks for understanding. Meanwhile I try to get in touch with the vstest team internally to better understand why they are terminating the application too soon.

Issue #275, opened 2018-12-21
False negative with "/p:MergeWith" async/await branches


MarcoRossignoli:
I moved discussion from old PR to new issue.

---------------------
@pape77 issue https://github.com/tonerdo/coverlet/pull/158#issuecomment-448654060

I have a similar issue with this "ghost branches" which are never covered.
My code looks like this

    public class KeyProvider : IKeyProvider
    {
        private readonly IKeyRepository _repository;
        private readonly IMemoryCache _cache;

        public KeyProvider(IKeyRepository repository, IMemoryCache cache)
        {
            _repository = repository;
            _cache = cache;
        }

        public async Task<Key> GetByUuidAndTypeAsync(Guid uuid, KeyType type, CancellationToken cancellationToken)
        {
            return await _cache.GetOrCreateAsync($"uuid-{uuid.ToString()}-key-{type}",
                async entry => await _repository.FindOrCreateAsync(uuid, type, cancellationToken));
        }

   public class Key
    {
        public int Id { get; set; }

        public Guid Uuid { get; set; }

        [Required]
        public Guid Kid { get; set; }

        [Required]
        public KeyType KeyType { get; set; }
    }

 public enum KeyType
    {
        Cenc,
        Fairplay,
        Aes
    }
}
This code generates 4 branches, 2 of which I'm never able to cover

my tests:

  public class KeyProviderTests
    {
        private readonly Mock<IKeyRepository> _keyRepositoryMock;
        private readonly MemoryCache _memoryCache;
        private readonly Guid _testGuid = Guid.NewGuid();
        private readonly KeyProvider _keyProvider;

        public KeyProviderTests()
        {
            _memoryCache = new MemoryCache(new MemoryCacheOptions());

            _keyRepositoryMock = new Mock<IKeyRepository>();

            _keyProvider = new KeyProvider(_keyRepositoryMock.Object, _memoryCache);
        }

        [Theory(DisplayName = "Should call repository method with correct values")]
        [InlineData(KeyType.Cenc)]
        [InlineData(KeyType.Aes)]
        [InlineData(KeyType.Fairplay)]
        public async Task GetByUuidAndTypeAsync01(KeyType keyType)
        {
            _keyRepositoryMock.Setup(kr =>
                    kr.FindOrCreateAsync(_testGuid, keyType, It.IsAny<CancellationToken>()))
                .ReturnsAsync(new Key());

[...]

PR #158, opened 2018-07-19
Try to fix branch coverage with async calls bug

MarcoRossignoli:
fixes https://github.com/tonerdo/coverlet/issues/113

The issue is related to compiler generated state machine:

```cs
...
TaskAwaiter awaiter;
if (num != 0)
{
    CoverageTracker.MarkExecuted(@"...\AppData\Local\Temp\CoverletAwaitBranchCoverage_04c15549-0ca8-4cd2-977b-01bd0a8ea52e", @"L,...\Git\CoverletAwaitBranchCoverage\CoverletAwaitBranchCoverage\Program.cs,9,9");
    CoverageTracker.MarkExecuted(@"...\AppData\Local\Temp\CoverletAwaitBranchCoverage_04c15549-0ca8-4cd2-977b-01bd0a8ea52e", @"L,...\Git\CoverletAwaitBranchCoverage\CoverletAwaitBranchCoverage\Program.cs,10,10");
    awaiter = Task.get_CompletedTask().GetAwaiter();
    if (!awaiter.IsCompleted)
    {
        CoverageTracker.MarkExecuted(@"...\AppData\Local\Temp\CoverletAwaitBranchCoverage_04c15549-0ca8-4cd2-977b-01bd0a8ea52e", @"B,...\Git\CoverletAwaitBranchCoverage\CoverletAwaitBranchCoverage\Program.cs,10,0");
        this.<>1__state = num = 0;
        this.<>u__1 = awaiter;
        Program.<Main>d__0 stateMachine = this;
        this.<>t__builder.AwaitUnsafeOnCompleted<TaskAwaiter, Program.<Main>d__0>(ref awaiter, ref stateMachine);
        return;
    }
}
...
```

If task complete synchronously this branch doens't hit.
I borrowed this idea from OpenCover.
I tested 2 strategies:
https://github.com/tonerdo/coverlet/pull/158/commits/3edf519e7ca578445f9c58fc020f5907f7528420#diff-9eb1afa9bd8141e48d456c120df924caR185 this one force hit on not covered branch
https://github.com/tonerdo/coverlet/pull/158/commits/06a5e59d06c14e7319cb88d80d9e02b74ab7c862#diff-9eb1afa9bd8141e48d456c120df924caR199 this one remove completely the branch if not hitted(but other one yes)

I prefer remove the branch...is more real than force hit. 
Let me know what do you think.

/cc @tonerdo

PR #277, opened 2018-12-23
Fix false negative with "/p:MergeWith" async/await branches

MarcoRossignoli:
fixes https://github.com/tonerdo/coverlet/issues/275

Similar issue fixed in https://github.com/tonerdo/coverlet/pull/158
If we use merge options we need to remove uncoverable branch of async state machine.

For instance if we have 2 different tests project, and first test project doesn't cover async call but the second one yes we end with this situation:
we call branches b1 and b2 and we have(hints)
for first test b1 = 0 b2 = 0
for second test b1 = 1 and b2 = 0 -> in this case we remove b2 because false negative, it depends on async or sync execution(#158)
So if we merge results we have b1 = 1 and b2 = 0 and we need to remove b2 = 0 because will never be covered, it depends on async/sync execution.

Here you can find repro project(thank's to @pape77) https://github.com/MarcoRossignoli/marcorossignoli.github.io/tree/master/src/coverlet/Issue1
before fix
```
      "Keys.Data.Repositories.KeyRepository/<FindAsync>d__3": {
        "System.Void Keys.Data.Repositories.KeyRepository/<FindAsync>d__3::MoveNext()": {
          "Lines": {
            "27": 2,
            "28": 2,
            "29": 2
          },
          "Branches": [
            {
              "Line": 28,
              "Offset": 61,
              "EndOffset": 127,
              "Path": 1,
              "Ordinal": 1,
              "Hits": 2
            },
            {
              "Line": 28,
              "Offset": 61,
              "EndOffset": 63,
              "Path": 0,
              "Ordinal": 0,
              "Hits": 0
            }
          ]
        }
      },
```
after fix
```
      "Keys.Data.Repositories.KeyRepository/<FindAsync>d__3": {
        "System.Void Keys.Data.Repositories.KeyRepository/<FindAsync>d__3::MoveNext()": {
          "Lines": {
            "27": 2,
            "28": 2,
            "29": 2
          },
          "Branches": [
            {
              "Line": 28,
              "Offset": 61,
              "EndOffset": 127,
              "Path": 1,
              "Ordinal": 1,
              "Hits": 2
            }
          ]
        }
      },
```

/cc @tonerdo @pape77 @kevindqc


pape77:
I would extract the logic to a method in a new class not to have duplicated code here and in Coverage.cs :)

MarcoRossignoli:
> I would extract the logic to a method in a new class not to have duplicated code here and in Coverage.cs :)

Logic is the same but objects and place are different, Branch vs BranchInfo, different collections objects and different if conditions. I'll try to understand if I can dedup.

pape77:
@MarcoRossignoli I slightly modified the project that I sent you earlier in the issue #275 and the same seems to happen, it seems. I'll post it in the issue for you to take a look

MarcoRossignoli:
I did some changes now I'm sure that MoveNext() comes from  async state machine, the only way I found to recognize it on merge it was add one property on branche structure when persist it, otherwise I should reload lib and search method name literal. New "coverlet" format would be i.e.:
```
"Keys.Data.Repositories.KeyRepository/<FindAsync>d__3": {
        "System.Void Keys.Data.Repositories.KeyRepository/<FindAsync>d__3::MoveNext()": {
          "Lines": {
            "27": 2,
            "28": 2,
            "29": 2
          },
          "Branches": [
            {
              "Line": 28,
              "Offset": 61,
              "EndOffset": 127,
              "Path": 1,
              "Ordinal": 1,
              "Hits": 2,
              "IsAsyncStateMachineBranch": true
            }
          ]
        }
      },
```

/cc @pape77 @tonerdo

src/coverlet.core/CoverageResult.cs
```diff
@@ -121,7 +122,7 @@ internal void Merge(Modules modules)
                         {
                             foreach (var branch in method.Value.Branches)
                             {
-                                if (branch.IsAsyncStateMachineBranch && method.Key.EndsWith("::MoveNext()") && branch.Hits > 0)
+                                if (IsAsyncStateMachineMethod(method.Key))
```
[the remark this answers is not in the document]
MarcoRossignoli: Good catch yes...I'm OOF I'll fix ASAP...I missed that because I found other coverage branch ok on json thank's for review!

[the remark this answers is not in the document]
MarcoRossignoli: fixed!

pape77:
@MarcoRossignoli 
You can also get rid of the ```IsMoveNext``` method in ```CecilSymbolHelper```
My only doubt is, I see the ```GetBranchPoints``` in that class is still using the regex of moveNext , could this be an issue?

MarcoRossignoli:
> @MarcoRossignoli
You can also get rid of the IsMoveNext method in CecilSymbolHelper
My only doubt is, I see the GetBranchPoints in that class is still using the regex of moveNext , could this be an issue?

In past I only exposed `IsMoveNext` for my usage...we cannot remove that code...is foundamental for instrumentation...maybe we could make it private again.

src/coverlet.core/Coverage.cs
```diff
@@ -232,14 +232,14 @@ private void CalculateCoverage()
...
                 // for MoveNext() compiler autogenerated method we need to patch false positive (IAsyncStateMachine for instance) 
-                // we'll remove all MoveNext() not covered branch              
+                // we'll remove all MoveNext() not covered branch
                 foreach (var document in result.Documents)
                 {
                     List<KeyValuePair<(int, int), Branch>> branchesToRemove = new List<KeyValuePair<(int, int), Branch>>();
                     foreach (var branch in document.Value.Branches)
                     {
                         //if one branch is covered we search the other one only if it's not covered
-                        if (CecilSymbolHelper.IsMoveNext(branch.Value.Method) && branch.Value.Hits > 0)
```
pape77: I mean this method, you don't use it anymore and it was only used here, so can be removed. the regex can maybe remain there (it has the same name, so maybe my comment wasn't completely understood :P)

MarcoRossignoli: I did it...I edited my last message!Thank's again!

pape77: check, thanks! :D

pape77:
@MarcoRossignoli yes! Works for me. I'm using it in my own fork of coverlet ;)

tonerdo:
@pape77 are there any noticeable regressions in other branch related functionality?
@MarcoRossignoli do you think you can add a test or two for this fix?

MarcoRossignoli:
> @pape77 are there any noticeable regressions in other branch related functionality?

The only regression I could see is more memory pressure due to instrumentedResult.AsyncMachineStateMethod string list(one record for every async/await method more or less), because to avoid to "search" method(too slow to me) I use a memory list cache, another approach could be write to file and check on it(but more plumbing and maybe slowest that memory approach).
Another "imperfect" approach is to check for method ends with `::MoveNext()` but we could have false positive.

> @MarcoRossignoli do you think you can add a test or two for this fix?

I'll take a look!

pape77:
@tonerdo not that I noticed. We use this for several projects at my work. Everything looks great now

tonerdo:
@MarcoRossignoli seems you need to fix a rebase. Will merge once that's done. Sorry for the delay on this

MarcoRossignoli:
>  Will merge once that's done

Merged

> Sorry for the delay on this

No problem!Thank's again for coverlet!

topres:
Did this change get in the 2.5.1 release on NuGet? I still seem to have missing branch coverage on my await statements

MarcoRossignoli:
@topres I think so https://github.com/tonerdo/coverlet/releases https://github.com/tonerdo/coverlet/blob/master/coverlet.msbuild.nuspec#L5
Can you provide a repro?

PR #286, opened 2019-01-07
Make the project function with .Net Framework classes again

glennawatson:
This fixes #285, fixes #274 and fixes #272

Changed back to using a .NET Standard project for the coverlet.core. Allows for more targets (eg Xamarin based targets too potentially). 

Also do some redirecting of StdError for the process output, xunit for example sends errors there, and put them into the log.

.NET STandard 2.1 adds RelativePath() so using a place holder until that comes out.

PR #291, opened 2019-01-13
Record hits without thread local variables

sharwell:
On my machine (2990wx), this change appears to be good for a 25% performance increase. I measured performance by setting the number of iterations in `TestPerformance` to 1,000,000 and repeating the runs a few times using the instructions in the README file. The times before this change were 22.5 seconds, and after this change were 15.5 seconds.

| Scenario | Time (seconds, approx.) |
| --- | --- |
| No coverage | 7.0 |
| Empty `RecordHit` | 8.9 |
| `Interlocked.Increment` | 15.5 |
| Thread static | 22.5 |

### DIFF
--- a/src/coverlet.template/ModuleTrackerTemplate.cs
+++ b/src/coverlet.template/ModuleTrackerTemplate.cs
@@ -20,24 +20,17 @@ namespace Coverlet.Core.Instrumentation
         public static string HitsFilePath;
         public static int[] HitsArray;
 
-        // Special case when instrumenting CoreLib, the thread static below prevents infinite loop in CoreLib
+        // Special case when instrumenting CoreLib, the static below prevents infinite loop in CoreLib
         // while allowing the tracker template to call any of its types and functions.
-        [ThreadStatic]
-        private static bool t_isTracking;
-
-        [ThreadStatic]
-        private static int[] t_threadHits;
-
-        private static List<int[]> _threads;
+        private static bool s_isTracking;
 
         static ModuleTrackerTemplate()
         {
-            t_isTracking = true;
-            _threads = new List<int[]>(2 * Environment.ProcessorCount);
-
+            s_isTracking = true;
             AppDomain.CurrentDomain.ProcessExit += new EventHandler(UnloadModule);
             AppDomain.CurrentDomain.DomainUnload += new EventHandler(UnloadModule);
-            t_isTracking = false;
+            s_isTracking = false;
+
             // At the end of the instrumentation of a module, the instrumenter needs to add code here
             // to initialize the static fields according to the values derived from the instrumentation of
             // the module.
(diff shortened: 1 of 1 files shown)


codemzs:
Looks great! @petli would be nice if we can also get your PR #276 in very soon as this is a major blocker for us. @tonerdo , can you please look at these two PRs soon and issue a new nuget?

petli:
There was a problem using interlocking that slowed down bigger tests that was the reason to use thread local variables. It especially affected CoreLib, if I remember correctly. Look at the git blame log and it should show up the PR and discussion. (I can't find that right now as I'm just replying from my phone).

sharwell:
I'm really interested to see what that issue was. Typically if something is going to have a problem during parallel execution, it will show up when running on my machine (64 logical processors).

sharwell:
I checked the history back to the creation of the template class, and didn't see anything related. I'll try to look back further today.

petli:
@sharwell I've been able to dig through the PRs now, the relevant comment is this one:
https://github.com/tonerdo/coverlet/pull/241#issuecomment-440717124

It is worthwhile to check if that is still true for CoreFx, if you can.

sharwell:
@petli From the comments in that thread, it sounds like interlocked wasn't a problem after all, and the original analysis was a false reading from combining the use of Interlocked.Increment with memory locations mapped to a file.

petli:
@sharwell The thread as a whole was about memory maps, but that comment was the best source I could find for the discussion about interlocked in general. It refers to #181 where interlocked was first attempted (without any memory maps). It was replaced in https://github.com/tonerdo/coverlet/pull/181/commits/58b89a56d77238029ddcf4d1e7bf3989d48b1632 with the thread-local arrays.

tonerdo:
@codemzs I typically default to @petli for all multi-threading related functionality as he has much better experience with it. If he says this is good to go then I'm good

petli:
I had a go at testing this PR with CoreFx, and outcome is that it improves some tests by the 25% seen on the basic performance test, and doesn't affect others much. Test execution time in seconds:

Project | 1.4.0 | PR
--------|------|---
System.Collections.Concurrent.Tests | 13 | 12
System.Collections.Tests | 24 | 20
System.Reflection.Metadata.Tests | 72 | 50
System.Xml.Linq.Events.Tests | 160 | 158
System.Runtime.Serialization.Formatters.Tests | 588 | 420

@pjanotti, do you think I was lucky in the selection of the tests I've run? Could the different ways of calling Interlocked (by generating IL code as in your first commit, vs calling it from RecordHit like here) make a difference?

@tonerdo @sharwell With these numbers I'm comfortable merging this, but keeping an eye on the CoreFx projects once this is released to ensure there's no degradation outside the small sample here.

petli:
Btw, described how to run these tests in #292

sharwell:
@petli I would expect this change to show improvement primarily relative to the baseline (No coverage). In other words, for cases above where the difference between 1.4.0 and this PR was smaller, I would expect those same tests to show a smaller difference between the No coverage and coverage with 1.4.0 cases.

ViktorHofer:
Note that System.Private.CoreLib coverage isn't working right now and hits are not recorded. So make sure to choose projects that don't depend on it --> only test projects that don't have the `TestRuntime` property set to true.

System.Reflection.Metadata.Tests | 72 | 50 (correct)
System.Xml.Linq.Events.Tests | 160 | 158 (correct)

That said I think the PR clearly shows an improvement.

sharwell:
Rebased without significant changes to this PR to address the merge conflict from #286

tonerdo:
> Note that System.Private.CoreLib coverage isn't working right now and hits are not recorded.

tonerdo:
> Note that System.Private.CoreLib coverage isn't working right now and hits are not recorded.

@ViktorHofer any particular reason why this is the case?

src/coverlet.template/ModuleTrackerTemplate.cs
```diff
@@ -20,68 +20,33 @@ public static class ModuleTrackerTemplate
...
+        // Special case when instrumenting CoreLib, the static below prevents infinite loop in CoreLib
         // while allowing the tracker template to call any of its types and functions.
-        [ThreadStatic]
-        private static bool t_isTracking;
-
-        [ThreadStatic]
-        private static int[] t_threadHits;
-
-        private static List<int[]> _threads;
+        private static bool s_isTracking;
```
Cyberboss: Late to the party, but is making this non-thread static safe?

sharwell: No. There are only two cases where this flag is used:

1. The static constructor, which the CLR guarantees will complete on exactly one thread.
2. The `UnloadModule` handler, after which point the value will always be true.

ViktorHofer:
@tonerdo The instrumented System.Private.CoreLib.dll throws a StackOverflowException. This very painful for us in corefx but unfortunately I couldn't find the issue in the instrumentation yet. If you want to offer help I can quickly write up steps to reproduce the issue. I would be thrilled.

ViktorHofer:
```
Failed to request MethodData, not in JIT code range
MethodDesc:   00007ffe8f19b9c0
Method Name:          Coverlet.Core.Instrumentation.Tracker.System.Private.CoreLib_c64a6f40-c11a-4e9c-b1eb-86506d4b4c35.RecordHit(Int32)
Class:                00007ffe8f1a2980
MethodTable:          00007ffe8f19b9f0
mdToken:              00000000060052c9
Module:               00007ffe8eff6020
IsJitted:             yes
Current CodeAddr:     00007ffe8f116510
Code Version History:
  CodeAddr:           00007ffe8f116510  (Tier 0)
  NativeCodeVersion:  0000000000000000
```

PR #292, opened 2019-01-14
Instructions on how to test dev versions of coverlet against CoreFx

petli:


ViktorHofer:
Thanks a lot for documenting this. Just as a side note, System.Private.CoreLib measurement doesn't correctly work as the instrumentation causes a StackOverflowException in it. I have disabled it for now until I have more time to diagnose what's going on.

PR #297, opened 2019-01-15
Remove reliance on s_isTracking to detect recursion

sharwell:
Fixes dotnet/coreclr#21994

The stack overflow was caused when a call to `RecordHit` occurred prior to the initialization of `HitsArray`, which in turn called an instrumented constructor for `NullReferenceException`.

### DIFF
--- a/src/coverlet.core/Instrumentation/Instrumenter.cs
+++ b/src/coverlet.core/Instrumentation/Instrumenter.cs
@@ -29,6 +29,7 @@ namespace Coverlet.Core.Instrumentation
         private FieldDefinition _customTrackerHitsFilePath;
         private ILProcessor _customTrackerClassConstructorIl;
         private TypeDefinition _customTrackerTypeDef;
+        private MethodReference _customTrackerRegisterUnloadEventsMethod;
         private MethodReference _customTrackerRecordHitMethod;
 
         public Instrumenter(string module, string identifier, string[] excludeFilters, string[] includeFilters, string[] excludedFiles, string[] excludedAttributes)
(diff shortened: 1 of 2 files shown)


src/coverlet.template/ModuleTrackerTemplate.cs
```diff
@@ -38,15 +32,18 @@ static ModuleTrackerTemplate()
...
                 return;
 
             Interlocked.Increment(ref HitsArray[hitLocationIndex]);
         }
 
         public static void UnloadModule(object sender, EventArgs e)
         {
-            s_isTracking = true;
+            // Claim the current hits array and reset it to prevent double-counting scenarios.
+            var hitsArray = Interlocked.Exchange(ref HitsArray, new int[HitsArray.Length]);
```
petli: Shouldn't HitsArray be set to null here in the exchange to avoid RecordHit doing any additional counting in CoreLib now when UnloadModule is called?

sharwell: Additional counting is safe and not inherently a bug. This approach guarantees that if the null check in `RecordHit` passes, then the access of an array element on the next line will also see a non-null array.

petli: True, those counts will be ignored since it's in a new array.

codemzs:
@sharwell Is this PR ready to be merged?

sharwell:
@codemzs No, I'm working with @ViktorHofer to figure this out still

sharwell:
OK this should be ready now 😄

src/coverlet.core/Instrumentation/Instrumenter.cs
```diff
@@ -76,6 +77,7 @@ private void InstrumentModule()
 
                 using (var module = ModuleDefinition.ReadModule(stream, parameters))
                 {
+                    var containsAppContext = module.GetType(nameof(System), nameof(AppContext)) != null;
```
sharwell: 📝 This could be modified to also check the name of the containing assembly, but it didn't _seem_ necessary.

ViktorHofer:
cc @tonerdo

tonerdo:
I see you all have been busy 😄. Thanks for all the help guys

PR #299, opened 2019-01-16
Improve build tasks packaging

sharwell:
* [x] Create the **coverlet.msbuild.nupkg** package as a primary build output of **coverlet.msbuild.tasks**
* [x] Split **coverlet.msbuild.tasks** into **netstandard2.0** (or **netcoreapp2.0**) and **net472** target frameworks
* [x] (optional) Attach the .nupkg files as artifacts of the AppVeyor build
* [x] ~~(optional) Use Nerdbank.GitVersioning to dramatically simplify the versioning process for the whole project~~ (submitted as #301)

This change will allow the build-time instrumentation to work with both `dotnet msbuild` (works already using .NET Core as the host runtime) and `msbuild` (currently fails using .NET Framework as the host runtime). The strategy will be similar to the one I use for ANTLR, except much simpler because I don't need to account for very old hosts like MSBuild 4.0.

I verified that c35018acdfe6a878ccb3c49361d4cde760df055e produces the same output assembly that shipped as 2.5.0 on NuGet except for the following changes:

1. A few additional PDB files are included (not a problem)
2. **coverlet.template.dll** is included (even if unnecessary, it's very small, but let me know if this needs to be removed)
3. The **coverlet.msbuild.props** and **coverlet.msbuild.targets** files are moved up a folder, so they are included in the build regardless of the target framework of the project using coverlet


sharwell:
Marked as WIP while I rebase to account for recent merges.

sharwell:
@tonerdo This is now updated. GitHub will detect a trivial conflict between this and #303, so when you merge one I'll update the other to account for the change.

src/coverlet.msbuild.tasks/coverlet.msbuild.tasks.csproj
```diff
@@ -1,18 +1,63 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <OutputType>Library</OutputType>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <TargetFrameworks>netcoreapp2.0;net472</TargetFrameworks>
```
ViktorHofer: You can probably remove the extra TargetFrameworkOverride stuff with this:

```suggestion
    <TargetFrameworks Condition="'$(OS)' == 'Windows_NT'">netcoreapp2.0;net472</TargetFrameworks>
    <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">netcoreapp2.0</TargetFrameworks>
```

ViktorHofer:
What's the reason that coverlet.msbuild cross-compiles against net472 but not coverlet.console?

ViktorHofer:
Wouldn't it also make sense to merge coverlet.msbuild and coverlet.msbuild.tasks into one folder?

src/coverlet.core/coverlet.core.csproj
```diff
@@ -2,13 +2,14 @@
...
     <OutputType>Library</OutputType>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <TargetFrameworks>netcoreapp2.0;net472</TargetFrameworks>
     <AssemblyVersion>4.1.0</AssemblyVersion>
   </PropertyGroup>
 
   <ItemGroup>
     <PackageReference Include="Mono.Cecil" Version="0.10.1" />
     <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
+    <PackageReference Include="System.Reflection.Metadata" Version="1.6.0" />
```
ViktorHofer:
```suggestion
    <PackageReference Condition="'$(TargetFramework)' == 'net472'" Include="System.Reflection.Metadata" Version="1.6.0" />
```

src/coverlet.msbuild/coverlet.msbuild.props
```diff
@@ -13,5 +13,8 @@
     <Threshold Condition="$(Threshold) == ''">0</Threshold>
     <ThresholdType Condition="$(ThresholdType) == ''">line,branch,method</ThresholdType>
     <ThresholdStat Condition="$(ThresholdStat) == ''">minimum</ThresholdStat>
+
+    <CoverletBuildTaskPath Condition="'$(MSBuildRuntimeType)' != 'Core'">$(MSBuildThisFileDirectory)net472\</CoverletBuildTaskPath>
+    <CoverletBuildTaskPath Condition="'$(MSBuildRuntimeType)' == 'Core'">$(MSBuildThisFileDirectory)netcoreapp2.0\</CoverletBuildTaskPath>
```
ViktorHofer: Thanks a lot for that! 👍

src/coverlet.msbuild.tasks/coverlet.msbuild.tasks.csproj
```diff
@@ -1,18 +1,63 @@
...
+    <ProjectReference Include="$(MSBuildThisFileDirectory)..\coverlet.core\coverlet.core.csproj" PrivateAssets="All" />
   </ItemGroup>
 
   <ItemGroup>
-    <ProjectReference Include="$(MSBuildThisFileDirectory)..\coverlet.core\coverlet.core.csproj" />
+    <None Include="..\coverlet.msbuild\coverlet.msbuild.props">
+      <Link>build\coverlet.msbuild.props</Link>
+      <Pack>true</Pack>
+    </None>
+    <Content Include="..\coverlet.msbuild\coverlet.msbuild.targets">
```
ViktorHofer: is it intentional that you are using a Content Element here but for the props a None element? Also for the props you set `Pack` to true but not here?

src/coverlet.template/coverlet.template.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netcoreapp2.0;net472</TargetFrameworks>
```
ViktorHofer: Are you sure this change is correct? I believe the template targets netstandard2.0 because of the references that get copied over into the instrumented assembly with Mono.Cecil. At least for S.P.CoreLib we are removing the netstandard references again afterwards.

src/coverlet.msbuild.tasks/coverlet.msbuild.tasks.csproj
sharwell: Not intentional. I was experimenting and found that both worked, but forgot to settle on one. 😆

src/coverlet.template/coverlet.template.csproj
sharwell: Yes. On .NET Framework builds, the façade assemblies are not guaranteed to be available.

src/coverlet.core/coverlet.core.csproj
sharwell: 💭 I don't think this change would have any real impact. It seems simpler to keep it like this?

sharwell:
> What's the reason that coverlet.msbuild cross-compiles against net472 but not coverlet.console?

I haven't done any work to get coverlet.console working differently. This pull request is about making MSBuild-based instrumentation work on .NET Framework builds.

src/coverlet.msbuild.tasks/coverlet.msbuild.tasks.csproj
sharwell: Either way feels like a hack; not sure one is better than the other.

src/coverlet.core/coverlet.core.csproj
sharwell: @ViktorHofer If we were creating a NuGet package from this and trying to avoid unnecessary declared dependencies in the result, I could see limiting this. However, that's not _currently_ the case. I can certainly make this change if it's needed or useful for some reason but wanted to verify first.

src/coverlet.msbuild.tasks/coverlet.msbuild.tasks.csproj
sharwell: Resolved this along with https://github.com/tonerdo/coverlet/pull/299#issuecomment-454998349 in 1f6c7ea6b36cf07e68eeb643ec4d35a9a274cf7f.

src/coverlet.msbuild.tasks/coverlet.msbuild.tasks.csproj
ViktorHofer: The way that I outlined is the one that customers are usually using. I would prefer this offer the explicit override. If you directly build this one project on Unix with dotnet wouldn't that result in an error with net472 as the target framework not conditioned?

src/coverlet.core/coverlet.core.csproj
ViktorHofer: Right we are not producing a package out of coverlet.core yet but IMO it's always a good thing to limit dependencies when introducing them. Others could keep that dependency even if we ship a package because they don't have context why this was added. As it's a one line change, pls do it 😁

ViktorHofer:
Oh right. Coverlet.Console is a global tool, that's why it never supported .NET Framework.

ViktorHofer:
@tonerdo as we now cross-compile coverlet.core and coverlet.template here I propose to undo the unnecessary coverlet.template project and move the template back into core.

sharwell:
> as we now cross-compile coverlet.core and coverlet.template here I propose to undo the unnecessary coverlet.template project and move the template back into core.

I do have a remaining concern that's difficult to validate prior to some of the unmerged pull requests. This pull request separates the instrumentation task according to host runtime (msbuild vs. dotnet msbuild). However, the target framework of the individual projects might also come into play, and I haven't yet verified all the combinations there.

ViktorHofer:
ok fine. Let's keep it in mind.

src/coverlet.msbuild.tasks/coverlet.msbuild.tasks.csproj
sharwell: > If you directly build this one project on Unix with dotnet wouldn't that result in an error with net472 as the target framework not conditioned?

Yes. The error alerts users to the fact that `dotnet` on Linux cannot currently build a valid package for this project. I will add a comment to the TargetFrameworkOverride line to make it clear that it's only there because of dotnet/sdk#335. For the project file itself, the current form better captures the desired end state.

src/coverlet.core/coverlet.core.csproj
sharwell: Fixed

src/coverlet.msbuild.tasks/coverlet.msbuild.tasks.csproj
sharwell: Comment now added

src/coverlet.msbuild.tasks/coverlet.msbuild.props
```diff
@@ -13,5 +13,8 @@
     <Threshold Condition="$(Threshold) == ''">0</Threshold>
     <ThresholdType Condition="$(ThresholdType) == ''">line,branch,method</ThresholdType>
     <ThresholdStat Condition="$(ThresholdStat) == ''">minimum</ThresholdStat>
+
+    <CoverletBuildTaskPath Condition="'$(MSBuildRuntimeType)' != 'Core'">$(MSBuildThisFileDirectory)net472\</CoverletBuildTaskPath>
+    <CoverletBuildTaskPath Condition="'$(MSBuildRuntimeType)' == 'Core'">$(MSBuildThisFileDirectory)netcoreapp2.0\</CoverletBuildTaskPath>
```
ViktorHofer: Why not point to the full path here and use it? We do the same for ApiCompat and GenApi, i.e. https://github.com/dotnet/arcade/blob/master/src/Microsoft.DotNet.ApiCompat/build/Microsoft.DotNet.ApiCompat.targets#L9

sharwell: They seem equivalent to me, and the form I used appears in some other packages that generate code during builds (e.g. ANTLR). Deferring a change here as 🚲 🏛

ViktorHofer: I don't really have a strong opinion here I just don't see a reason to introduce a property that points to the msbuild task assembly's directory. The directory doesn't bring any value to the user but a path to the assembly does.

ViceIce:
Any timeframe when this will be merged and released? I really need that this. 😀

ViktorHofer:
@sharwell can you please resolve the conflicts

ViktorHofer:
@tonerdo can we please get this into master after the conflicts are resolved? would be great.

tonerdo:
@ViktorHofer I'll need to get some time to take a closer look at this after the conflicts are fixed. Been quite busy with work lately

sharwell:
@tonerdo I fixed the conflicts

src/coverlet.core/Coverage.cs
```diff
@@ -294,9 +294,10 @@ private string GetSourceLinkUrl(Dictionary<string, string> sourceLinkDocuments,
                 string key = sourceLinkDocument.Key;
                 if (Path.GetFileName(key) != "*") continue;
 
-                string relativePath = Path.GetRelativePath(Path.GetDirectoryName(key), Path.GetDirectoryName(document));
+                if (!Path.GetDirectoryName(document).StartsWith(Path.GetDirectoryName(key) + Path.DirectorySeparatorChar))
+                    continue;
 
-                if (relativePath.Contains("..")) continue;
+                var relativePath = Path.GetDirectoryName(document).Substring(Path.GetDirectoryName(key).Length + 1);
```
tonerdo: If this is indeed a functional workaround to using `Path.GetRelativePath` then we can make `coverlet.msbuild` be fully compliant with `netcoreapp2.0` and `net472` frameworks by making it target `netstandard2.0` instead

src/coverlet.msbuild.tasks/coverlet.msbuild.tasks.csproj
```diff
@@ -2,20 +2,62 @@
...
+    <PackageTags>coverage testing unit-test lcov opencover quality</PackageTags>
+    <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
+
+    <!-- Items in the Content group are placed in this folder in the NuGet package. -->
+    <ContentTargetFolders>build</ContentTargetFolders>
+
+    <!-- Build tasks should not be added to the lib folder. -->
+    <IncludeBuildOutput>false</IncludeBuildOutput>
+    <IncludeSymbols>true</IncludeSymbols>
+    <IncludeSources>true</IncludeSources>
```
tonerdo: Any reason we're including symbols and sources?

src/coverlet.core/Coverage.cs
tonerdo: If this is indeed a functional workaround to using `Path.GetRelativePath` then we can make `coverlet.msbuild` be fully compliant with `netcoreapp2.0` and `net472` frameworks by making it target `netstandard2.0` instead

src/coverlet.msbuild.tasks/coverlet.msbuild.tasks.csproj
sharwell: It can improve the debugging experience if needed. We can alter the way this is included in a future PR if desired.

src/coverlet.core/Coverage.cs
tonerdo: To my knowledge Coverlet worked quite well with `net472` when the target was `netstandard2.0`. Only got issues when it was changed to `netcoreapp2.0`

ViceIce: It is even working with net35

tonerdo: @sharwell are you still able to this? If not, I'll just merge it in and make the necessary changes if that's alright with you

tonerdo: Thanks for your help so far

sharwell: @tonerdo I wanted to merge this one like this, then update and merge #301, and only then start to focus on enabling .NET Framework scenarios. Currently testing for the latter is completely blocked so I can't say for sure what will or will not work.

tonerdo: I'd prefer if it gets done in this PR. #301 might take a bit to get through and I want the next release (which I want to happen this week) to fix the .NET Framework issue

sharwell: I'm not sure I'll have time to make and test this modification.

zsd4yr:
Just wanted to chime in and say that I am trying to get code coverage setup for https://github.com/dotnet/winforms, and I am blocked until this gets in

src/coverlet.core/Coverage.cs
tonerdo: No worries. I'll go ahead and merge this and effect the necessary changes. Thank you so much for all your work on this

PR #302, opened 2019-01-16
Guard instrumentation exclude Interlocked

ViktorHofer:
Excluding `System.Threading.Interlocked` from instrumentation which is used for recording hits. Not excluding that type would cause an infinite loop as it would call itself.

cc @sharwell @tonerdo


ViktorHofer:
@tonerdo it would be great if we could release a new package with this and https://github.com/tonerdo/coverlet/pull/276.

PR #303, opened 2019-01-16
Fix strong name validation failures

sharwell:
This is a prerequisite for coverlet.msbuild.tasks working on .NET Framework (#299).


tonerdo:
@sharwell I think #286 already took care of this. Did you notice anything missing?

ViktorHofer:
Why do we need the new coverlet templates assembly at all?

sharwell:
@ViktorHofer There will be two copies of coverlet.templates in the NuGet package one way or the other. There doesn't seem to be any drawback to multi-targeting this assembly according to the runtime support.

ViktorHofer:
@tonerdo this is a simple change which resolves the long standing strong name validation errors that we were hitting in coverlet plus it removes a dependency which always nice. Can we get this merged?

sharwell:
> ... plus it removes a dependency …

This was an undesirable side effect of a bug outside control of this repository. I'm going to file a follow-up issue to switch back to using the dependency once the source bug has been resolved.

PR #308, opened 2019-01-17
Avoid unnecessary null checks in RecordHit

### LINKED ISSUE #305: Avoid null check for RecordHit
The null check in `RecordHit` is only required for corelib instrumentation. Other libraries do not need this check, and it can be omitted to improve performance.

sharwell:
Fixes #305 

The performance benefit of this change is minimal in debug builds, but could be observable in other cases. Either way, an unnecessary null check is removed from the hot path.

### DIFF
--- a/src/coverlet.core/Instrumentation/Instrumenter.cs
+++ b/src/coverlet.core/Instrumentation/Instrumenter.cs
@@ -23,6 +23,7 @@ namespace Coverlet.Core.Instrumentation
         private readonly string[] _includeFilters;
         private readonly string[] _excludedFiles;
         private readonly string[] _excludedAttributes;
+        private readonly bool _isCoreLibrary;
         private InstrumenterResult _result;
         private FieldDefinition _customTrackerHitsFilePath;
         private FieldDefinition _customTrackerHitsArray;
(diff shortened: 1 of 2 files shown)


src/coverlet.core/Instrumentation/Instrumenter.cs
```diff
@@ -43,6 +43,14 @@ public Instrumenter(string module, string identifier, string[] excludeFilters, s
             _excludedAttributes = excludedAttributes;
         }
 
+        private bool IsCoreLibrary
```
ViktorHofer:
```suggestion
        private bool IsCoreLibrary { get; } = Path.GetFileNameWithoutExtension(_module) == "System.Private.CoreLib";
```

sharwell: Won't that throw a `NullReferenceException` at the beginning of the constructor?

src/coverlet.core/Instrumentation/Instrumenter.cs
```diff
@@ -41,8 +41,12 @@ public Instrumenter(string module, string identifier, string[] excludeFilters, s
             _includeFilters = includeFilters;
             _excludedFiles = excludedFiles ?? Array.Empty<string>();
             _excludedAttributes = excludedAttributes;
+
+            IsCoreLibrary = Path.GetFileNameWithoutExtension(_module) == "System.Private.CoreLib";
         }
 
+        private bool IsCoreLibrary { get; }
```
tonerdo: Why not make this a field?

PR #309, opened 2019-01-17
Single hit

### LINKED ISSUE #306: Special case hit threshold=1
For projects that do not care about the number of times a hit is recorded, the `RecordHit` method can be modified to avoid the read/write dependency for performance, and simply use this:

```csharp
HitsArray[index] = 1;
```

sharwell:
~~📝 This pull request builds on #308~~

This pull request adds the `<CoverletSingleHit>` option (`--single-hit` for the global tool), which can be used to improve the performance by only recording a single hit at each location. Where #291 brought the time down to 15.5 seconds, this pull request further reduces the execution time to ~10.3 seconds.

Fixes #306

### DIFF
--- a/src/coverlet.core/Instrumentation/Instrumenter.cs
+++ b/src/coverlet.core/Instrumentation/Instrumenter.cs
@@ -23,17 +23,19 @@ namespace Coverlet.Core.Instrumentation
         private readonly string[] _includeFilters;
         private readonly string[] _excludedFiles;
         private readonly string[] _excludedAttributes;
+        private readonly bool _singleHit;
         private readonly bool _isCoreLibrary;
         private InstrumenterResult _result;
         private FieldDefinition _customTrackerHitsArray;
         private FieldDefinition _customTrackerHitsFilePath;
+        private FieldDefinition _customTrackerSingleHit;
         private ILProcessor _customTrackerClassConstructorIl;
         private TypeDefinition _customTrackerTypeDef;
         private MethodReference _customTrackerRegisterUnloadEventsMethod;
         private MethodReference _customTrackerRecordHitMethod;
         private List<string> _asyncMachineStateMethod;
 
-        public Instrumenter(string module, string identifier, string[] excludeFilters, string[] includeFilters, string[] excludedFiles, string[] excludedAttributes)
+        public Instrumenter(string module, string identifier, string[] excludeFilters, string[] includeFilters, string[] excludedFiles, string[] excludedAttributes, bool singleHit)
         {
             _module = module;
             _identifier = identifier;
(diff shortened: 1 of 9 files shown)


src/coverlet.template/ModuleTrackerTemplate.cs
```diff
@@ -52,6 +52,30 @@ public static void RecordHit(int hitLocationIndex)
...
+            ref int location = ref HitsArray[hitLocationIndex];
+            if (location == 0)
+                location = 1;
+        }
+
+        public static void RecordSingleHit(int hitLocationIndex)
+        {
+            ref int location = ref HitsArray[hitLocationIndex];
+            if (location == 0)
+                location = 1;
```
sharwell: 📝 The main value here is avoiding this write invalidating cache lines. OpenCover uses a configurable threshold value. I could rewrite it to use a configurable threshold, though most uses I've seen in practice either omit this value or set it to 1.

src/coverlet.core/Instrumentation/Instrumenter.cs
```diff
@@ -23,24 +23,35 @@ internal class Instrumenter
...
             _excludedFiles = excludedFiles ?? Array.Empty<string>();
             _excludedAttributes = excludedAttributes;
+            _singleHit = singleHit;
+        }
+
+        private bool IsCoreLibrary
+        {
+            get
+            {
+                return Path.GetFileNameWithoutExtension(_module) == "System.Private.CoreLib";
```
MarcoRossignoli: this will be cached by https://github.com/tonerdo/coverlet/pull/308/files right?I mean is this tmp code?

src/coverlet.template/ModuleTrackerTemplate.cs
MarcoRossignoli: Why don't pass `_singleHit` as parameter or use static var and have less methods?Inlining?

sharwell: It's always a possibility to come back and do this if it proves to be equally fast. Since the expanded form is at least that fast and not difficult to work with, it seemed like the straightforward initial approach.

src/coverlet.core/Instrumentation/Instrumenter.cs
sharwell: This pull request builds on #308. This change is handled in the other pull request.

sharwell:
@tonerdo Note that OpenCover uses [`-threshold`](https://github.com/opencover/opencover/wiki/Usage) for this feature. I didn't retain the same name because it appears Coverlet uses that term with a different meaning, and I didn't want to confuse things.

tonerdo:
@sharwell @petli @MarcoRossignoli What are your thoughts on making this option an environment variable? It seems like an optimization detail that shouldn't be at the forefront of the primary options, kinda like how we can turn on/off tiered compilation in .NET Core

MarcoRossignoli:
I'm for parameter `--single-hit` my impression is that this parameter will be used a lot and hiding it's a shame, I could be wrong but the vast majority of coverage usage is check final percentage of coverage, and in case of issue people check how many tests(hits) cover particular code section. I don't know how many teams check hits amount regularly.

sharwell:
@tonerdo I lean towards what @MarcoRossignoli said, but I'm happy to implement it either way as long as I can use it.

MarcoRossignoli:
Not strictly related only to this PR,but in general, don't you think that we've too few unit/perf testing for these core part?

Issue #312, opened 2019-01-19
Sourcelink filter


ViktorHofer:
We need a filter to not use the source link argument for all assemblies, i.e. ones that we build ourselves (i.e. corefx assemblies) vs. ones that we not build but consume and still need coverage (i.e. System.Private.CoreLib). There are two options:

- Rename the existing argument `--use-source-link` to `--source-link-filter` which accepts a list of assembly names or `all`.
- Add an additional argument `--source-link-filter` which accepts a lit of assembly names which is only used if the option `--use-source-link` is provided.

@tonerdo what's your preference here?

ViktorHofer:
Reopening as this is now pops up again. We want to use the source link information embedded in the assemblies/pdbs and write it into the report but only for non live building assets like System.Private.CoreLib.

cc @MarcoRossignoli

MarcoRossignoli:
I prefer non breaking changes approach so new filter args works well to me cc: @tonerdo

MarcoRossignoli:
@ViktorHofer is this feature useful again?If so do you need it for .net tool driver?

ViktorHofer:
It's not very important for us anymore as we are moving assemblies together.

MarcoRossignoli:
Ok feel free to close!

PR #313, opened 2019-01-19
Add source link filter & code cleanup

ViktorHofer:
Fixes https://github.com/tonerdo/coverlet/issues/312

Also applied code cleanup suggestions from VS in files that I touched.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,7 @@ Options:
   --exclude-by-attribute  Attributes to exclude from code coverage.
   --merge-with            Path to existing coverage result to merge.
   --use-source-link       Specifies whether to use SourceLink URIs in place of file system paths.
+  --source-link-filter    Filter assemblies to probe for source link information.
 ```
 
 #### Code Coverage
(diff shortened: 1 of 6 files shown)


tonerdo:
I'm quite wary of adding too many options for every single edge case. What's the specific reason for this? 

If it's purely to allow choice, then we should consider introducing assembly level attributes that can override specific global coverlet options. Like this source link issue for example

For situations where the source link flag is set, but assemblies you don't own don't have source link information (which I think is the case here). The best solution is simply to update the code to fallback to regular debug information if there's no source link data

ViktorHofer:
I understand your concerns, I don't many options either.

We live build some assemblies in our product and some we just consume. Hence the paths in the pdb of the consumed assemblies can't be used as they point to CI machine's local paths. Currently coverlet only allows using source link for either all or none assemblies. We just want to use source link for the assemblies that we don't live build.

What about if we change the existing option `--use-source-link` to `--source-link-filter` which allows either `all` or a list of assemblies (like in my PR). With that we would keep the number of options sane. I don't think that somebody already relies on the added source link feature so I wouldn't think about a breaking change here.

I don't think attributes on an assembly level is a good idea here as we would need to put them on source assemblies which we ship.

tonerdo:
> For situations where the source link flag is set, but assemblies you don't own don't have source link information (which I think is the case here). The best solution is simply to update the code to fallback to regular debug information if there's no source link data

@ViktorHofer, how about this bit? Do you think it helps your use case?

ViktorHofer:
It's exactly the other way around. We want source link data not to be used for assemblies that we own/build (which in fact do have source link data in them).

Maybe I should question myself why we add source link information when building locally. Let me check with my team and get back to you.

ViktorHofer:
Ok this can be closed as we will just disable sourcelink completely for local builds.

PR #318, opened 2019-01-24
Fix failure to cover test code

sharwell:
Filtering of test modules is a per-project policy applied during report creation. It should not be assumed or applied as a default configuration.

Users who rely on the current behavior for performance reasons may exclude the test modules via instrumentation properties, though this might not be required since other recent performance improvements in coverlet would be adopted at the same time an upgrade caused this change to take effect.

### DIFF
--- a/test/coverlet.core.tests/Samples/Samples.cs
+++ b/test/coverlet.core.tests/Samples/Samples.cs
@@ -122,6 +122,13 @@ namespace Coverlet.Core.Samples.Tests
             return value;
         }
 
+        /// <summary>
+        /// This method is used by a unit test that verifies the behavior of the instrumentation process on an assembly
+        /// which is not instrumented. Excluding this method from code coverage prevents the bytecode for this reference
+        /// method from getting modified prior to test execution so it retains its original form for the test. This is
+        /// not a problem for the test because the instrumentation process only runs on assemblies which have not
+        /// already been instrumented.
+        /// </summary>
         [ExcludeFromCodeCoverage]
         public void HasSimpleTaskWithLambda()
         {
--- a/test/coverlet.core.tests/Samples/Samples.cs
+++ b/test/coverlet.core.tests/Samples/Samples.cs
@@ -122,6 +122,7 @@ namespace Coverlet.Core.Samples.Tests
             return value;
         }
 
+        [ExcludeFromCodeCoverage]
         public void HasSimpleTaskWithLambda()
         {
             var t = new Task(() => { });
--- a/src/coverlet.core/Helpers/InstrumentationHelper.cs
+++ b/src/coverlet.core/Helpers/InstrumentationHelper.cs
@@ -82,12 +82,20 @@ namespace Coverlet.Core.Helpers
         public static void BackupOriginalModule(string module, string identifier)
         {
             var backupPath = GetBackupPath(module, identifier);
+            var backupSymbolPath = Path.ChangeExtension(backupPath, ".pdb");
             File.Copy(module, backupPath, true);
+
+            var symbolFile = Path.ChangeExtension(module, ".pdb");
+            if (File.Exists(symbolFile))
+            {
+                File.Copy(symbolFile, backupSymbolPath, true);
+            }
         }
 
         public static void RestoreOriginalModule(string module, string identifier)
         {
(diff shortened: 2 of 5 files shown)


codemzs:
@tonerdo We need this change as this is causing incorrect reporting in dotnet/machinelearning repo. I will appreciate if you can approve this soon and release a nuget.

sharwell:
@tonerdo Sorry, I built and ran tests within VS but forgot to run the tests again after a command line build. I'll fix the build issues and update this PR.

MarcoRossignoli:
> We need this change as this is causing incorrect reporting in dotnet/machinelearning repo

I'm corious,  do you check threshold also of test module?

sharwell:
@MarcoRossignoli The intent was to remove all hard-coded special handling of test modules. This is the responsibility of downstream operations, not the core instrumentation.

MarcoRossignoli:
> The intent was to remove all hard-coded special handling of test modules. 

I agree, maybe there was a real sample, only curiosity

ViktorHofer:
Why would you instrument a test assembly?

sharwell:
@ViktorHofer coverage of test assemblies has the same advantages as coverage of production assemblies for the purpose of guided code review (which is by far the most valuable feature of code review tooling).

ViktorHofer:
I'm not opposed to the feature but I don't like the breaking change.

sharwell:
@ViktorHofer It's not a breaking change so much as a bug fix. The current release fails to instrument the named input assembly to the instrumentation request. It does not exclude test code specifically (e.g. test helper assemblies referenced by the unit test assembly), leading to reports that contain incomplete information about tests.

test/coverlet.core.tests/Samples/Samples.cs
```diff
@@ -122,6 +122,7 @@ public string HasSimpleUsingStatement()
             return value;
         }
 
+        [ExcludeFromCodeCoverage]
```
MarcoRossignoli: I've got a doubt `GetBranchPoints_GeneratedBranches_DueToCachedAnonymousMethodDelegate_Ignored` is again the reason for `Empty`?

sharwell: If this method is instrumented, it no longer detects the pattern that would cause the branch to be ignored. Previously, the lack of instrumentation was caused by the method being located in the test assembly. The current behavior would only be a problem if a user attempted to instrument an assembly that is already instrumented, so it seemed safe to just make this small change that prevents the instrumentation process from making changes to this method.

MarcoRossignoli: Clear thank's

MarcoRossignoli: Could be useful thank's a lot

tonerdo:
I don't think this is a breaking change. All this will do is include the test assembly coverage in the final coverage output. Basically like when I added support for embedded PDBs and xunit assemblies started showing up in the results.

I wonder if we should put this behind a flag (`IncludeTestAssembly`?) or have the wrapper programs (msbuild and global tool) exclude it by default but make it overridable with the existing `exclude` or `include` variants.

Thoughts? @MarcoRossignoli @ViktorHofer @sharwell

sharwell:
The main problem with `<IncludeTestAssembly>` triggering the *current* exclusion behavior is it doesn't correctly handle test code in multiple assemblies. For example, if a test project has two project references: a production project and a different test project, the referenced test project will appear in the coverage results even though it's not production code. IMO the best way to exclude test assemblies if users wish to do so is the existing assembly and/or path-based filtering.

MarcoRossignoli:
Personally I wouldn't add another flag, @sharwell sample is pertinent, this is slight change in behaviour, should be advertised(on release note), with this change coverlet will be more useful(https://github.com/tonerdo/coverlet/pull/318#issuecomment-457223596) add test assembly exclusion to scripts doesn't seem a great problem, if this will break something...well we'll fix it. I'd like to listen @ViktorHofer's final words...good opinions are never enough.

ViktorHofer:
I agree with others that we shouldn't introduce an option for that but definitely outline it in the release notes. Giving it a second thought I agree with Sam that the current behavior is the broken one and this change is mere a fix to it.

PR #322, opened 2019-01-24
Revert "Merge pull request #276 from petli/memory-mapped-hit-counts"

ViktorHofer:
This reverts commit 0d285b187a445bbbe50a451c116c003dd509134b

Reasoning: https://github.com/tonerdo/coverlet/pull/276#issuecomment-457281295

cc @petli @sharwell @tonerdo

### DIFF
--- a/test/coverlet.core.tests/Instrumentation/ModuleTrackerTemplateTests.cs
+++ b/test/coverlet.core.tests/Instrumentation/ModuleTrackerTemplateTests.cs
@@ -1,9 +1,6 @@
 ﻿using Coverlet.Core.Instrumentation;
-using Coverlet.Core.Helpers;
 using System;
-using System.Collections.Generic;
 using System.IO;
-using System.IO.MemoryMappedFiles;
 using System.Threading;
 using System.Threading.Tasks;
 using Xunit;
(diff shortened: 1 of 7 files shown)


test/coverlet.core.tests/Instrumentation/ModuleTrackerTemplateTests.cs
```diff
@@ -12,72 +9,62 @@ namespace Coverlet.Core.Tests.Instrumentation
...
+        [Fact(Skip="Failed CI Job: https://ci.appveyor.com/project/tonerdo/coverlet/builds/21145989/job/9gx5jnjs502vy1fv")]
         public void HitsOnMultipleThreadsCorrectlyCounted()
         {
-            for (int i = 0; i < 4; ++i)
+            ModuleTrackerTemplate.HitsArray = new[] { 0, 0, 0, 0 };
+            for (int i = 0; i < ModuleTrackerTemplate.HitsArray.Length; ++i)
             {
                 var t = new Thread(HitIndex);
                 t.Start(i);
-                t.Join();
```
petli: This line should be kept, because not waiting for the test threads to finish running makes the test unreliable. Then the Skip on the test can probably be removed.

MarcoRossignoli: Maybe better track with issue and fix with a PR after merge.

sharwell: I generally prefer to do a full revert if possible, and then re-apply the part(s) that are still desired, but I wouldn't be opposed to either approach.

petli: Agreed, that is better. Optionally, the code this test is testing is no longer in the code base after changing to Interlocked.Increment in RecordHit so it could be removed altogether.

PR #326, opened 2019-01-25
Expand MergeWith documentation

### LINKED ISSUE #325: MergeWith should create the file, or fail if its missing
Currently testing out the MergeWith option. It seems that it neither fails the test, logs a warning or creates the target file - if it's missing. This seems like a bug to me. It simply silently ignores writing any output.

Workarounds:

* Place an empty JSON Object in the target file (`{}` in the target file)
* Run first test without Merge, then the remaining ones with Merge

LordMike:
Fixes #325

### DIFF
--- a/README.md
+++ b/README.md
@@ -304,7 +304,7 @@ With Coverlet you can combine the output of multiple coverage runs into a single
 dotnet test /p:CollectCoverage=true /p:MergeWith='/path/to/result.json'
 ```
 
-The value given to `/p:MergeWith` **must** be a path to Coverlet's own json result format.
+The value given to `/p:MergeWith` **must** be a path to Coverlet's own json result format. The results in the file will be read, and added to the new results written to by coverlet.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -304,7 +304,7 @@ With Coverlet you can combine the output of multiple coverage runs into a single
 dotnet test /p:CollectCoverage=true /p:MergeWith='/path/to/result.json'
 ```
 
-The value given to `/p:MergeWith` **must** be a path to Coverlet's own json result format.
+The value given to `/p:MergeWith` **must** be a path to Coverlet's own json result format. The results in the file will be read, and added to the new results written to by Coverlet.
```
tonerdo:
```suggestion
The value given to `/p:MergeWith` **must** be a path to Coverlet's own json result format. The results in `result.json` will be read, and added to the new results written to by Coverlet.
```

PR #330, opened 2019-01-30
Remove unused solution platforms

sharwell:


tonerdo:
I think this is a visual studio thing. I envisage they will return in subsequent PRs. Is there some editor config like settings file for Visual Studio that allows us to lock the Solution file config?

ViktorHofer:
Please somebody merge :D

PR #341, opened 2019-02-14
Added logging to Coverage to help with debugging

derhally:
There was no indication why coverage didn't occur because the exception occurring inside Coverage class was being swallowed.

I was missing an dependency and had no way of knowing that.  This should help.


tonerdo:
@derhally thanks a lot for this! Will take a look later today

src/coverlet.core/Logging/ILogger.cs
```diff
@@ -1,11 +1,15 @@
-namespace Coverlet.Console.Logging
+namespace Coverlet.Core
```
petli: Minor namespace nagging: should be `Coverlet.Core.Logging` to be consistent with directory structure.

src/coverlet.core/Logging/NullLogger.cs
```diff
@@ -0,0 +1,29 @@
+﻿using Coverlet.Core;
+
+namespace coverlet.console.Logging
```
petli: More minor namespace nagging: `Coverlet.Core.Logging`

src/coverlet.core/Logging/ILogger.cs
derhally: Perfectly acceptable.  Fixed

MarcoRossignoli:
@derhally CI fail on `MSBuildLogger.cs` for wrong namespace `Coverlet.Core.ILogger`

src/coverlet.msbuild.tasks/InstrumentationTask.cs
```diff
@@ -88,7 +88,7 @@ public override bool Execute()
                 var excludedSourceFiles = _excludeByFile?.Split(',');
                 var excludeAttributes = _excludeByAttribute?.Split(',');
 
-                _coverage = new Coverage(_path, includeFilters, includeDirectories, excludeFilters, excludedSourceFiles, excludeAttributes, _singleHit, _mergeWith, _useSourceLink);
+                _coverage = new Coverage(_path, includeFilters, includeDirectories, excludeFilters, excludedSourceFiles, excludeAttributes, _singleHit, _mergeWith, _useSourceLink, new MSBuildLogger(Log));
```
MarcoRossignoli: We should use field `ILogger _logger` on top like other

src/coverlet.msbuild.tasks/MSBuildLogger.cs
```diff
@@ -0,0 +1,41 @@
+﻿using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using ILogger = Coverlet.Core.ILogger;
+
+namespace Coverlet.MSbuild.Tasks
+{
+    class MSBuildLogger : ILogger
```
MarcoRossignoli: nit: for this simple wrapper to msbuild logger you should use Expression-bodied members for better readability https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/statements-expressions-operators/expression-bodied-members

src/coverlet.msbuild.tasks/MSBuildLogger.cs
```diff
@@ -0,0 +1,41 @@
...
+        }
+
+        public void LogSuccess(string message)
+        {
+            LogInformation(message);
+        }
+
+        public void LogVerbose(string message)
+        {
+            _log.LogMessageFromText(message, MessageImportance.Low);
```
MarcoRossignoli: maybe this(and LogInformation) should be [TaskLoggingHelper.LogMessage](https://docs.microsoft.com/en-us/dotnet/api/microsoft.build.utilities.tasklogginghelper.logmessage?view=netframework-4.7.2) on guide [`LogMessageFromText`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.build.utilities.tasklogginghelper.logmessagefromtext?view=netframework-4.7.2) is used for `Logs an error message or warning from the given line of text.` but I'm not so fluent with this part of msbuild so maybe it's ok.

src/coverlet.msbuild.tasks/MSBuildLogger.cs
```diff
@@ -0,0 +1,41 @@
...
+    class MSBuildLogger : ILogger
+    {
+        private readonly TaskLoggingHelper _log;
+
+        public MSBuildLogger(TaskLoggingHelper _log)
+        {
+            this._log = _log;
+        }
+
+        public void LogSuccess(string message)
```
MarcoRossignoli: @tonerdo I don't see any usage of this method https://github.com/tonerdo/coverlet/search?q=LogSuccess&unscoped_q=LogSuccess could be removed?

src/coverlet.console/Program.cs
```diff
@@ -59,8 +60,12 @@ static int Main(string[] args)
...
                 process.StartInfo.RedirectStandardError = true;
                 process.Start();
-                logger.LogInformation(process.StandardOutput.ReadToEnd());
-                logger.LogError(process.StandardError.ReadToEnd());
+
+                process.StandardOutput.ReadToEndAsync()
+                    .ContinueWith(task => logger.LogInformation(task.Result), TaskContinuationOptions.OnlyOnRanToCompletion);
+                process.StandardError.ReadToEndAsync()
+                    .ContinueWith(task => logger.LogError(task.Result), TaskContinuationOptions.OnlyOnRanToCompletion);
+
```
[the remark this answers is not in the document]
derhally: I never noticed those events.  Interesting

src/coverlet.core/Logging/NullLogger.cs
```diff
@@ -0,0 +1,28 @@
+﻿
+namespace Coverlet.Core.Logging
+{
+    public sealed class NullLogger :ILogger
```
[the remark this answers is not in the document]
derhally: Replaced with a mock object

src/coverlet.core/Logging/ILogger.cs
```diff
@@ -1,11 +1,19 @@
-namespace Coverlet.Console.Logging
+using System;
+
+namespace Coverlet.Core.Logging
 {
-    interface ILogger
+    public interface ILogger
     {
         void LogSuccess(string message);
+
```
MarcoRossignoli: please revert this new line

src/coverlet.core/Coverage.cs
```diff
@@ -25,14 +24,24 @@ public class Coverage
         private bool _singleHit;
         private string _mergeWith;
         private bool _useSourceLink;
+        private readonly ILogger _logger;
```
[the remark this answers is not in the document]
MarcoRossignoli: oh sorry my fault I wasn't clear, I prefer maitain old style for scope of this PR so refert readonly and don't use readonly for new _logger, sorry again :sweat:

src/coverlet.msbuild.tasks/InstrumentationTask.cs
```diff
@@ -17,6 +17,7 @@ public class InstrumentationTask : Task
         private bool _singleHit;
         private string _mergeWith;
         private bool _useSourceLink;
+        private readonly MSBuildLogger _logger;
```
MarcoRossignoli: remove `readonly` for now, maybe on some next PR we should change for in all codebase

src/coverlet.msbuild.tasks/MSBuildLogger.cs
```diff
@@ -0,0 +1,29 @@
...
+using Microsoft.Build.Utilities;
+using ILogger = Coverlet.Core.Logging.ILogger;
+
+namespace Coverlet.MSbuild.Tasks
+{
+    class MSBuildLogger : ILogger
+    {
+        private readonly TaskLoggingHelper _log;
+
+        public MSBuildLogger(TaskLoggingHelper log)
```
MarcoRossignoli: you can simplify also constructor `public MSBuildLogger(TaskLogginHelper log) => _log = log;`

src/coverlet.core/Coverage.cs
```diff
@@ -72,9 +82,10 @@ public void PrepareModules()
                         var result = instrumenter.Instrument();
                         _results.Add(result);
                     }
-                    catch (Exception)
+                    catch (Exception ex)
                     {
                         // TODO: With verbose logging we should note that instrumentation failed.
```
MarcoRossignoli: I think we can remove comment with this change

src/coverlet.msbuild.tasks/InstrumentationTask.cs
MarcoRossignoli: here I agree first ctor param

src/coverlet.console/Logging/ConsoleLogger.cs
```diff
@@ -1,39 +1,39 @@
...
-        public void LogWarning(string message)
+        private static void WriteLine(string message, ConsoleColor color  = ConsoleColor.White)
         {
-            ForegroundColor = ConsoleColor.Yellow;
-            WriteLine(message);
-            ForegroundColor = ConsoleColor.White;
+            ConsoleColor currentForegroundColor;
+            if (color != (currentForegroundColor = SystemConsole.ForegroundColor))
+            {
+                lock (_sync)
```
MarcoRossignoli: nit: lock outermost
```cs
WriteLine(string message, ConsoleColor color)
{
     lock(_sync)
     {
       ForegroundColor currentForegroundColor;
       if (color != (currentForegroundColor = ForegroundColor))
       {
          ForegroundColor = color;
          WriteLine(message);
          ForegroundColor = currentForegroundColor;       
       }
       else
       {
           WriteLine(message);
       }
    }
}
```

src/coverlet.msbuild.tasks/MSBuildLogger.cs
```diff
@@ -0,0 +1,24 @@
...
+
+        public void LogVerbose(string message) => _log.LogMessageFromText(message, MessageImportance.Low);
+
+        public void LogInformation(string message)=> _log.LogMessage(message);
+
+        public void LogWarning(string message) => _log.LogWarning(message);
+
+        public void LogError(string message) => _log.LogError(message);
+      
+        public void LogError(string message, Exception exception) => _log.LogErrorFromException(exception);
```
MarcoRossignoli: better remove message here, I think it's better have two distinct method one with message and one with Exception what do you think?Here we cannot use message so or we call 2 method on task logger or remove first param

src/coverlet.console/Logging/ConsoleLogger.cs
```diff
@@ -1,39 +1,39 @@
...
-            WriteLine(message);
-            ForegroundColor = ConsoleColor.White;
-        }
+        public void LogError(string message, Exception ex) => LogError($"{message}{Environment.NewLine}{ex}");
 
-        public void LogVerbose(string message)
-        {
-            throw new System.NotImplementedException();
-        }
+        public void LogInformation(string message) => WriteLine(message);
```
MarcoRossignoli: You could call same helper with current `ConsoleColor` to be consistent with other methods

MarcoRossignoli: never mind...defaul color,  missed that perfect

src/coverlet.msbuild.tasks/MSBuildLogger.cs
derhally: I usually prefer to provide a message with the exception to help provide a little more info in case the exception is cryptic.   I thought I'd leave it here if other parts wanted to use it down the road, but I can remove it if you wish.

MarcoRossignoli: I understand your idea, but maybe could be misleading, because if we don't know which logger we're using it's not clear why we pass a value that we'll not print on logger, I would remove

MarcoRossignoli: Ah you could try a "fake" exception to better understand how msbuild logger print this message, but this is an exception and I think that message+stack will be enough

src/coverlet.console/Logging/ConsoleLogger.cs
MarcoRossignoli: Because we can change "default" foreground color
image

src/coverlet.console/Logging/ConsoleLogger.cs
```diff
@@ -1,39 +1,39 @@
...
+
+        public void LogVerbose(string message) => throw new NotImplementedException();
+
+        public void LogWarning(string message) => WriteLine(message, ConsoleColor.Yellow);
 
-        public void LogWarning(string message)
+        private static void WriteLine(string message, ConsoleColor color = ConsoleColor.White)
```
MarcoRossignoli: We can avoid default color

src/coverlet.msbuild.tasks/MSBuildLogger.cs
```diff
@@ -0,0 +1,24 @@
...
+
+namespace Coverlet.MSbuild.Tasks
+{
+    class MSBuildLogger : ILogger
+    {
+        private readonly TaskLoggingHelper _log;
+
+        public MSBuildLogger(TaskLoggingHelper log) => _log = log;
+
+        public void LogVerbose(string message) => _log.LogMessageFromText(message, MessageImportance.Low);
```
MarcoRossignoli: `LogMessage(message,MessageImportance.Low)` ?

src/coverlet.core/Logging/ILogger.cs
```diff
@@ -1,11 +1,17 @@
...
+using System;
+
+namespace Coverlet.Core.Logging
 {
-    interface ILogger
+    public interface ILogger
     {
-        void LogSuccess(string message);
         void LogVerbose(string message);
+
```
MarcoRossignoli: nit: remove extra line, keep formatting like original pls

src/coverlet.msbuild.tasks/MSBuildLogger.cs
derhally: If the log level is verbose, MS Build logging equivalence is MessageImportance.Low.

MarcoRossignoli: Yes I mean that we should use `LogMessage` and not `LogMessageFromText` guide says that is used for error and warning and this is verbose info, right? `Logs an error message or warning from the given line of text.` https://docs.microsoft.com/en-us/dotnet/api/microsoft.build.utilities.tasklogginghelper.logmessagefromtext?redirectedfrom=MSDN&view=netframework-4.7.2#Microsoft_Build_Utilities_TaskLoggingHelper_LogMessageFromText_System_String_Microsoft_Build_Framework_MessageImportance_

derhally: Oh, I thought you were wondering why it was that.  Didn't notice you were suggesting to use LogMessage instead of LogMessageFromText

src/coverlet.console/Logging/ConsoleLogger.cs
```diff
@@ -1,39 +1,39 @@
 using System;
-using static System.Console;
+using Coverlet.Core.Logging;
+using SystemConsole = System.Console;
```
MarcoRossignoli: Is this necessary?I'm OOF and I cannot try with VS is there a name collision somewhere?

derhally: Before it was using static namespace.   Which causes a ambiguity with the locally defined WriteLine method.

src/coverlet.core/Coverage.cs
```diff
@@ -72,9 +82,10 @@ public void PrepareModules()
...
+                        _logger.LogWarning($"Unable to instrument module: {module} because : {ex.Message}");
```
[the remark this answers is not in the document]
tonerdo: You mean `.resx`files?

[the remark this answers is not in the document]
MarcoRossignoli: Yes

MarcoRossignoli:
@tonerdo Yes I'm OOF at the moment tomorrow I'll do last check and I'll do my approval if needed, I like this change a lot great thank's to @derhally!

src/coverlet.msbuild.tasks/MSBuildLogger.cs
```diff
@@ -0,0 +1,24 @@
...
+
+namespace Coverlet.MSbuild.Tasks
+{
+    class MSBuildLogger : ILogger
+    {
+        private readonly TaskLoggingHelper _log;
+
+        public MSBuildLogger(TaskLoggingHelper log) => _log = log;
+
+        public void LogVerbose(string message) => _log.LogMessage(message, MessageImportance.Low);
```
MarcoRossignoli: I think this is wrong, should be ` _log.LogMessage(MessageImportance.Low, message);` https://docs.microsoft.com/en-us/dotnet/api/microsoft.build.utilities.tasklogginghelper.logmessage?view=netframework-4.7.2&viewFallbackFrom=netcore-2.2#Microsoft_Build_Utilities_TaskLoggingHelper_LogMessage_Microsoft_Build_Framework_MessageImportance_System_String_System_Object___

MarcoRossignoli:
@tonerdo @derhally I did some tests I cannot show LogInformation/LogVerbose message, Warning and Exception are ok, do you have same issue?

derhally:
> @tonerdo @derhally I did some tests I cannot show LogInformation/LogVerbose(https://docs.microsoft.com/en-us/visualstudio/msbuild/obtaining-build-logs-with-msbuild?view=vs-2017#verbosity-settings) message, Warning and Exception are ok, do you have same issue?

So, I completely glossed over that I didn't switch CoverageResultTask to use the logger instead of console.  I fixed that.

It seems if you use "dotnet test" you need to specify `-v n` to see the coverage report and that's even with the MessageImportance set to High.   I'm not sure if that is the expected behavior.  I would think that people would want to see output by default.  This is very odd.

MarcoRossignoli:
> It seems if you use "dotnet test" you need to specify `-v n` to see the coverage report and that's even with the MessageImportance set to High. I'm not sure if that is the expected behavior. I would think that people would want to see output by default. This is very odd.

For now I think we should use `Console.WriteLine` and use logger only for "debugging" purpouse.
The odd thing is that if I try to add  for instance 
```
  _logger.LogInformation("TEST");
  _logger.LogWarning("WARNING");
```
and run build
```
 msbuild build.proj /verbosity:diagnostic
```
I get only warning notification
```
...
  Build completed. (TaskId:7)
   (TaskId:7)
D:\git\coverlet\build\Debug\coverlet.msbuild.targets(21,5): warning : WARNING [D:\git\coverlet\test\coverlet.core.tests\coverlet.core.tests.csproj] [D:\git\coverlet\build.proj]
  Test run for D:\git\coverlet\test\coverlet.core.tests\bin\Debug\netcoreapp2.0\coverlet.core.tests.dll(.NETCoreApp,Version=v2.0) (TaskId:7)
  Microsoft (R) Test Execution Command Line Tool Version 16.0.0-preview-20181205-02 (TaskId:7)
  Copyright (c) Microsoft Corporation.  All rights reserved. (TaskId:7)
   (TaskId:7)
...
```
I dunno why information is not shown @tonerdo do you have some idea, maybe you've more experience with msbuild.

derhally:
An option is to use Console for the Logger.Information and still use the MSBuild logger for Error and Warning so that those go through the  build system

MarcoRossignoli:
> An option is to use Console for the Logger.Information and still use the MSBuild logger for Error and Warning so that those go through the build system

Could be a breaking change, because until today coverlet users expect console output with percentace, if we use build system log a "quiet" verbosity will suppress all output.

src/coverlet.core/Logging/ILogger.cs
MarcoRossignoli: @derhally could you remove extraline to keep old formatting pls?

src/coverlet.msbuild.tasks/CoverageResultTask.cs
```diff
@@ -54,11 +55,16 @@ public string ThresholdStat
...
+        {
+            _logger = new MSBuildLogger(Log);
+        }
+
         public override bool Execute()
         {
             try
             {
-                Console.WriteLine("\nCalculating coverage result...");
+                _logger.LogInformation("Calculating coverage result...");
```
MarcoRossignoli: As I said for the moment I would revert this to old `Console.WriteLine`

src/coverlet.msbuild.tasks/MSBuildLogger.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    class MSBuildLogger : ILogger
+    {
+        private readonly TaskLoggingHelper _log;
+
+        public MSBuildLogger(TaskLoggingHelper log) => _log = log;
+
+        public void LogVerbose(string message) => _log.LogMessage(MessageImportance.Low, message);
+
+        public void LogInformation(string message)=> _log.LogMessage(MessageImportance.High, message);
```
MarcoRossignoli: WIth normal and `v:n` does it show?If so I would use `MessageImportance.Normal`

src/coverlet.msbuild.tasks/CoverageResultTask.cs
derhally: if LogInformation's implementation is Console.WriteLine then it would be the same as the previous behavior

src/coverlet.msbuild.tasks/MSBuildLogger.cs
derhally: you need to use the dotnet test switch `-v n` for it to show

src/coverlet.msbuild.tasks/CoverageResultTask.cs
MarcoRossignoli: Ok understood your point, personally I wouldn't mix infrastructure vs non infrastructure(mandatory console output) in same "interface".
I expect that when I'll pass v:quiet we won't show anything from this interface calls, but in this case we won't show for all but LogInformation usage.
I prefer another interface `IConsole.WriteLine` if we want make it testable and pluggable.
BTW this is only my opinion wait for @petli @tonerdo one.

src/coverlet.msbuild.tasks/MSBuildLogger.cs
MarcoRossignoli: ok but it works also with `MessageImportance.Normal`? I wonder if it's correct use `MessageImportance.High` and not `MessageImportance.Normal`.

derhally: Unfortunately Normal doesn't show up when using `-v n`.   This may be a bug in dotnet test driver.  I haven't had time to investigate this further.

MarcoRossignoli: Ok please add comment above like "We use `MessageImportance.High` because with `MessageImportance.Normal` doesn't show anything"

MarcoRossignoli:
@derhally are you able to complete this PR if not I could do last minor open changes with no problem(if you're busy, you already did a great work on this)!
@tonerdo I think that logging could help in some of open issues what do you think?I mean we should add verbose logging in some parts(inclusion exclusion filter, file instrumenting mainly).

derhally:
@MarcoRossignoli which track would you prefer, should the CovverageResultTask use 

1. Console.WriteLine as before 
2. Use ILogger api and change MSBuildLogger implementation to log to console?

Either way, the behavior would be same as before.

MarcoRossignoli:
* I would use Console.WriteLine as before because that message is not an infrastructure message, I mean we want to see that message with or without swich applied(verbose etc...)
* And use msbuild logger on LogInformation with comment https://github.com/tonerdo/coverlet/pull/341#discussion_r257469931

src/coverlet.core/Logging/ILogger.cs
MarcoRossignoli: @derhally can you revert this formatting(new lines)?I would see only green line on class visibility and new LogError method. We try to not change formatting of old files, it's to easy review and in case found bug/issue on updates.

src/coverlet.msbuild.tasks/CoverageResultTask.cs
```diff
@@ -54,11 +55,16 @@ public string ThresholdStat
...
+        public CoverageResultTask()
+        {
+            _logger = new MSBuildLogger(Log);
+        }
+
         public override bool Execute()
         {
             try
             {
-                Console.WriteLine("\nCalculating coverage result...");
```
MarcoRossignoli: We should not change old output formatting `\n' here and also on other(space before text)

src/coverlet.msbuild.tasks/CoverageResultTask.cs
```diff
@@ -85,7 +91,7 @@ public override bool Execute()
                     if (reporter.OutputType == ReporterOutputType.Console)
                     {
                         // Output to console
-                        Console.WriteLine("  Outputting results to console");
```
MarcoRossignoli: ditto

src/coverlet.msbuild.tasks/CoverageResultTask.cs
```diff
@@ -96,7 +102,7 @@ public override bool Execute()
                         filename = Path.HasExtension(filename) ? filename : $"{filename}.{reporter.Extension}";
 
                         var report = Path.Combine(directory, filename);
-                        Console.WriteLine($"  Generating report '{report}'");
```
MarcoRossignoli: ditto

src/coverlet.msbuild.tasks/CoverageResultTask.cs
```diff
@@ -146,7 +152,6 @@ public override bool Execute()
                     coverageTable.AddRow(Path.GetFileNameWithoutExtension(module.Key), $"{linePercent}%", $"{branchPercent}%", $"{methodPercent}%");
                 }
 
-                Console.WriteLine();
```
MarcoRossignoli: ditto

tonerdo:
> @tonerdo I think that logging could help in some of open issues what do you think?I mean we should add verbose logging in some parts(inclusion exclusion filter, file instrumenting mainly).

@MarcoRossignoli  totally agree. I'm thinking we could add a `diag` (or similar) flag to turn on verbose logging for debugging purposes

MarcoRossignoli:
> @MarcoRossignoli totally agree. I'm thinking we could add a `diag` (or similar) flag to turn on verbose logging for debugging purposes

Seems that for now we can use `dotnet test /p:CollectCoverage=true -v:n` https://github.com/tonerdo/coverlet/pull/341#issuecomment-464355847 I did some test with:
`dotnet test "C:\git\coverlet\test\coverlet.core.tests\coverlet.core.tests.csproj" -c Debug /p:CollectCoverage=true /p:CoverletOutputFormat=opencover /p:Include=[coverlet.*]* -v:n`

image
image

The worste part is that we get output also for others, but better than nothing for now.

Issue #350, opened 2019-02-21
Error code on merged results despite aggregate being fine


Vassi:
I couldn't quite find an issue of this, but I apologize if I just missed it.

I run coverlet across three different test projects using the Merge option to get a total code coverage report at the end. My issue is that on each test project the coverage on the assemblies varies so, individually, some may fail my code coverage threshold but in the aggregate it's fine. Still, coverlet errors on the individual so my CI fails my build erroneously (in my opinion).

Below please note how the first two steps of testing have line or method under 80, but the final is listed as well above 90%. Not sure what I can do?

Example:
```
Total tests: 13. Passed: 13. Failed: 0. Skipped: 0.
Test Run Successful.
Test execution time: 2.4824 Seconds

Calculating coverage result...
  Generating report '../../coverage/coverage.json'
  Generating report '../../coverage/coverage.info'

+----------------------------+--------+--------+--------+
| Module                     | Line   | Branch | Method |
+----------------------------+--------+--------+--------+
| ACT.AssumptionLibrary.Data | 40.7%  | 95.5%  | 13.5%  |
+----------------------------+--------+--------+--------+

+---------+--------+--------+--------+
|         | Line   | Branch | Method |
+---------+--------+--------+--------+
| Total   | 40.7%  | 95.5%  | 13.5%  |
+---------+--------+--------+--------+
| Average | 40.7%  | 95.5%  | 13.5%  |
+---------+--------+--------+--------+

/root/.nuget/packages/coverlet.msbuild/2.5.1/build/netstandard2.0/coverlet.msbuild.targets(33,5): error : The minimum line coverage is below the specified 80 

... trim ...

Calculating coverage result...
  Generating report '../../coverage/coverage.json'
  Generating report '../../coverage/coverage.info'

+------------------------------------------+--------+--------+--------+
| Module                                   | Line   | Branch | Method |
+------------------------------------------+--------+--------+--------+
| ACT.AssumptionLibrary.Data               | 91.7%  | 95.5%  | 82.7%  |
+------------------------------------------+--------+--------+--------+
| ACT.AssumptionLibrary.Services.Contracts | 85.7%  | 100%   | 66.7%  |
+------------------------------------------+--------+--------+--------+
| ACT.AssumptionLibrary.Services           | 100%   | 100%   | 100%   |
+------------------------------------------+--------+--------+--------+

+---------+--------+-------------------+--------+
|         | Line   | Branch            | Method |
+---------+--------+-------------------+--------+
[...]

PR #351, opened 2019-02-21
when "dotnet test" runs test projects in parallel mode, json file writes invalid

loning:
dotnet test --no-build /p:CollectCoverage=true /p:CoverletOutputFormat='json%2copencover' /p:CoverletOutput="../results/coverage" /p:MergeWith="../results/coverage.json" /p:Exclude="[coverlet.*.tests?]*"

../results/coverage.json may faces concurrency problem, so when making the file share mode be none, it will solve the problem


src/coverlet.core/Coverage.cs
```diff
@@ -32,7 +30,9 @@ public string Identifier
             get { return _identifier; }
         }
 
-        public Coverage(string module, string[] includeFilters, string[] includeDirectories, string[] excludeFilters, string[] excludedSourceFiles, string[] excludeAttributes, bool singleHit, string mergeWith, bool useSourceLink)
+        public Coverage(string module, string[] includeFilters, string[] includeDirectories, string[] excludeFilters,
```
MarcoRossignoli: Please could you revert formatting to old one here and on other places?

src/coverlet.core/Coverage.cs
```diff
@@ -170,11 +183,22 @@ public CoverageResult GetCoverageResult()
...
+                var json = RetryHelper.Do(() =>
+                {
+                    using (var file = File.Open(_mergeWith, FileMode.Open, FileAccess.Read, FileShare.None))
+                    {
+                        using (var reader = new StreamReader(file))
+                        {
+                            return reader.ReadToEnd();
+                        }
+                    }
+                }, () => TimeSpan.FromMilliseconds(100), 5);
```
MarcoRossignoli: I'm not sure about max attempt, I would try for "more standard seconds" 10 or 30.

src/coverlet.core/Coverage.cs
```diff
@@ -170,11 +183,22 @@ public CoverageResult GetCoverageResult()
...
             }
 
-            var coverageResult = new CoverageResult { Identifier = _identifier, Modules = modules, InstrumentedResults = _results };
+            var coverageResult = new CoverageResult
+                {Identifier = _identifier, Modules = modules, InstrumentedResults = _results};
 
             if (!string.IsNullOrEmpty(_mergeWith) && !string.IsNullOrWhiteSpace(_mergeWith) && File.Exists(_mergeWith))
             {
-                string json = File.ReadAllText(_mergeWith);
+                var json = RetryHelper.Do(() =>
```
MarcoRossignoli: Please can you add some comment above like `// Possible concurrency access to merge file between write/read on parallel testing`  here and below?

tonerdo:
@loning I'm not very clear the problem this PR is trying to solve

loning:
I have lots of unit test project under one solution, and when we call "dotnet test any.sln", the unit test project will run in parallel at the same time, when we use /p:CoverletOutput="../results/coverage" /p:MergeWith="../results/coverage.json", may two project will write to ../results/coverage.json at the same time, and also when other project want to merge, it may read a file not completed.

so my solution is to make the file not shared, and when we cannot access a file, we wait 100ms, cause 100ms is enough for another read or write

https://travis-ci.org/AElfProject/AElf/builds/497464201?utm_source=github_status&utm_medium=notification

loning:
I just found once error, and rerun the CI, disappeared. but I thought it's not hard to fix it

MarcoRossignoli:
> and when we call "dotnet test any.sln",

@loning run coverlet solution wide is not a supported scenario for now https://github.com/tonerdo/coverlet/issues/350#issuecomment-466670698

loning:
https://github.com/tonerdo/coverlet/pull/351#issuecomment-466682952

thanks，@MarcoRossignoli，but it seems worked well in our solution, we do not want to use shell to run test one by one

MarcoRossignoli:
@loning for the moment I'll close this stale PR.
Even if you idea works well we should find a more robust way, we've an open issue on it https://github.com/tonerdo/coverlet/issues/357

Issue #364, opened 2019-03-08
MSBuild data should not be passed between tasks via static fields


sharwell:
The following static field access should be rewritten using MSBuild inputs and outputs:

https://github.com/tonerdo/coverlet/blob/062b907735901bdf3ad9d4d40953f53a4fe3f364/src/coverlet.msbuild.tasks/CoverageResultTask.cs#L69

https://github.com/tonerdo/coverlet/blob/062b907735901bdf3ad9d4d40953f53a4fe3f364/src/coverlet.msbuild.tasks/InstrumentationTask.cs#L10

I believe the most straightforward and reliable implementation would involve generating GUIDs and/or temporary file names within MSBuild, and then passing those names to the build tasks. With this approach, the instrumentation task would produce known outputs for a given set of inputs, which can be tracked by the MSBuild targets.

📝 This limitation forces test execution to use `/m:1` since the static property cannot keep track of information from multiple test executions in parallel.

MarcoRossignoli:
Fixed in #409

PR #366, opened 2019-03-08
Fix failure to process coverage reports for parallel builds

sharwell:
This is a workaround for #364. Eventually this code should be updated to avoid the state sharing altogether.


tonerdo:
@sharwell Can you please look into the test failures?

MarcoRossignoli:
fixing in https://github.com/tonerdo/coverlet/pull/393

MarcoRossignoli:
Fixed in #409

PR #367, opened 2019-03-12
fix: catch exception

ViceIce:
This pr catches exceptions caused by invalid pdb's.

Eg [Oracle.ManagedDataAccess 18.3.0](https://www.nuget.org/packages/Oracle.ManagedDataAccess/18.3.0) is failing.

### DIFF
--- a/src/coverlet.core/Instrumentation/Instrumenter.cs
+++ b/src/coverlet.core/Instrumentation/Instrumenter.cs
@@ -50,7 +50,18 @@ namespace Coverlet.Core.Instrumentation
             _logger = logger;
         }
 
-        public bool CanInstrument() => InstrumentationHelper.HasPdb(_module);
+        public bool CanInstrument()
+        {
+            try
+            {
+                return InstrumentationHelper.HasPdb(_module);
+            }
+            catch (Exception ex)
+            {
+                _logger.LogWarning($"Unable to instrument module: '{_module}' because : {ex.Message}");
+                return false;
+            }
+        }
 
         public InstrumenterResult Instrument()
         {


src/coverlet.core/Helpers/InstrumentationHelper.cs
```diff
@@ -57,24 +57,33 @@ public static string[] GetCoverableModules(string module, string[] directories)
...
                         }
-
-                        return File.Exists(codeViewData.Path);
                     }
-                }
 
+                    return false;
+                }
+            }
+            catch (Exception ex)
```
MarcoRossignoli: I think it's better try/catch where we use HasPdb https://github.com/tonerdo/coverlet/blob/13c069cde013d6cf63c90a91a9f19588b0d37ea7/src/coverlet.core/Instrumentation/Instrumenter.cs#L53
and use logger 
https://github.com/tonerdo/coverlet/blob/13c069cde013d6cf63c90a91a9f19588b0d37ea7/src/coverlet.core/Instrumentation/Instrumenter.cs#L29

https://github.com/tonerdo/coverlet/blob/f25833315cd2f0a124330754dc85f01a16ce6efa/src/coverlet.msbuild.tasks/InstrumentationTask.cs#L102

ViceIce: Ok

src/coverlet.core/Coverage.cs
```diff
@@ -80,23 +80,30 @@ public void PrepareModules()
...
-
-                    // Guard code path and restore if instrumentation fails.
-                    try
-                    {
-                        var result = instrumenter.Instrument();
-                        _results.Add(result);
-                        _logger.LogInformation($"Instrumented module: '{module}'");
-                    }
-                    catch (Exception ex)
+                    if (instrumenter.CanInstrument())
```
MarcoRossignoli: Why not inside `CanInstrument()` prop?
We can log there and return false here, code seems more clean to me

ViceIce: Ok, will change it tomorrow.

MarcoRossignoli: Yup don't worry no rush!thank's for help!

ViceIce: dome

src/coverlet.core/Instrumentation/Instrumenter.cs
```diff
@@ -50,7 +50,18 @@ public Instrumenter(string module, string identifier, string[] excludeFilters, s
...
-        public bool CanInstrument() => InstrumentationHelper.HasPdb(_module);
+        public bool CanInstrument()
+        {
+            try
+            {
+                return InstrumentationHelper.HasPdb(_module);
+            }
+            catch (Exception ex)
+            {
+                _logger.LogWarning($"Unable to instrument module: {_module} because : {ex.Message}");
```
MarcoRossignoli:
```suggestion
                _logger.LogWarning($"Unable to instrument module: '{_module}' because : {ex.Message}");
```

ViceIce: done

PR #368, opened 2019-03-12
fix: package layout

### LINKED ISSUE #358: coverlet.msbuild does not work for projects targeting .NET Framework 4.6
This regression was introduced in #355.

> ❗️ This change broke coverlet support for projects targeting .NET Framework prior to 4.6.1. Placing the props/targets in a framework-specific folder makes them dependent on the declared target framework of the project, which we do not want. We only want them to be dependent on the target framework of the currently executing version of MSBuild.

_Originally posted by @sharwell in https://github.com/tonerdo/coverlet/pull/355#discussion_r262329107_

ViceIce:
closes #358

### DIFF
--- a/src/coverlet.msbuild.tasks/coverlet.msbuild.tasks.csproj
+++ b/src/coverlet.msbuild.tasks/coverlet.msbuild.tasks.csproj
@@ -21,7 +21,7 @@
     <RepositoryUrl>https://github.com/tonerdo/coverlet</RepositoryUrl>
     <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
     <TargetsForTfmSpecificContentInPackage>$(TargetsForTfmSpecificContentInPackage);PackBuildOutputs</TargetsForTfmSpecificContentInPackage>
-    <ContentTargetFolders>build\$(TargetFramework)\</ContentTargetFolders>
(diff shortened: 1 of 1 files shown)


src/coverlet.msbuild.tasks/coverlet.msbuild.tasks.csproj
```diff
@@ -21,7 +21,7 @@
     <RepositoryUrl>https://github.com/tonerdo/coverlet</RepositoryUrl>
     <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
     <TargetsForTfmSpecificContentInPackage>$(TargetsForTfmSpecificContentInPackage);PackBuildOutputs</TargetsForTfmSpecificContentInPackage>
-    <ContentTargetFolders>build\$(TargetFramework)\</ContentTargetFolders>
+    <ContentTargetFolders>build\</ContentTargetFolders>
```
sharwell: 💡 This does not need a trailing `\`.

🔗 https://github.com/NuGet/NuGet.Client/blob/23b76fef368db1df85d4ff5418bb66c25c499c29/src/NuGet.Core/NuGet.Build.Tasks.Pack/NuGet.Build.Tasks.Pack.targets#L38

```suggestion
    <ContentTargetFolders>build</ContentTargetFolders>
```

Issue #343, opened 2019-02-18
Incorrect coverage for statement lambda function


driseley:
Coverlet appears to report incorrect coverage for anonymous statement lambda functions.

For the following class:

```cs
using System;
using System.IO;
using System.Text;

namespace DemoLibrary
{
	public class DemoClass
	{
		protected T WriteToStream<T>(Func<Stream, bool, T> getResultFunction)
		{
			using (var stream = new MemoryStream())
			{
				var result = getResultFunction(stream, false);
				stream.Close();
				Console.WriteLine(Encoding.Default.GetString(stream.ToArray()));
				return result;
			}
		}

		public bool WriteHelloWorldToStream(Stream stream , bool condition)
		{
			if (condition) stream.Write(Encoding.Default.GetBytes("Hello World"));
			else stream.Write(Encoding.Default.GetBytes("Goodbye World"));        
			return condition;
		}

		public bool InvokeFunction()
		{
			return WriteToStream((stream,condition) => WriteHelloWorldToStream(stream, condition));
		}

		public bool InvokeAnonymous()
		{
			return WriteToStream((stream, condition) =>
			{
				if (condition) stream.Write(Encoding.Default.GetBytes("Hello World"));
				else stream.Write(Encoding.Default.GetBytes("Goodbye World"));               
				return condition;
			}
			);
		}
	}
}
```
The following tests:

```cs
using DemoLibrary;
using System;
using System.IO;
using Xunit;

namespace DemoLibraryTests
{
	public class DemoClassTests
	{
		[Fact]
		public void InvokeFunction_Test()
		{
			DemoClass demoClass = new DemoClass();
			bool result = demoClass.InvokeFunction();
			Assert.False(result);
		}

		[Fact]
		public void InvokeAnonymous_Test()
		{
			DemoClass demoClass = new DemoClass();
			bool result = demoClass.InvokeAnonymous();
			Assert.False(result);
		}
	}
}
```

Report the following results:
![image](https://user-images.githubusercontent.com/6659301/52945061-cb2f4980-3368-11e9-9af9-69ff3e5b39a3.png)

Notice that the coverage for the InvokeAnonymous function shows all lines and all branches covered ( which clearly cannot be correct ).  For the named function, the coverage is correct.

driseley:
I have created and example project that demonstrates the issue:
https://github.com/driseley/coverlet-issue-343

driseley:
Coverlet version 2.5.1, Project .Net Core 2.1

[...]

PR #369, opened 2019-03-12
[Bug fix]Incorrect coverage for statement lambda function

MarcoRossignoli:
fixes https://github.com/tonerdo/coverlet/issues/343

I try to explain what happened.
When we use anonymous delegate compiler generate custom class and passes it as delegate.
If inside this method there are branches this hits are counted for "anonymous" compiler generated class.
This lead to missing "branches" from "actual" class(the method of the class that contains delegate).
Fortunately line of branches inside anonymous compiler generated  class are correct, so my stategy is move all "anonymous" branches to "actual" class and after remove generated class from report to avoid noise(however it doesn't show because generated class has got 0 lines but only branches list).
In case of @driseley repro seems work well.

image

It's not possible write a test, or rather we could in future for this type of test generate a project and run real coverage and check "reports"(similar to BDN), but it's not easy/fast work.

@driseley I ask to you if kindly can clone build my fork/branch and test with your actual project. If you need help to setup let me know we could chat.

@sharwell I IIRW you work for Roslyn, please can you trust my "strategy" and help with review?

/cc @tonerdo @petli


driseley:
Hi @MarcoRossignoli 

Thanks for looking at this - I think have been able to build and test your branch ok, as I was able to reproduce the fixed coverage on my demo project.  However when I applied the fix  to my real project, it did not seem to address the issue.

Here is a screenshot of the coverage of the named function:

image

And here is the lamda function:

image

As you can see the results are still quite different for the same set of tests.

Unfortunately I can't share the whole code , but let me know if there is some more information I can give you to help fully reproduce this.

Kind Regards
Dave

sharwell:
Interesting. I would expect coverage of the named function to show _partially_ covered for each of the branches, but for some reason they seem to be showing as fully covered.

sharwell:
@MarcoRossignoli I don't understand this case well enough right now to figure out the difference, but I may be able to help with setting up a test harness.

MarcoRossignoli:
@sharwell I thought also to add a verbose file output(maybe with switch i.e. `/p:CoverletDebug=true`) and generate a file with detailed operation/class/line/branch/hit maybe this could be used by @driseley without too much disclosure, only namespace/class/method name if possible, could be possible @driseley?

sharwell:
Have you tried to reproduce this with an `async` lambda? I find those sometimes have different results in code coverage tools.

MarcoRossignoli:
> Have you tried to reproduce this with an async lambda? I find those sometimes have different results in code coverage tools.

Actually not, I used repro, maybe the issue is there.

sharwell:
Yep, I noticed reply above confirmed that you fixed the original scenario, but the new screenshot shows async code.

MarcoRossignoli:
Maybe related to different generated code I searched for specific compiler attribute(CompilerGeneratedAttribute), maybe we're missing async state machine.

driseley:
I'll try to abstract the code that I have in the screenshots above to my reproducer repo, as that might help things.  Should take a couple of hours if can untangle it.

MarcoRossignoli:
> I'll try to abstract the code that I have in the screenshots above to my reproducer repo, as that might help things. Should take a couple of hours if can untangle it.

Great, the code should be very close but with async lambda.

driseley:
I have added a new class and test case here:

https://github.com/driseley/coverlet-issue-343/commit/afb84e01d54f1664522efb0ee7fb4e7448336484

that reproduces the issue in the screenshots above

MarcoRossignoli:
I did quick investigation and the issue seems unrelated to this PR, maybe related to nested async state machine where we remove one of two branch(because expected not covered), but removed branch is not the correct one(not related to sync/async execution).
I'll come back...

MarcoRossignoli:
I tried another way to solve this. 
Issue is not related to new features(report lambda methods branches) but to the way we remove async branch in async state machine MoveNext() method.
New strategy is to find async branch pattern(search `call instance bool [System.Runtime]System.Runtime.CompilerServices.TaskAwaiter``1<!T>::get_IsCompleted()`) and avoid to instrument it, because inside this branch there is no interesting code, it only enqueue async callback.
In case of sync code this code will never be covered, old strategy remove non covered branches in MoveNext() method, but it's not correct because inside this method we could have "application logic"(for nested async/await code) and we cannot simply "remove non covered branches".

@driseley can you try this new version in your prod code?

/cc @tonerdo

driseley:
@MarcoRossignoli  - thanks for the updates, however they don't appear to have made any difference to the test case (or my prod code).  Could you confirm that the test case is now reporting correctly for you?

MarcoRossignoli:
@driseley you should see a missed branch on if, btw you should see again covered line inside if block.

MarcoRossignoli:
This is not a complete solution of your problem...I don't know if it's solvable at the moment 😞 

TL;DR;

The issue is that compiler generate 2 state machine and one state machine create and run other.
Coverlet combines "hit"(coveret point) with "c# source range lines" to undestand what is covered, for instance hit number x cover lines from y-z and does this using [DebugInformation](https://github.com/tonerdo/coverlet/blob/master/src/coverlet.core/Instrumentation/Instrumenter.cs#L323)
In your case when we call `method.DebugInformation.GetSequencePoint(instruction)` for outermost async machine state the range returned is from start method to end method.
This lead to "false positive" because we "cover" and "hit" that point on runtime but during "the count" we count +1 hit for every "line of code associated" you can see alg [here](https://github.com/tonerdo/coverlet/blob/master/src/coverlet.core/Coverage.cs#L249) but actually it's false, the issue here is that we cannot change the behaviour of DebugInformation API.

My current update is to "change" the strategy to avoid false "negative" on async branch of machine state, and works well , old PR https://github.com/tonerdo/coverlet/pull/158 to almost see correct "branch" results(uncovered branch on ifs in your case, better than nothing you can see on reports)

But I have no idea at the moment how we can fix this use case, I'm trying to understand if it's possible find out "end line of branch" for now we've only the start, in that way if we found uncovered branch we can "override" false "positive"(because branch win on others).

I ask to @tonerdo @sharwell if they've some idea or other point of view of the problem.

SteveGilham:
The algorithm I use is derived from OpenCover's process for finding branch `offsetend` and `offsetchain` values, and is implemented in the [AltCover `Visitor.getJumps` function](https://github.com/SteveGilham/altcover/blob/master/AltCover/Visitor.fs#L601)

This is what OpenCover essentially does --

For each conditional branch instruction get final destination:
  get end of containing sequence point: step until the next instruction is either null (end of method) or starts another sequence point
  for each possible next location from the conditional branch
  label:
   if the target is an unconditional branch, add to the chain for this target and recurse to label
   if the target is after the end of the original sequence point, or is a conditional branch, terminate recursion
   step to next instruction and recurse to label

I embellish by adding --

For non-switch conditional branches, if the range of instructions from the next instruction (branch not taken) to the branch target inclusive contains neither a sequence point boundary nor a return instruction (as in a ternary return expression), then I drop the branch from the report as being something compiler generated within the sequence point.

MarcoRossignoli:
@driseley can you re-try with new updates?On my local seems fixed, take a look also if all other coverage is ok

image

Fix: the issue as explained above is that for nested async/away/lambda there is a chance that one state machine create another and pdb report "large block" as covered, but it's a false positive i.e.
Source.cs--> StateMachineA.MoveNext()---Lines 10-50
Source.cs--> StateMachineB.MoveNext()---Lines 20-23 <-- non covered
StateMachineA creates StateMachineB
If we instrument StateMachineA and we cover that point(before new StateMachineB) we count 10-50 as covered lines, but actually the "real" user code is in StateMachineB.MoveNext().
Update "corrects" counting resetting hits for lines that are "non covered" by every covered line registered.
The idea should work because we expect that if we have more than one "copy" of line in covered lines if covered should be covered in "all" copies, in sample above:

```
nonCoveredLines = CountHits(lines, hitfile);
for nonCoveredline in nonCoveredLines 
for line in lines
if line == nonCoveredLine then line.hits = 0
```

driseley:
@MarcoRossignoli  - Thanks for keeping working on this.  My local testing matches your results, so it seems to largely address the coverage issue.  There is a remaining issue - the coverage counts still seem to be one higher than expected:
image
In my repo - there is only one test case that covers the code - but some lines are shown as being covered twice.  I can live with it - but is that something you want to continue to look at?

MarcoRossignoli:
> but is that something you want to continue to look at?

I think so, I understood the problem now...I'll give it a try...BTW other coverage is ok(other branch lines unrelated to this)?

driseley:
Thanks , the only other problem I can spot is in the sync lambda coverage - which again has the double counting:
image

otherwise it all looks good to me.

SteveGilham:
@driseley have you checked the raw coverage file for visit counts?  ReportGenerator's assessment of whether a line is covered can be optimistic when sequence points span many lines -- it gives you comfy reports with rounded up numbers that will make management feel warm fuzzies.

driseley:
@SteveGilham  - for the screenshot above:
```json
        "System.Boolean DemoLibrary.DemoClass::InvokeAnonymous()": {
          "Lines": {
            "36": 1,
            "37": 1,
            "38": 2,
            "39": 2,
            "40": 0,
            "41": 0,
            "42": 0,
            "43": 2,
            "44": 2,
            "45": 2,
            "46": 1,
            "47": 1
          },
```

MarcoRossignoli:
@driseley @SteveGilham the issue is clear to me...I cleaned up false positive on line but "other lines" are accounted again, the greatest problem here is understand how to exclude instrumentation of async lambda it's similar to async state machine(it has got MoveNext()) but inside there isn't user code, but at the moment I don't have idea how to exclude without remove user code.
To better explain the issue is "async lambda that call async code".

driseley:
Hi Marco - just to be clear, in the example above - that was for sync code, not async.  So could the issue be with lambda calls - not sync/async?

MarcoRossignoli:
> So could the issue be with lambda calls...

I think so the issue should be the same...I'll take a look again, should be more simple to fix if the pattern of issue is the same.

PR #383, opened 2019-04-25
Restore backup assemblies when sigkill

ViktorHofer:
With sigkill or other execution interruptions the instrumented assemblies will remain in the output directory and the file backups won't be restored. Adding a cache to restore the backups in such cases.

cc @MarcoRossignoli @tonerdo

### DIFF
--- a/src/coverlet.core/Helpers/InstrumentationHelper.cs
+++ b/src/coverlet.core/Helpers/InstrumentationHelper.cs
@@ -14,6 +14,13 @@ namespace Coverlet.Core.Helpers
 {
     internal static class InstrumentationHelper
     {
+        private static readonly Dictionary<string, string> _backupList = new Dictionary<string, string>();
+
+        static InstrumentationHelper()
+        {
+            AppDomain.CurrentDomain.ProcessExit += (s,e) => RestoreOriginalModules();
+        }
+
         public static string[] GetCoverableModules(string module, string[] directories, bool includeTestAssembly)
         {
             Debug.Assert(directories != null);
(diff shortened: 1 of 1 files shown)


MarcoRossignoli:
@ViktorHofer let me understand better the case...if the "test process" is "interrupted" for some reason why recover?I mean if it's stopped will be eventually replaced by new build/test, no?

ViktorHofer:
If you want to "safely" instrument assemblies - means the output should be exactly the same as before - a sigkill could cause big troubles here. We are hitting this in corefx whenever a coverage measurement takes too long and the user kills the process. We don't rebuild all the assemblies that we use during the measurement, i.e. CoreLib.

But this isn't tied to corefx at all. Since the `--include-directory` option was added, additional assembly directories can be specified. Usually those additional inputs aren't rebuilt and should not get into a dirty state.

src/coverlet.core/Helpers/InstrumentationHelper.cs
```diff
@@ -14,6 +14,13 @@ namespace Coverlet.Core.Helpers
 {
     internal static class InstrumentationHelper
     {
+        private static readonly Dictionary<string, string> _backupList = new Dictionary<string, string>();
```
MarcoRossignoli: sooner or later we should remove al these statics and use msbuild input/output https://github.com/tonerdo/coverlet/issues/364

MarcoRossignoli:
@tonerdo CI fail with some test on cecil tests...is it never happened before?No issue on my local.

tonerdo:
@MarcoRossignoli taking a look now

ViktorHofer:
Is there ready to be merged?

MarcoRossignoli:
@ViktorHofer in future I'll remove that static https://github.com/tonerdo/coverlet/issues/387 for now if it's a blocking issue LGTM cc: @tonerdo

ViktorHofer:
Yeah let's merge this first and ideally publish a new package first and afterwards do the refactoring.

Issue #381, opened 2019-04-25
Adjust log level when using coverlet.console (global tool)


ViktorHofer:
How can I adjust the log level when using coverlet.console - the global tool - and not coverlet.msbuild?

cc @MarcoRossignoli @tonerdo

MarcoRossignoli:
AFAIK at the moment is not possible avoid [`PrepareModules`](https://github.com/tonerdo/coverlet/blob/master/src/coverlet.console/Program.cs#L64) logs, coverlet.console use a simple `ConsoleLogger` implementation.
We should add control switch similar to `TaskLoggingHelper` https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-command-line-reference?view=vs-2019
```
-v|--verbosity q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic]
```
At the moment we use information/warning/error so we could print  always error/warning and info for level > `q[uiet]` (like now `minimal` level log inclusion/exclusion list)
What do you think @tonerdo?

PR #384, opened 2019-04-25
Add log level option

ViktorHofer:
Fixes https://github.com/tonerdo/coverlet/issues/381

@MarcoRossignoli @tonerdo

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ Options:
   -t|--target             Path to the test runner application.
   -a|--targetargs         Arguments to be passed to the test runner.
   -o|--output             Output of the generated coverage report
-  -v|--verbosity          Sets the verbosity level of the command. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed].
+  -v|--verbosity          Sets the verbosity level of the command. Allowed values are quiet, minimal, normal, detailed.
   -f|--format             Format of the generated coverage report.
   --threshold             Exits with error if the coverage % is below value.
   --threshold-type        Coverage type to apply the threshold to.
--- a/README.md
+++ b/README.md
@@ -61,6 +61,7 @@ Options:
   -t|--target             Path to the test runner application.
   -a|--targetargs         Arguments to be passed to the test runner.
   -o|--output             Output of the generated coverage report
+  -v|--verbosity          Sets the verbosity level of the command. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed].
   -f|--format             Format of the generated coverage report.
   --threshold             Exits with error if the coverage % is below value.
   --threshold-type        Coverage type to apply the threshold to.
--- a/src/coverlet.console/Logging/ConsoleLogger.cs
+++ b/src/coverlet.console/Logging/ConsoleLogger.cs
@@ -7,19 +7,22 @@ namespace Coverlet.Console.Logging
     class ConsoleLogger : ILogger
     {
         private static readonly object _sync = new object();
+        public LogLevel Level { get; set; } = LogLevel.Normal;
 
-        public void LogError(string message) => Log(message, ConsoleColor.Red);
+        public void LogError(string message) => Log(LogLevel.Minimal, message, ConsoleColor.Red);
 
         public void LogError(Exception exception) => LogError(exception.ToString());
 
-        public void LogInformation(string message) => Log(message, ForegroundColor);
+        public void LogInformation(string message) => Log(LogLevel.Detailed, message, ForegroundColor);
 
         public void LogVerbose(string message) => throw new NotImplementedException();
 
-        public void LogWarning(string message) => Log(message, ConsoleColor.Yellow);
+        public void LogWarning(string message) => Log(LogLevel.Normal, message, ConsoleColor.Yellow);
 
-        private void Log(string message, ConsoleColor color)
+        private void Log(LogLevel level, string message, ConsoleColor color)
         {
+            if (level < Level) return;
+
             lock (_sync)
             {
                 ConsoleColor currentForegroundColor;
--- a/src/coverlet.console/Logging/ConsoleLogger.cs
+++ b/src/coverlet.console/Logging/ConsoleLogger.cs
@@ -13,7 +13,7 @@ namespace Coverlet.Console.Logging
 
         public void LogError(Exception exception) => LogError(exception.ToString());
 
-        public void LogInformation(string message) => Log(LogLevel.Normal, message, ForegroundColor);
+        public void LogInformation(string message, bool important = false) => Log(important ? LogLevel.Minimal : LogLevel.Normal, message, ForegroundColor);
 
         public void LogVerbose(string message) => Log(LogLevel.Detailed, message, ForegroundColor);
 
--- a/src/coverlet.console/Logging/ConsoleLogger.cs
+++ b/src/coverlet.console/Logging/ConsoleLogger.cs
@@ -9,15 +9,15 @@ namespace Coverlet.Console.Logging
         private static readonly object _sync = new object();
         public LogLevel Level { get; set; } = LogLevel.Normal;
 
-        public void LogError(string message) => Log(LogLevel.Minimal, message, ConsoleColor.Red);
+        public void LogError(string message) => Log(LogLevel.Quiet, message, ConsoleColor.Red);
 
         public void LogError(Exception exception) => LogError(exception.ToString());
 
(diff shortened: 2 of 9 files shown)


src/coverlet.console/coverlet.console.csproj
```diff
@@ -21,7 +21,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.CommandLineUtils" Version="1.1.1" />
+    <PackageReference Include="McMaster.Extensions.CommandLineUtils" Version="2.3.4" />
```
ViktorHofer: `Microsoft.Extensions.CommandLineUtils` is a stale dependency. The community fork supports enum parsing which was the reason why I upgraded the dependency.

src/coverlet.console/Program.cs
```diff
@@ -29,6 +28,7 @@ static int Main(string[] args)
             CommandOption target = app.Option("-t|--target", "Path to the test runner application.", CommandOptionType.SingleValue);
             CommandOption targs = app.Option("-a|--targetargs", "Arguments to be passed to the test runner.", CommandOptionType.SingleValue);
             CommandOption output = app.Option("-o|--output", "Output of the generated coverage report", CommandOptionType.SingleValue);
+            CommandOption<LogLevel> verbosity = app.Option<LogLevel>("-v|--verbosity", "Sets the verbosity level of the command. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed].", CommandOptionType.SingleValue);
```
MarcoRossignoli: Never used this, if we specify `q` map first enum starting with `q` so in this case `quiet` or we need to specify full name?

README.md
```diff
@@ -61,6 +61,7 @@ Options:
   -t|--target             Path to the test runner application.
   -a|--targetargs         Arguments to be passed to the test runner.
   -o|--output             Output of the generated coverage report
+  -v|--verbosity          Sets the verbosity level of the command. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed].
```
MarcoRossignoli: would you mind remove duplicate `--include-directory ...` switch explanation duplicate below?

ViktorHofer: done

src/coverlet.console/Program.cs
ViktorHofer: I don't know yet. Need to test my change.

MarcoRossignoli: ok let us know

src/coverlet.console/Logging/LogLevel.cs
```diff
@@ -0,0 +1,30 @@
...
+        /// <summary>
+        /// Logs that highlight when the current flow of execution is stopped due to a failure. These should indicate a
+        /// failure in the current activity, not an application-wide failure.
+        /// </summary>
+        Minimal = 2,
+
+        /// <summary>
+        /// Not used for writing log messages. Specifies that a logging category should not write any messages.
+        /// </summary>
+        Quiet = 3
```
MarcoRossignoli: `Quiet` should not hide warning and errors...I did some test with `dotnet test "D:\git\coverlet\test\coverlet.core.tests\coverlet.core.tests.csproj" -c Debug /p:CollectCoverage=true /p:CoverletOutputFormat=opencover /p:Include=[coverlet.*]* -verbosity:quiet` and added some LogInfo/LogError and they show on output.

MarcoRossignoli:
@ViktorHofer feel free to add https://github.com/tonerdo/coverlet/commit/51ea7930008cb581d9bf0d1b356eb8434ccc7c0c#commitcomment-33309599 here...to avoid conflicts.

src/coverlet.console/Logging/LogLevel.cs
ViktorHofer: done.

src/coverlet.console/Program.cs
ViktorHofer: Ok the short forms didn't work. Removed them.

ViktorHofer:
@MarcoRossignoli this is now ready

README.md
MarcoRossignoli: Please update with "long form".

ViktorHofer: Done.

src/coverlet.console/Logging/ConsoleLogger.cs
```diff
@@ -7,19 +7,22 @@ namespace Coverlet.Console.Logging
...
         private static readonly object _sync = new object();
+        public LogLevel Level { get; set; } = LogLevel.Normal;
 
-        public void LogError(string message) => Log(message, ConsoleColor.Red);
+        public void LogError(string message) => Log(LogLevel.Quiet, message, ConsoleColor.Red);
 
         public void LogError(Exception exception) => LogError(exception.ToString());
 
-        public void LogInformation(string message) => Log(message, ForegroundColor);
+        public void LogInformation(string message, bool important = false) => Log(important ? LogLevel.Minimal : LogLevel.Normal, message, ForegroundColor);
```
MarcoRossignoli: @ViktorHofer I don't understand new `important` flag. If default level is `Normal` and flag can pass `Minimal` or default `Normal` and we check `if (level < Level) return;`  default level won't skip informations(I think it's what you want).
Or your idea is to have a way to log information also in case of `Quiet`?

src/coverlet.console/Logging/LogLevel.cs
```diff
@@ -0,0 +1,30 @@
...
+        Normal = 1,
+
+        /// <summary>
+        /// Logs that highlight when the current flow of execution is stopped due to a failure. These should indicate a
+        /// failure in the current activity, not an application-wide failure.
+        /// </summary>
+        Minimal = 2,
+
+        /// <summary>
+        /// Not used for writing log messages. Specifies that a logging category should not write any messages.
```
MarcoRossignoli: add `except warning and errors`

README.md
```diff
@@ -61,6 +61,7 @@ Options:
   -t|--target             Path to the test runner application.
   -a|--targetargs         Arguments to be passed to the test runner.
   -o|--output             Output of the generated coverage report
+  -v|--verbosity          Sets the verbosity level of the command. Allowed values are quiet, minimal, normal, detailed.
```
MarcoRossignoli: I did some test with coverlet tool and msbuild and this PR...and seem that is not possible keep same behaviour between msbuild logger and console logger levels.
I mean we want behave in a different way if we use command or msbuild(msbuild rule are not under our control), and to do this we need to add some "flag" on interface(i.e. `important` that is not honored on msbuild, not so clean).
Again maybe if we support verbosity we should concat it as a `--targetargs`?

I wonder if could be more "clear" avoid levels at all(my first idea with better hope of success :smile:), and simply add `-quiet` switch, and nop the console.
The idea to see all coverlet log(on coloured console, red yellow white) is not so bad to me, we explicity run `coverlet` tool command so it's expected some logs from it.
But it's also true that someone would want not see anything and only verify "reports" at the end.

After Viktor's attempt(thank's a lot!) I don't like the idea to mix two different behaviour under same "verbosity concepts" that cannot be honorated except with some "non clean" solution. 
What do you think Viktor?

I would like to know @tonerdo's thought

ViktorHofer: Why are different log outputs between msbuild and coverlet.console an issue for you? The former is running as part of an assembly's build, the latter is a standalone tool. I would not expect to have identical outputs.

MarcoRossignoli: > Why are different log outputs between msbuild and coverlet.console an issue for you?

I would expect that output using msbuild or using coverlet tool should be the same for same level, for instance if I would like to pass `minimal` to tool and msbuild and get same output(at the moment we show the list of instrumented assemblies), and possibly avoid to change contract(log interface) only to accomodate one scenario.

MarcoRossignoli: @tonerdo is [not a "fan"](https://github.com/tonerdo/coverlet/issues/388#issuecomment-488676943) of my sync idea...so I lose 😄 
I think we can go on with "unsync" one.

src/coverlet.console/Logging/LogLevel.cs
MarcoRossignoli:
```suggestion
        /// Not used for writing log messages. Specifies that a logging category should not write any messages except warning and errors.
```

ViktorHofer:
So can this be merged now?

MarcoRossignoli:
cc: @tonerdo

Issue #389, opened 2019-05-01
Switch symbol tests fail in "release mode"


MarcoRossignoli:
It's been a while that cecil symbol tests on switch fails in release mode.
I did some investigation and seem that the issue is the way that compiler compile pdb for debug/release.
Debug:
```
// Token: 0x0600001B RID: 27 RVA: 0x00002DDC File Offset: 0x00000FDC
.method public hidebysig 
	instance bool HasSwitch (
		int32 input
	) cil managed 
{
	// Header Size: 12 bytes
	// Code Size: 43 (0x2B) bytes
	// LocalVarSig Token: 0x11000009 RID: 9
	.maxstack 1
	.locals init (
		[0] int32,
		[1] int32,
		[2] bool
	)

	/* (45,9)-(45,10) D:\git\coverlet\test\coverlet.core.tests\Samples\Samples.cs */
	/* 0x00000FE8 00           */ IL_0000: nop
	/* (46,13)-(46,27) D:\git\coverlet\test\coverlet.core.tests\Samples\Samples.cs */
	/* 0x00000FE9 03           */ IL_0001: ldarg.1
	/* 0x00000FEA 0B           */ IL_0002: stloc.1
	/* (hidden)-(hidden) D:\git\coverlet\test\coverlet.core.tests\Samples\Samples.cs */
	/* 0x00000FEB 07           */ IL_0003: ldloc.1
	/* 0x00000FEC 0A           */ IL_0004: stloc.0
	/* (hidden)-(hidden) D:\git\coverlet\test\coverlet.core.tests\Samples\Samples.cs */
	/* 0x00000FED 06           */ IL_0005: ldloc.0
	/* 0x00000FEE 450300000002000000060000000A000000 */ IL_0006: switch    (IL_0019, IL_001D, IL_0021)

	/* 0x00000FFF 2B0C         */ IL_0017: br.s      IL_0025

	/* (49,21)-(49,33) D:\git\coverlet\test\coverlet.core.tests\Samples\Samples.cs */
	/* 0x00001001 17           */ IL_0019: ldc.i4.1
	/* 0x00001002 0C           */ IL_001A: stloc.2
	/* 0x00001003 2B0C         */ IL_001B: br.s      IL_0029

	/* (51,21)-(51,34) D:\git\coverlet\test\coverlet.core.tests\Samples\Samples.cs */
	/* 0x00001005 16           */ IL_001D: ldc.i4.0
[...]

PR #390, opened 2019-05-01
Skip switch symbol tests fail in "release mode"

MarcoRossignoli:
closes https://github.com/tonerdo/coverlet/issues/389
Temporary skip tests in release mode.

/cc @tonerdo @petli

### DIFF
--- a/test/coverlet.core.tests/Symbols/CecilSymbolHelperTests.cs
+++ b/test/coverlet.core.tests/Symbols/CecilSymbolHelperTests.cs
@@ -108,7 +108,7 @@ namespace Coverlet.Core.Symbols.Tests
             Assert.Equal(34, points[1].StartLine);
         }
 
-#if !RELEASE
+#if !RELEASE // Issue https://github.com/tonerdo/coverlet/issues/389
         [Fact]
         public void GetBranchPoints_Switch()
         {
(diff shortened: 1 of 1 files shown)


MarcoRossignoli:
@tonerdo I tried to fix checking right line in release but it doesn't work...the issue raise only on Windows/Release

MarcoRossignoli:
I did some test with Ubuntu 18.04 and I get same exception from release config...maybe we should use a different Ubuntu image on appveyor to get consistent behaviour `Ubuntu1804` https://www.appveyor.com/updates/
VS 2019 will be available this month https://github.com/appveyor/ci/issues/2907#issuecomment-488345102

MarcoRossignoli:
Another idea could be avoid to rely on "machine sdks" and download a run build/test with "specific" sdk in a cross way https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-install-script so we would have same "compiler"

tonerdo:
What version of Ubuntu does AppVeyor use by default?

MarcoRossignoli:
We use `Ubuntu` and from AppVeyor docs 
```
Linux images:
Ubuntu or Ubuntu1604 - Ubuntu 16.04
Ubuntu1804 - Ubuntu 18.04
```

Issue #391, opened 2019-05-02
System.Reflection.Metadata version mismatch unable to find file System.Reflection.Metadata,Version=1.4.3.0


vagisha-nidhi:
I am trying to load coverlet into a [datacollector ](https://github.com/Microsoft/vstest-docs/blob/master/docs/extensions/datacollector.md)which would collect coverage and report back the results to [TestPlatform](https://github.com/Microsoft/vstest). The issue with loading coverlet APIs is that coverlet.core depends on [System.Reflection.Metadata, AssemblyVersion : 1.4.3.0](https://github.com/tonerdo/coverlet/blob/062b907735901bdf3ad9d4d40953f53a4fe3f364/src/coverlet.core/coverlet.core.csproj#L12) and the datacollector process (in which it is being loaded) has a dependency on System.Reflection.Metadata, AssemblyVersion=1.4.2.0

This results in **System.Reflection.Metadata, AssemblyVersion=1.4.2.0** being loaded and coverlet erroring out with `System.IO.FileLoadException: Could not load file or assembly 'System.Reflection.Metadata, Version=1.4.3.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. Could not find or load a specific file.`

This can be resolved with lowering down the version of [System.Reflection.Metadata](https://github.com/tonerdo/coverlet/blob/062b907735901bdf3ad9d4d40953f53a4fe3f364/src/coverlet.core/coverlet.core.csproj#L12).

I am a Software Engineer on the [TestPlatform](https://github.com/Microsoft/vstest) team and can raise a PR to address the issue.

MarcoRossignoli:
> I am trying to load coverlet into a datacollector which would collect coverage and report back the results to TestPlatform. 

Can you elaborate the scenario?We have some work in place to use coverlet with data collector to fix issue where test plat kill process before we can save instrumentation report file to disk https://github.com/tonerdo/coverlet/pull/329

vagisha-nidhi:
@MarcoRossignoli 
We have written an [outproc datacollector](https://github.com/Microsoft/vstest-docs/blob/master/docs/extensions/datacollector.md) that references coverlet.core.dll. This calls into coverlet APIs such as coverage.PrepareModules(), coverage.GetCoverageResult(). We get coverage results thus and send these back to testplatform using [DataCollectionSink](https://github.com/Microsoft/vstest-docs/blob/master/docs/extensions/datacollector.md#datacollectionsink). 

[...]

PR #392, opened 2019-05-02
Updating System.Reflection.Metadata version

vagisha-nidhi:
Issue related #391 
Updating version of System.Reflection.Metadata

### DIFF
--- a/src/coverlet.core/coverlet.core.csproj
+++ b/src/coverlet.core/coverlet.core.csproj
@@ -9,6 +9,7 @@
   <ItemGroup>
     <PackageReference Include="Mono.Cecil" Version="0.10.1" />
     <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
+     <!--Do not change System.Reflection.Metadata version since we need to support VSTest DataCollectors. Goto https://www.nuget.org/packages/System.Reflection.Metadata to check versions.-->
     <PackageReference Include="System.Reflection.Metadata" Version="1.5.0" />
     <PackageReference Include="Microsoft.Extensions.FileSystemGlobbing" Version="2.0.1" />
   </ItemGroup>
--- a/src/coverlet.core/coverlet.core.csproj
+++ b/src/coverlet.core/coverlet.core.csproj
@@ -9,7 +9,11 @@
   <ItemGroup>
     <PackageReference Include="Mono.Cecil" Version="0.10.1" />
     <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
-     <!--Do not change System.Reflection.Metadata version since we need to support VSTest DataCollectors. Goto https://www.nuget.org/packages/System.Reflection.Metadata to check versions.-->
+    <!--
+              Do not change System.Reflection.Metadata version since we need to support VSTest DataCollectors. Goto https://www.nuget.org/packages/System.Reflection.Metadata to check versions.
+              We need to load assembly version 1.4.2.0 to properly work
+              We can check minimum supported package version here https://github.com/Microsoft/vstest/blob/master/src/Microsoft.TestPlatform.ObjectModel/Microsoft.TestPlatform.ObjectModel.csproj#L37
+     -->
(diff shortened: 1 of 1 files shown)


src/coverlet.core/coverlet.core.csproj
```diff
@@ -9,7 +9,7 @@
   <ItemGroup>
     <PackageReference Include="Mono.Cecil" Version="0.10.1" />
     <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
-    <PackageReference Include="System.Reflection.Metadata" Version="1.6.0" />
+    <PackageReference Include="System.Reflection.Metadata" Version="1.5.0" />
```
MarcoRossignoli: @vagisha-nidhi can add a comment above like `// Do not change System.Reflection.Metadata version we need to support VSTest DataCollectors` and maybe link some resource where we can check min version if exists

MarcoRossignoli: @vagisha-nidhi I mean if there is a link where we can check minimal version of System.Reflection.Metadata to support DataCollectors

vagisha-nidhi: There isn't any such link as per my knowledge.

MarcoRossignoli: ok no problem

vagisha-nidhi: Here is the link which shows object model dependency on System.Reflection.Metadata [https://github.com/Microsoft/vstest/blob/master/src/Microsoft.TestPlatform.ObjectModel/Microsoft.TestPlatform.ObjectModel.csproj#L37](https://github.com/Microsoft/vstest/blob/master/src/Microsoft.TestPlatform.ObjectModel/Microsoft.TestPlatform.ObjectModel.csproj#L37)
Will this work?

MarcoRossignoli: Could be, but reports 1.3.0 and you amend to 1.5.0 can you explain why?

vagisha-nidhi: The assembly version has only been changes in 1.6.0 that is, 1.4.3.0
Both 1.3.0 and 1.5.0 have the same assembly version(1.4.2.0) and works as expected. So thought of bumping just 1 version down.

MarcoRossignoli: Silly question you're right(verified)

vagisha-nidhi: I'll edit the link to point to the object model csproj for reference? Then this would be good to go, right?

src/coverlet.core/coverlet.core.csproj
```diff
@@ -9,7 +9,8 @@
   <ItemGroup>
     <PackageReference Include="Mono.Cecil" Version="0.10.1" />
     <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
-    <PackageReference Include="System.Reflection.Metadata" Version="1.6.0" />
+     <!--Do not change System.Reflection.Metadata version since we need to support VSTest DataCollectors. Goto https://www.nuget.org/packages/System.Reflection.Metadata to check versions.-->
```
MarcoRossignoli:
```suggestion
     <!--
              Do not change System.Reflection.Metadata version since we need to support VSTest DataCollectors. Goto https://www.nuget.org/packages/System.Reflection.Metadata to check versions.
              We need to load assembly version 1.4.2.0 to properly work
              We can check minimum supported package version here https://github.com/Microsoft/vstest/blob/master/src/Microsoft.TestPlatform.ObjectModel/Microsoft.TestPlatform.ObjectModel.csproj#L37
     -->
```

src/coverlet.core/coverlet.core.csproj
MarcoRossignoli: yes, apply my comments please and to me is ok, we need to wait @tonerdo for merge.

src/coverlet.core/coverlet.core.csproj
vagisha-nidhi: Done

Issue #387, opened 2019-04-30
Improve instrumentation testability


MarcoRossignoli:
Check/debug code coverage is not so simple; languages and compilers abstracts IL.

Working on some issues https://github.com/tonerdo/coverlet/pull/369 https://github.com/tonerdo/coverlet/pull/344 https://github.com/tonerdo/coverlet/pull/158 I found that could be very useful to have a way to "test" entire process to verify the lines we expect to cover, this is the core goal of coverlet.

My proposal is to discuss and do some architectural changes to allow "simple" F5 debug all process and also allow to add "complex" test to discover bugs related to compilers changes that sometimes invalidates assumptions, for instance async/await state machines.

Plan:
- [x] https://github.com/tonerdo/coverlet/pull/409 Find a way to remove all `static` from code, we have already an issue open to track it https://github.com/tonerdo/coverlet/issues/364
In this way we can run "parallel" tests
- [x] ~Abstract file system, at the moment we cannot unload assemblies and we cannot "restore" instrumented locked module~
After some attempts I found unuseful at the moment abstract file system to improve instrumentation testability, in addition to file locking we've a problem with tracker shared statics. To avoid statics on tracker we need to  re-engineer too much code for "testing" pourpose.
Next idea is to write an out of process "runner".
- [x] ~Create an out of process helper to run full instrumentation test (https://github.com/tonerdo/coverlet/pull/420))~
- [x] Add specific test for current and future issues on coverage

Postponed(did some initial work on https://github.com/tonerdo/coverlet/pull/393 but too drastic for now https://github.com/tonerdo/coverlet/pull/393#issuecomment-491637617)
- [x] Start to abstract "main features" for instance `Coverlet.Core.Coverage` 
- [x] ~Add and use a container for DI(https://github.com/tonerdo/coverlet/pull/420)~

Changes above allows us to also measure and improve the performance(complete benchmarks) and verify if the instrumentation is ok(i.e. inspect IL) and to expose in future [`Coverlet.Core`](https://github.com/tonerdo/coverlet/issues/212) as package with "stable apis".

This steps should resolve 

https://github.com/tonerdo/coverlet/blob/0956d1e8da6b7265b0b5e4f93af3499c4be99cc8/test/coverlet.core.tests/CoverageTests.cs#L23-L24

/cc @tonerdo @petli

MarcoRossignoli:
If we agree on the plan, I could start to work on it.

tonerdo:
[...]

PR #393, opened 2019-05-02
Remove static Coverage object

MarcoRossignoli:
closes https://github.com/tonerdo/coverlet/issues/364
replace https://github.com/tonerdo/coverlet/pull/366

* Removed static `InstrumentationTask.Coverage` and used `[Output]` msbuild parameter

Instrumenter state is serialized using `IInstrumentStateSerializer` for now implemented with JsonSerializer

* Organized coverlet internals in two main interface `IInstrumenter` and `ICoverageCalculator`
* Organized cs files under project folder `Instrumentation` and `Coverage`
* Cleanup msbuild Task parameter using auto prop

/cc @tonerdo @petli @sharwell


src/coverlet.core/Abstracts/IInstrumenter.cs
```diff
@@ -0,0 +1,126 @@
...
+        public bool Equals(BranchKey other) => other is BranchKey branchKey && branchKey.Line == this.Line && branchKey.Ordinal == this.Ordinal;
+
+        public override int GetHashCode()
+        {
+            return (this.Line, this.Ordinal).GetHashCode();
+        }
+
+        public override string ToString()
+        {
+            return JsonConvert.SerializeObject(this);
```
MarcoRossignoli: for simplicity I used Json serializer.

src/coverlet.core/Abstracts/IInstrumenter.cs
```diff
@@ -0,0 +1,126 @@
...
+
+    public class Branch : Line
+    {
+        public int Offset;
+        public int EndOffset;
+        public int Path;
+        public uint Ordinal;
+    }
+
+    public class BranchKeyConverter : TypeConverter
```
MarcoRossignoli: we need custom converter because Json serializer does not support serialization of complex dictionary key

MarcoRossignoli:
@tonerdo fixed conflicts...the change touch a lot of file if you can review this before others we avoid too much conflicts...I think(hope) this is the only "huge" update of checklist.

tonerdo:
@MarcoRossignoli this is a good start. I'm a bit skeptical about the drastic changes and I wonder if things could be simpler. The most important things that need to be serialized in the `Coverage` class is `identifier`, `_module`, `_mergeWith`, `_useSourceLink`, and `_results` (maybe `_logger`). I wonder if we can just extract those into their own class, serialize those and add a sort of "builder" method on `Coverage` that takes that class as a parameter.

Also, it's best not to do major style refactorings for important features so it doesn't get so large

MarcoRossignoli:
Replaced by https://github.com/tonerdo/coverlet/pull/409 keep this PR as useful notes.

PR #394, opened 2019-05-02
Use ConcurrentDictionary for static var

MarcoRossignoli:
follow up https://github.com/tonerdo/coverlet/pull/383

I prefer use concurrent collection for a static in case of concurrent test/build

/cc @tonerdo @ViktorHofer


ViktorHofer:
LGTM

PR #396, opened 2019-05-03
Make build more coherent with same "dotnet sdk" on all CI machine.

MarcoRossignoli:
Related to https://github.com/tonerdo/coverlet/pull/390#issuecomment-488669074

We have two script

```
// default is debug
./build.sh debug/release <- unix 
build.cmd debug/release  <- win
```

They download sdk to `reporoot\.tools\` and run dotnet cli from there, this should lead to more coherent build/test result because build is "machine dotnet runtime version agnostic".

/cc @tonerdo @petli


eng/build.sh
```diff
@@ -0,0 +1,71 @@
...
+echo "Downloading '$install_script_url'"
+
+ # Use curl if available, otherwise use wget
+if command -v curl > /dev/null; then
+  curl "$install_script_url" -sSL --retry 10 --create-dirs -o "$install_script"
+else
+  wget -q -O "$install_script" "$install_script_url"
+fi
+
+bash "$install_script" --version 2.2.203 --install-dir "$dotnet_root" || {
```
MarcoRossignoli: we could move version in a simple text file on eng folder and take version from there.

eng/build.ps1
```diff
@@ -0,0 +1,35 @@
...
+{
+    set-strictmode -version 2.0
+    $ErrorActionPreference = "Stop"
+    [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12  
+
+     $toolsRoot = Join-Path $PSScriptRoot "..\.tools"
+     Create-Directory $toolsRoot
+     $scriptPath = Join-Path $toolsRoot "dotnet-install.ps1"
+     Invoke-WebRequest "https://dot.net/v1/dotnet-install.ps1" -OutFile $scriptPath
+     & $toolsRoot\dotnet-install.ps1 -InstallDir $toolsRoot\dotnet -Version 2.2.203
```
MarcoRossignoli: we could move version in a simple text file on eng folder and take version from there.

MarcoRossignoli:
@tonerdo seems work, I moved docs under Documentation in line with `Join foundation` issue.

MarcoRossignoli:
@tonerdo  we should move to AzurePipelines https://azure.microsoft.com/en-us/blog/announcing-azure-pipelines-with-unlimited-ci-cd-minutes-for-open-source/ 
I use AzurePipelines in my day by day job...if you link the repo and add me as admin of pipeline I could try to setup.

MarcoRossignoli:
Another plus of this approach is that we can run CI on a list of SDK which could produce different IL and invalidate code assumptions and again in future we could have different instrumentation related to different SDK(we can read it runtime).

MarcoRossignoli:
@tonerdo PTAL

MarcoRossignoli:
@tonerdo merged and fixed azp

MarcoRossignoli:
@tonerdo  woot command works!

MarcoRossignoli:
@tonerdo I think it's ready to go!

AArnott:
I wonder if you're trying too hard. Your repo's build requirements really aren't that strange to justify all these custom build scripts.

Add a global.json file to your repo to "pin" the right .NET Core SDK used to build with. [Here's an example](https://github.com/microsoft/vs-threading/blob/master/src/global.json).

Then modify your azure pipelines YAML file to install that version. [Here's an example](https://github.com/microsoft/vs-threading/blob/8c9c7a44eb4ff6284b373fc416fc1e9dbfae4cab/azure-pipelines/build.yml#L18-L23).

MarcoRossignoli:
@AArnott the idea of download sdk is for future use...I mean instrumenting code and verify coverage depends on compiler version, different version of compilers generage different pdb and in future we would like to have a way to run tests with more than one sdk version...for instance take a look at this PR https://github.com/tonerdo/coverlet/issues/389 https://github.com/tonerdo/coverlet/pull/390 behaviour of hide sequence point is changed and our tests started to fail.

AArnott:
[Here's what it looks like in full for your repo](https://github.com/tonerdo/coverlet/compare/master...AArnott:PinSdkVersion?expand=1)

Disclaimer: not fully tested yet.

AArnott:
It's now tested.

MarcoRossignoli:
@AArnott do you have better way to provide a CI with more than one SDK?
My idea as I've explained above is the possibility to have more than one SDK, so download from a list and rebuild/test for every different version and in case amend tests with assertion related to version number(i.e. line x is not covered in version x.y.x due to different pdb).

AArnott:
Yes, I think we can handle that too.

My personal recommendation is that you *build* with exactly one SDK version (that you've pinned for consistency). Then you *test* on as many SDKs as you need your coverlet build extension to support for your target customers.

MarcoRossignoli:
Now CI depends on custom sdk that is one of the goal of this PR.
Close for the moment we'll pickup some script idea when/if we'll decide how to approach to this particular scenario.

Issue #382, opened 2019-04-25
Line coverage output should be rounded


ViktorHofer:
+------------------------+---------------------+--------+--------+
  | Module                 | Line                | Branch | Method |
  +------------------------+---------------------+--------+--------+
  | System.AppContext      | 100%                | 100%   | 100%   |
  +------------------------+---------------------+--------+--------+
  | System.Private.CoreLib | 14.000000000000002% | 9.5%   | 12%    |
  +------------------------+---------------------+--------+--------+

  +---------+---------------------+--------+--------+
  |         | Line                | Branch | Method |
  +---------+---------------------+--------+--------+
  | Total   | 14.000000000000002% | 9.5%   | 12%    |
  +---------+---------------------+--------+--------+
  | Average | 7.000000000000001%  | 4.75%  | 6%     |
  +---------+---------------------+--------+--------+

tonerdo:
@ViktorHofer do you propose a floor or a ceiling?

ViktorHofer:
Probably floor but I don't have a strong opinion here.

MarcoRossignoli:
I'm for floor

tonerdo:
The floors have it then. I think method and branch coverage could also benefit from this

sasivishnu:
Hi, 
If no one is working on the issue, can I take a look into it? 
Also, I think we use floor & keep it to maximum of 2 decimal places. What you think?Especially in bigger projects, It will be better to have 82.23% then 82%

MarcoRossignoli:
@sasivishnu sure give it a try!
I'm ok with 2 decimal places, but we can delay the choice at review time!
Let me know if you'll be in trouble with something.

sasivishnu:
Thanks, I will work on this weekend. 

I was able to build the project and will spend time understanding the flow. If there is any document/guide/tips on how to run & debug project locally will be helpful.

MarcoRossignoli:
You can use some of summary unit test https://github.com/tonerdo/coverlet/blob/master/test/coverlet.core.tests/CoverageSummaryTests.cs
Open `coverlet.sln` in VS and debug some test:
![image](https://user-images.githubusercontent.com/7894084/57176095-e7189680-6e53-11e9-8038-5bcfd7af3e06.png)

sasivishnu:
Since it is fixed, should we close this issue now?

MarcoRossignoli:
Yes thank's again!

Issue #346, opened 2019-02-18
Suspicious Average


bmccord:
We have an application with 5 projects.  I installed coverlet and ran the coverage.  We currently have 100% coverage (project is just starting).  Here are the results:

+------------------------------------------------+--------+--------+--------+
| Module                                         | Line   | Branch | Method |
+------------------------------------------------+--------+--------+--------+
| Project 1                                       | 100%   | 100%   | 100%   |
+------------------------------------------------+--------+--------+--------+
| Project 2                                       | 100%   | 100%   | 100%   |
+------------------------------------------------+--------+--------+--------+
| Project 3                                       | 100%   | 100%   | 100%   |
+------------------------------------------------+--------+--------+--------+
| Project 4                                       | 100%   | 100%   | 100%   |
+------------------------------------------------+--------+--------+--------+
| Project 5                                       | 100%   | 100%   | 100%   |
+------------------------------------------------+--------+--------+--------+

+---------+--------+--------+--------+
|                 | Line       | Branch | Method |
+---------+--------+--------+--------+
| Total        | 100%     | 100%    | 100%   |
+---------+--------+--------+--------+
| Average  | 20%       | 20%      | 20%    |
+---------+--------+--------+--------+

This doesn't seem right to me.  It's 100% coverage across the board for each project, but the average is only 20% across the board.  I understand how it is getting it (100 / 5), but I don't think the average is an accurate reflection.

It seems that the average should also be 100%.

tonerdo:
@bmccord I think this might be a side effect of how Coverlet displays results. For example, if the code contains no branches, Coverlet will display 100% instead of 0%. However, when calculating the total and average it'll use the `0` value.

bmccord:
@tonerdo Is that the way it should function or is this considered a bug?

danielloganking:
I was also wondering if excluded projects had something to do with this?

diegofox2:
Hi guys. I see this bug still opened, is this right?

AlexTeixeira:
Same issue for me. 

+-------------------------------+------+--------+--------+
| Module                        | Line | Branch | Method |
+-------------------------------+------+--------+--------+
| GardenDynamics.Infrastructure | 100% | 100%   | 100%   |
+-------------------------------+------+--------+--------+
| GardenDynamics.Domain         | 100% | 100%   | 100%   |
+-------------------------------+------+--------+--------+
| GardenDynamics.Application    | 100% | 100%   | 100%   |
[...]

PR #397, opened 2019-05-04
Coverage percentage output value rounded down with max 2 decimal precision

sasivishnu:
[Reports-After_Change.zip](https://github.com/tonerdo/coverlet/files/3148605/Reports-After_Change.zip)
[Reports-Before_Change.zip](https://github.com/tonerdo/coverlet/files/3148606/Reports-Before_Change.zip)

Related to issue #382 

Change Details:
1. All percentage output are rounded down (Floor) with maximum of 2 decimal precision.
2. Arithmetic precision error fixed by rounding- ex: `0.1400D * 100 = 14.000000000000002`
3. In Cobertura Report, "line-rate" and "branch-rate" value will have an extra decimal precision
Before: 0.143
After: 0.1432
4. In OpenCover report, "sequenceCoverage" and "branchCoverage" value will have an extra decimal precision
Before: 33.3
After: 33.33
5. In TeamCitry report, "CodeCoverageR" value will have an extra decimal precision
Before: 33.3
After: 33.33

I also updated and added new test cases to cover for the changes. Previously, we are rounding with max 3 decimal points before multiplying with 100. This yields value like 81.5%. I had to increase that to max 4 decimal points to yield like 81.52% when multiplied with 100%. In large projects, two decimal point reporting is useful and allows for incremental valuation without too much data loss after rounding down. Also I think, 2 decimal precision is commonly accepted standard.

Let me know If I missed something. Thanks

Edit: I have attached output for reports before my change (Coverlet.Console version 1.5.0) and after my change in zip.

Sample console output in version 1.5.0:
```
+-----------+-------+--------+--------+
| Module    | Line  | Branch | Method |
+-----------+-------+--------+--------+
| AppCenter | 13.8% | 25%    | 50%    |
+-----------+-------+--------+--------+

+---------+-------+--------+--------+
|         | Line  | Branch | Method |
+---------+-------+--------+--------+
| Total   | 13.8% | 25%    | 50%    |
+---------+-------+--------+--------+
| Average | 13.8% | 25%    | 50%    |
+---------+-------+--------+--------+
```

Sample console output after my change:
```
+-----------+--------+--------+--------+
| Module    | Line   | Branch | Method |
+-----------+--------+--------+--------+
| AppCenter | 13.79% | 25%    | 50%    |
+-----------+--------+--------+--------+

| Total   | 13.79% | 25%    | 50%    |
| Average | 13.79% | 25%    | 50%    |
```

### DIFF
--- a/src/coverlet.console/Program.cs
+++ b/src/coverlet.console/Program.cs
@@ -156,15 +156,15 @@ namespace Coverlet.Console
                 var summary = new CoverageSummary();
                 int numModules = result.Modules.Count;
 
-                var totalLinePercent = summary.CalculateLineCoverage(result.Modules).Percent * 100;
-                var totalBranchPercent = summary.CalculateBranchCoverage(result.Modules).Percent * 100;
-                var totalMethodPercent = summary.CalculateMethodCoverage(result.Modules).Percent * 100;
+                var totalLinePercent = summary.CalculateLineCoverage(result.Modules).GetCoveragePercentage();
+                var totalBranchPercent = summary.CalculateBranchCoverage(result.Modules).GetCoveragePercentage();
+                var totalMethodPercent = summary.CalculateMethodCoverage(result.Modules).GetCoveragePercentage();
 
                 foreach (var _module in result.Modules)
                 {
-                    var linePercent = summary.CalculateLineCoverage(_module.Value).Percent * 100;
-                    var branchPercent = summary.CalculateBranchCoverage(_module.Value).Percent * 100;
-                    var methodPercent = summary.CalculateMethodCoverage(_module.Value).Percent * 100;
+                    var linePercent = summary.CalculateLineCoverage(_module.Value).GetCoveragePercentage();
+                    var branchPercent = summary.CalculateBranchCoverage(_module.Value).GetCoveragePercentage();
+                    var methodPercent = summary.CalculateMethodCoverage(_module.Value).GetCoveragePercentage();
 
                     coverageTable.AddRow(Path.GetFileNameWithoutExtension(_module.Key), $"{linePercent}%", $"{branchPercent}%", $"{methodPercent}%");
                 }
--- a/src/coverlet.console/Program.cs
+++ b/src/coverlet.console/Program.cs
@@ -156,15 +156,15 @@ namespace Coverlet.Console
                 var summary = new CoverageSummary();
                 int numModules = result.Modules.Count;
 
-                var totalLinePercent = summary.CalculateLineCoverage(result.Modules).GetCoveragePercentage();
-                var totalBranchPercent = summary.CalculateBranchCoverage(result.Modules).GetCoveragePercentage();
-                var totalMethodPercent = summary.CalculateMethodCoverage(result.Modules).GetCoveragePercentage();
+                var totalLinePercent = summary.CalculateLineCoverage(result.Modules).Percent;
+                var totalBranchPercent = summary.CalculateBranchCoverage(result.Modules).Percent;
+                var totalMethodPercent = summary.CalculateMethodCoverage(result.Modules).Percent;
 
                 foreach (var _module in result.Modules)
                 {
-                    var linePercent = summary.CalculateLineCoverage(_module.Value).GetCoveragePercentage();
-                    var branchPercent = summary.CalculateBranchCoverage(_module.Value).GetCoveragePercentage();
-                    var methodPercent = summary.CalculateMethodCoverage(_module.Value).GetCoveragePercentage();
+                    var linePercent = summary.CalculateLineCoverage(_module.Value).Percent;
+                    var branchPercent = summary.CalculateBranchCoverage(_module.Value).Percent;
+                    var methodPercent = summary.CalculateMethodCoverage(_module.Value).Percent;
 
                     coverageTable.AddRow(Path.GetFileNameWithoutExtension(_module.Key), $"{linePercent}%", $"{branchPercent}%", $"{methodPercent}%");
                 }
--- a/src/coverlet.core/CoverageDetails.cs
+++ b/src/coverlet.core/CoverageDetails.cs
@@ -6,6 +6,6 @@ namespace Coverlet.Core
     {
         public double Covered { get; internal set; }
         public int Total { get; internal set; }
-        public double Percent => Total == 0 ? 1D : Math.Floor((Covered / Total) * 10000) / 100;
+        public double Percent => Total == 0 ? 100D : Math.Floor((Covered / Total) * 10000) / 100;
     }
 }
--- a/src/coverlet.core/CoverageDetails.cs
+++ b/src/coverlet.core/CoverageDetails.cs
@@ -8,7 +8,18 @@ namespace Coverlet.Core
         public int Total { get; internal set; }
         public double Percent
         {
-            get => Math.Round(Total == 0 ? 1 : Covered / Total, 3);
+            get => Math.Round(Total == 0 ? 1 : Covered / Total, 4);
+        }
+
+        public double GetCoveragePercentage()
+        {
+            double percentage = Percent * 100;
+            return RoundDown(percentage);
+        }
+
+        private double RoundDown(double percentage)
+        {
+            return Math.Floor(percentage * 100) / 100;
         }
     }
 }
--- a/src/coverlet.core/CoverageDetails.cs
+++ b/src/coverlet.core/CoverageDetails.cs
@@ -6,20 +6,6 @@ namespace Coverlet.Core
     {
         public double Covered { get; internal set; }
         public int Total { get; internal set; }
-        public double Percent
-        {
-            get => Math.Round(Total == 0 ? 1 : Covered / Total, 4);
-        }
-
-        public double GetCoveragePercentage()
-        {
-            double percentage = Percent * 100;
-            return RoundDown(percentage);
-        }
-
-        private double RoundDown(double percentage)
-        {
-            return Math.Floor(percentage * 100) / 100;
-        }
+        public double Percent => Total == 0 ? 1D : Math.Floor((Covered / Total) * 10000) / 100;
     }
 }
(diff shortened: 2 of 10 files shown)


src/coverlet.core/CoverageDetails.cs
```diff
@@ -8,7 +8,18 @@ public class CoverageDetails
         public int Total { get; internal set; }
         public double Percent
         {
-            get => Math.Round(Total == 0 ? 1 : Covered / Total, 3);
+            get => Math.Round(Total == 0 ? 1 : Covered / Total, 4);
+        }
+
+        public double GetCoveragePercentage()
```
MarcoRossignoli: Please use property `CoveragePercentace { get }`

src/coverlet.core/CoverageDetails.cs
```diff
@@ -8,7 +8,18 @@ public class CoverageDetails
...
+        {
+            double percentage = Percent * 100;
+            return RoundDown(percentage);
+        }
+
+        private double RoundDown(double percentage)
```
MarcoRossignoli: Can you avoid `RoundDown` helper and use inline directly on `CoveragePercentace` prop, if I'm not mistaken this helper is private and used only in one method so it's not so useful extract an helper method.

MarcoRossignoli:
@sasivishnu can you please run before/after coverage with all format and show to PR(like comment or attaching files) we need to be sure that reports are ok.

src/coverlet.core/CoverageDetails.cs
sasivishnu: Yeah, I can do that. Will convert method into prop CoveragePercentage with getter only.

src/coverlet.core/CoverageDetails.cs
sasivishnu: Yes, its used only in one place as of now. 

I forget to ask this earlier, Average Coverage Percentage value could go more than 2 decimal precision since its calculated inline. 
`coverageTable.AddRow("Average", $"{totalLinePercent / numModules}%", $"{totalBranchPercent / numModules}%", $"{totalMethodPercent / numModules}%");`
Should I Floor the value with 2 decimal precision there also? Should I create `MathHelper` class under Helper folder, add this `RoundDown `method in the class & refer it in `CoveragePercentage `property and also during Average calculation? Is this fine?

sasivishnu:
> 
> 
> @sasivishnu can you please run before/after coverage with all format and show result to PR(like comment or attaching files) we need to be sure that reports are ok.

@MarcoRossignoli Yeah, I can do that. I will attach before and after reports for all five formats with console output too.

src/coverlet.core/CoverageDetails.cs
sasivishnu: @MarcoRossignoli I have other changes ready and also before & after report formats output. 

I am waiting to see reply on what to do with "Average Percentage" calculation and then check-in code fully at once.

MarcoRossignoli: I did deeper review...and I think that we should change a bit the idea.
We should avoid to change `CoverageDetail` object only to "round data", I don't like the idea to have members with same meaning but only different "formatting" seems a duplication to me, I mean `GetCoveragePercentage` is `Percent` formatted.
I like the idea of `double MathHelper.RoundPercentage(double)` in a file MathHelper.cs under Helpers folder and apply this helper in every place we need, in this way we can also apply it on `Average Coverage Percentage`.
For instance 
```
var totalLinePercent =  MathHelper.RoundPercentage(summary.CalculateLineCoverage(result.Modules).Percent * 100);
...
var method = MathHelper.RoundPercentage(summary.CalculateMethodCoverage(module.Value));
...
 coverageTable.AddRow("Average", $"{MathHelper.RoundPercentage(totalLinePercent / numModules)}%",...
```
In this way we have also cetralized "percentage rounding" that we can apply in a uniform way in all code base project.

MarcoRossignoli: could be cute to have an extension method
```
namespace Coverlet.Core.Helpers
{
...
public static double RoundPercentage(this double value)
...
```

sasivishnu: Cool :) .. I too like the idea of extension method.

I will rewrite and submit the code tomorrow then.

MarcoRossignoli: No rush take your time thank's and sorry for "change of direction"

test/coverlet.core.tests/Reporters/OpenCoverReporterTests.cs
```diff
@@ -17,13 +17,13 @@ public void TestReport()
             result.Identifier = Guid.NewGuid().ToString();
 
             result.Modules = new Modules();
-            result.Modules.Add("Coverlet.Core.Reporters.Tests", CreateFirstDocuments());
+            result.Modules.Add("Coverlet.Core.Reporters.Tests", CreateFirstDocuments());
```
tonerdo:
```suggestion
            result.Modules.Add("Coverlet.Core.Reporters.Tests", CreateFirstDocuments());
```

src/coverlet.core/CoverageDetails.cs
tonerdo: @sasivishnu @MarcoRossignoli it'll probably be more straightforward to just update the `Percent` property to automatically multiply by 100 and round down. I don't think there's really any need to introduce a new `GetCoveragePercentage` method. Regarding calculation for the Average Coverage Percentage, you can leave it how it is now, I'll work on it once this PR is merged as I need to address #346

src/coverlet.core/CoverageDetails.cs
```diff
@@ -8,7 +8,18 @@ public class CoverageDetails
         public int Total { get; internal set; }
         public double Percent
         {
-            get => Math.Round(Total == 0 ? 1 : Covered / Total, 3);
+            get => Math.Round(Total == 0 ? 1 : Covered / Total, 4);
```
tonerdo: All rounding down and decimal place code should go here

src/coverlet.console/Program.cs
```diff
@@ -156,15 +156,15 @@ static int Main(string[] args)
                 var summary = new CoverageSummary();
                 int numModules = result.Modules.Count;
 
-                var totalLinePercent = summary.CalculateLineCoverage(result.Modules).Percent * 100;
-                var totalBranchPercent = summary.CalculateBranchCoverage(result.Modules).Percent * 100;
-                var totalMethodPercent = summary.CalculateMethodCoverage(result.Modules).Percent * 100;
+                var totalLinePercent = summary.CalculateLineCoverage(result.Modules).GetCoveragePercentage();
```
tonerdo: This can just be `var totalLinePercent = summary.CalculateLineCoverage(result.Modules).Percent` if you update the `Percent` property to contain the necessary calculations. Applies everywhere this new method is used

src/coverlet.core/CoverageDetails.cs
sasivishnu: @MarcoRossignoli no issues. I was also deliberating while fixing since it can be done in multiple ways. It needed this kind of discussion.

@tonerdo yes, that will be straight forward. 

Then, we need to make change in Cobertura Report on "line-rate" and "branch-rate" column.

Before: 
```
coverage.Add(new XAttribute("line-rate", summary.CalculateLineCoverage(result.Modules).Percent.ToString()));
coverage.Add(new XAttribute("branch-rate", summary.CalculateBranchCoverage(result.Modules).Percent.ToString()));
```

After:
```
coverage.Add(new XAttribute("line-rate", (summary.CalculateLineCoverage(result.Modules).Percent / 100).ToString()));
coverage.Add(new XAttribute("branch-rate", (summary.CalculateBranchCoverage(result.Modules).Percent / 100).ToString()));
```

Its not going to cause any impact. Since we already rounded down, dividing by 100 is not going to make any new issue. 

If all is fine, then I will make a change in `Percent `property itself.

MarcoRossignoli: The idea of helper was for future possible use of rounding percentage everywhere in same way, btw I'm ok also with this with "fix" where we need like above.

sasivishnu: I am planning to do both. Use `Percent `property only but add extension method in new `MathHelper.cs` class as `public static double RoundPercentage(this double value)`.

MarcoRossignoli: @sasivishnu avoid helper at all, less code to review and maintain, if in future we'll need to reuse we'll move to helper

tonerdo: Thanks @sasivishnu, let me know when you'd done this so I can merge. The rest of the code seems good to me

sasivishnu: @tonerdo Yep, I am done with my changes and pushed latest one.

sasivishnu:
> 
> 
> @sasivishnu can you please run before/after coverage with all format and show result to PR(like comment or attaching files) we need to be sure that reports are ok.

@MarcoRossignoli I have attached before and after sample report in [first comment](https://github.com/tonerdo/coverlet/pull/397#issue-275941054)

src/coverlet.core/CoverageDetails.cs
```diff
@@ -6,9 +6,6 @@ public class CoverageDetails
     {
         public double Covered { get; internal set; }
         public int Total { get; internal set; }
-        public double Percent
-        {
-            get => Math.Round(Total == 0 ? 1 : Covered / Total, 3);
-        }
+        public double Percent => Total == 0 ? 1D : Math.Floor((Covered / Total) * 10000) / 100;
```
MarcoRossignoli: should be `100D`?

src/coverlet.core/Reporters/CoberturaReporter.cs
```diff
@@ -26,7 +26,7 @@ public string Report(CoverageResult result)
             XDocument xml = new XDocument();
             XElement coverage = new XElement("coverage");
             coverage.Add(new XAttribute("line-rate", (summary.CalculateLineCoverage(result.Modules).Percent / 100).ToString()));
-            coverage.Add(new XAttribute("branch-rate", (summary.CalculateLineCoverage(result.Modules).Percent / 100).ToString()));
+            coverage.Add(new XAttribute("branch-rate", (summary.CalculateBranchCoverage(result.Modules).Percent / 100).ToString()));
```
MarcoRossignoli: good catch thank's!

MarcoRossignoli:
File compared and looks ok to me there is only expected extra decimal for cobertura `line-rate="0.1379"`

src/coverlet.core/CoverageDetails.cs
sasivishnu: Yeah, good catch. Thanks. Will push the new commit with fix

PR #412, opened 2019-05-15
Different exit codes to indicate particular failures

### LINKED ISSUE #388: Exit code for coverlet should communicate different failures
Today, when a test fails or code coverage fails, a non-zero exit code is returned. This is desirable because it allows build automation systems to fail when the coverlet command fails. Unfortunately, the return code is always 1 so it doesn't provide detail about what went wrong. One could, theoretically, grep the output of coverlet for things like "the minimum line coverage is below specified 95", but this is brittle and unnecessary (since that is already written to stdout and can be inspected by a human). 

To better support build automation systems it'd be great if coverlet exited with specific non-zero exit codes to communicate why the build failed so that programmatic action can be taken upon that code without needing to do something hacky like grep the output. 

Here's a proposal:
* Return 2 when a test fails
* Return 3 when coverage is below the configured threshold
* Return 5 when both a test fails AND coverage is below the configured threshold
* Updated documentation reflecting this behavior

I'm not married to the actual numbers, just that they are specific and documented.

sasivishnu:
Closes #388 

**Exit Codes:**
- 0 : Success
- 1 : Test failed (based on dotnet test)
- 2 : Coverage percentage is below any threshold type
- 3 : Test failed and Coverage percentage is below any threshold type

- 101: General exception during coverlet process
- 102: Missing options or invalid arguments for coverlet process

Attached the output screenshot for all the above scenarios.  
[ExitCodes.zip](https://github.com/tonerdo/coverlet/files/3184006/ExitCodes.zip)

For any exception related code, we can use values starting from 101. We can use 1-100 for any of custom failures from the process. I didn't wanted to go negative for exceptions since if someone is currently using `exitcode > 0` to determine error cases in their script, then we will be breaking them.

### DIFF
--- a/src/coverlet.console/Program.cs
+++ b/src/coverlet.console/Program.cs
@@ -14,43 +14,7 @@ using McMaster.Extensions.CommandLineUtils;
 namespace Coverlet.Console
 {
     class Program
-    {
-        /// <summary>
-        /// Exit Codes returned from Coverlet console process.
-        /// </summary>
-        enum CommandExitCodes
-        {
-            /// <summary>
-            /// Indicates successful run of dotnet test without any test failure and coverage percentage above threshold if provided.
-            /// </summary>
-            Success = 0,
-
-            /// <summary>
-            /// Indicates test failure by dotnet test.
-            /// </summary>
-            TestFailed = 1,
-
-            /// <summary>
-            /// Indicates coverage percentage is below given threshold for one or more threshold type.
-            /// </summary>
-            CoverageBelowThreshold = 2,
-
-            /// <summary>
-            /// Indicates test failure by dotnet test and coverage percentage is below given threshold for one or more threshold type.
-            /// </summary>
-            TestFailedAndCoverageBelowThreshold = TestFailed + CoverageBelowThreshold,
-
-            /// <summary>
-            /// Indicates exception occurred during Coverlet process.
-            /// </summary>
-            Exception = 101,
-
-            /// <summary>
-            /// Indicates missing options or empty arguments for Coverlet process.
-            /// </summary>
-            CommandParsingException = 102
-        }
-
+    {        
         static int Main(string[] args)
         {
             var logger = new ConsoleLogger();
@@ -59,7 +23,7 @@ namespace Coverlet.Console
             app.FullName = "Cross platform .NET Core code coverage tool";
             app.HelpOption("-h|--help");
             app.VersionOption("-v|--version", GetAssemblyVersion());
-            int exitCode = 0;
+            int exitCode = (int)CommandExitCodes.Success;
 
             CommandArgument module = app.Argument("<ASSEMBLY>", "Path to the test assembly.");
             CommandOption target = app.Option("-t|--target", "Path to the test runner application.", CommandOptionType.SingleValue);
--- a/src/coverlet.console/Program.cs
+++ b/src/coverlet.console/Program.cs
@@ -14,7 +14,7 @@ using McMaster.Extensions.CommandLineUtils;
 namespace Coverlet.Console
 {
     class Program
-    {        
+    {
         static int Main(string[] args)
         {
             var logger = new ConsoleLogger();
--- a/src/coverlet.console/Program.cs
+++ b/src/coverlet.console/Program.cs
@@ -15,6 +15,42 @@ namespace Coverlet.Console
 {
     class Program
     {
+        /// <summary>
+        /// Exit Codes returned from Coverlet console process.
+        /// </summary>
+        enum CommandExitCodes
+        {
+            /// <summary>
+            /// Indicates successful run of dotnet test without any test failure and coverage percentage above threshold if provided.
+            /// </summary>
+            Success = 0,
+
+            /// <summary>
+            /// Indicates test failure by dotnet test.
+            /// </summary>
+            TestFailed = 1,
+
+            /// <summary>
+            /// Indicates coverage percentage is below given threshold for one or more threshold type.
+            /// </summary>
+            CoverageBelowThreshold = 2,
+
+            /// <summary>
+            /// Indicates test failure by dotnet test and coverage percentage is below given threshold for one or more threshold type.
+            /// </summary>
+            TestFailedAndCoverageBelowThreshold = TestFailed + CoverageBelowThreshold,
+
+            /// <summary>
+            /// Indicates exception occurred during Coverlet process.
+            /// </summary>
+            Exception = 101,
+
+            /// <summary>
+            /// Indicates missing options or empty arguments for Coverlet process.
+            /// </summary>
+            CommandParsingException = 102
+        }
+
         static int Main(string[] args)
         {
             var logger = new ConsoleLogger();
@@ -23,6 +59,7 @@ namespace Coverlet.Console
             app.FullName = "Cross platform .NET Core code coverage tool";
             app.HelpOption("-h|--help");
             app.VersionOption("-v|--version", GetAssemblyVersion());
+            int exitCode = 0;
 
             CommandArgument module = app.Argument("<ASSEMBLY>", "Path to the test assembly.");
             CommandOption target = app.Option("-t|--target", "Path to the test runner application.", CommandOptionType.SingleValue);
@@ -185,10 +222,14 @@ namespace Coverlet.Console
                 coverageTable.AddRow("Average", $"{totalLinePercent / numModules}%", $"{totalBranchPercent / numModules}%", $"{totalMethodPercent / numModules}%");
 
                 logger.LogInformation(coverageTable.ToStringAlternative());
-
+                if (process.ExitCode > 0)
+                {
+                    exitCode = (int)CommandExitCodes.TestFailed;
+                }
                 thresholdTypeFlags = result.GetThresholdTypesBelowThreshold(summary, dThreshold, thresholdTypeFlags, dThresholdStat);
                 if (thresholdTypeFlags != ThresholdTypeFlags.None)
                 {
+                    exitCode += (int)CommandExitCodes.CoverageBelowThreshold;
                     var exceptionMessageBuilder = new StringBuilder();
                     if ((thresholdTypeFlags & ThresholdTypeFlags.Line) != ThresholdTypeFlags.None)
                     {
@@ -208,7 +249,7 @@ namespace Coverlet.Console
                     throw new Exception(exceptionMessageBuilder.ToString());
                 }
 
-                return process.ExitCode == 0 ? 0 : process.ExitCode;
+                return exitCode;
             });
 
             try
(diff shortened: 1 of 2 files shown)


src/coverlet.console/Program.cs
```diff
@@ -15,6 +15,42 @@ namespace Coverlet.Console
...
+            /// Indicates exception occurred during Coverlet process.
+            /// </summary>
+            Exception = 101,
+
+            /// <summary>
+            /// Indicates missing options or empty arguments for Coverlet process.
+            /// </summary>
+            CommandParsingException = 102
+        }
+
```
sasivishnu: I created for enum for exit codes to have code level documentation for reference and for ease of change in future.

MarcoRossignoli:
Thanks I'm OOF at the moment I'll take a look ASAP

src/coverlet.console/Program.cs
```diff
@@ -15,6 +15,42 @@ namespace Coverlet.Console
 {
     class Program
     {
+        /// <summary>
+        /// Exit Codes returned from Coverlet console process.
+        /// </summary>
+        enum CommandExitCodes
```
MarcoRossignoli: I like the idea of enum, but I prefer to have a custom file `ExitCodes.cs` with internal enum inside, the "real program logic" is clearer(this is only a struct).

src/coverlet.console/Program.cs
```diff
@@ -185,10 +222,14 @@ static int Main(string[] args)
                 coverageTable.AddRow("Average", $"{totalLinePercent / numModules}%", $"{totalBranchPercent / numModules}%", $"{totalMethodPercent / numModules}%");
 
                 logger.LogInformation(coverageTable.ToStringAlternative());
-
+                if (process.ExitCode > 0)
```
MarcoRossignoli: why not `== 1`?

src/coverlet.console/Program.cs
```diff
@@ -15,6 +15,42 @@ namespace Coverlet.Console
...
+
+            /// <summary>
+            /// Indicates coverage percentage is below given threshold for one or more threshold type.
+            /// </summary>
+            CoverageBelowThreshold = 2,
+
+            /// <summary>
+            /// Indicates test failure by dotnet test and coverage percentage is below given threshold for one or more threshold type.
+            /// </summary>
+            TestFailedAndCoverageBelowThreshold = TestFailed + CoverageBelowThreshold,
```
MarcoRossignoli: I would remove this...it's not used.

sasivishnu: Its not directly used but code value of 3 represent this. I thought by adding this in enum, we are kind of reserving and making meaning of exit code value 3 clearer for any reference.

src/coverlet.console/Program.cs
MarcoRossignoli: This should be a `[Flags]` i.e. https://www.ideone.com/9OvttX https://docs.microsoft.com/en-us/dotnet/api/system.flagsattribute?view=netframework-4.8

src/coverlet.console/Program.cs
MarcoRossignoli: With `Flags` this "meaning" is implicit.

src/coverlet.console/Program.cs
sasivishnu: I did read dotnet tests return 0 for test succees and 1 for any failure. But to be on safer side, I went for anything > 0.. I can change it.

src/coverlet.console/Program.cs
sasivishnu: Cool, I forget about it. I can remove `TestFailedAndCoverageBelowThreshold` from enum with this.

src/coverlet.console/Program.cs
MarcoRossignoli: For curiosity, can you try to run with "build fail" and check number?If it's different than `1` we could add and else with different type error.

src/coverlet.console/Program.cs
sasivishnu: Since discussion was to create only for console application, I didn't abstract much. Should I create `ExitCodes.cs` in Coverlet.Core project? or still in Coverlet.Console project?

src/coverlet.console/Program.cs
```diff
@@ -15,6 +15,42 @@ namespace Coverlet.Console
...
+
+            /// <summary>
+            /// Indicates exception occurred during Coverlet process.
+            /// </summary>
+            Exception = 101,
```
MarcoRossignoli: Why not simply 4,5 for exceptions?

sasivishnu: If incase tomorrow, we add enhancement to return different exit code for different Threadhold failures, then we can make use of values from 4. If we make Exceptions 4,5 now and then when we that enhancement, order might look. Changing the value then, will raise compatibility issue when people update to new version. 

Idea was 2-100 series represents "after process" failure exit codes.
> 100, represents "during process" failure exit codes.

src/coverlet.console/Program.cs
sasivishnu: Its returning 1 if build fails. But aren't its requirement that in Console tool, we should run with `--no-build` args?

src/coverlet.console/Program.cs
MarcoRossignoli: `Coverlet.Console`

src/coverlet.console/Program.cs
MarcoRossignoli: > But aren't its requirement that in Console tool, we should run with --no-build args?

Yes I was only curious...ok keep `> 0` at the moment I don't see particular problems

src/coverlet.console/Program.cs
MarcoRossignoli: > Changing the value then, will raise compatibility issue when people update to new version.

Yup...as I said I don't know if will be code that verify all these possible "error codes", btw I won't oppose to this if also @tonerdo agree.

sasivishnu: I will wait for @tonerdo comments before making changes. So, I can submit all code review changes in single commit.

sasivishnu: Ping @tonerdo ... Waiting for your confirmation...

sasivishnu: @tonerdo I assume since you had reviewed, you are fine with having different starting values for exceptions.

sasivishnu:
Committed changes to PR as per code review changes.

MarcoRossignoli:
I'll take a look this evening after dinner, thank's!

src/coverlet.console/Program.cs
```diff
@@ -185,10 +186,14 @@ static int Main(string[] args)
                 coverageTable.AddRow("Average", $"{totalLinePercent / numModules}%", $"{totalBranchPercent / numModules}%", $"{totalMethodPercent / numModules}%");
 
                 logger.LogInformation(coverageTable.ToStringAlternative());
-
+                if (process.ExitCode > 0)
+                {
+                    exitCode = (int)CommandExitCodes.TestFailed;
```
MarcoRossignoli:
```suggestion
                    exitCode += (int)CommandExitCodes.TestFailed;
```

src/coverlet.console/Program.cs
MarcoRossignoli: @tonerdo what do you think about > 100 for errors?I don't have a strong opinion...I would say ok...but I don't want fail this choice.

MarcoRossignoli:
@sasivishnu have you retested with flags?Is it ok?

MarcoRossignoli:
PS. We have to remember after this PR and https://github.com/tonerdo/coverlet/pull/422 merge we'll need to update console docs, this lead to increase minor ver as new feature.

src/coverlet.console/Program.cs
sasivishnu: I am fine with making this change. Just for my understanding. Is it because, we changed to use Flags in Enum, doing `+=` makes sense than `=` even though it be first time value is asserted based on some condition?

sasivishnu:
> 
> 
> @sasivishnu have you retested with flags?Is it ok?

Yes, everything is working fine.

src/coverlet.console/Program.cs
MarcoRossignoli: Yes nothing particular, only keep the "idea" coherent.
If in future we add something above is less prone to error.

src/coverlet.console/Program.cs
sasivishnu: ping @tonerdo  ...  what do think about Exception having > 100 exit codes?

MarcoRossignoli:
@sasivishnu until Toni has got some time to review could you also update .net tool guide with this new feature? https://github.com/tonerdo/coverlet/blob/master/Documentation/GlobalTool.md

sasivishnu:
> 
> 
> @sasivishnu until @tonerdo has got some time to review could you also update .net tool guide with this new feature? https://github.com/tonerdo/coverlet/blob/master/Documentation/GlobalTool.md

Yeah, sure... I was just waiting for @tonerdo comments on exit codes for Exceptions. Once its confirmed, then I thought to update the doc.

Anyway, I will update the doc and commit it with exit codes as of now. If any changes, then I will update the doc also.

sasivishnu:
> 
> 
> > @sasivishnu until @tonerdo has got some time to review could you also update .net tool guide with this new feature? https://github.com/tonerdo/coverlet/blob/master/Documentation/GlobalTool.md
> 
> Yeah, sure... I was just waiting for @tonerdo comments on exit codes for Exceptions. Once its confirmed, then I thought to update the doc.
> 
> Anyway, I will update the doc and commit it with exit codes as of now. If any changes, then I will update the doc also.

@MarcoRossignoli Seems like this pull request branch was created before the commit of https://github.com/tonerdo/coverlet/blob/master/Documentation/GlobalTool.md 

I will create separate pull request for updating doc.

MarcoRossignoli:
Ok no worries

src/coverlet.console/Program.cs
tonerdo: This is fine

src/coverlet.console/Program.cs
```diff
@@ -14,7 +14,7 @@
 namespace Coverlet.Console
 {
     class Program
-    {
+    {
```
tonerdo:
```suggestion
    {
```

src/coverlet.console/Program.cs
sasivishnu: Thanks @tonerdo @MarcoRossignoli

MarcoRossignoli: Thank you and sorry for the delay

MarcoRossignoli:
@tonerdo can we merge?

sasivishnu:
@MarcoRossignoli updated doc for this change #478

Issue #405, opened 2019-05-08
Use File.Exists instead of rely on exception in ModuleTrackerTemplate


MarcoRossignoli:
When `UnloadModule` is called to persist hits result we rely on exception handling to understand if create file or update.
Files are protected by global mutex so should be safe avoid slow exception handling and use File.Exists

https://github.com/tonerdo/coverlet/blob/d1a30cce46b42f8da49e762074cf6857b369b7d6/src/coverlet.core/Instrumentation/ModuleTrackerTemplate.cs#L79-L103

I think that perf cost is negligible against file write, btw we should improve code when possible

thank's to @joemey for reporting this

/cc @tonerdo @petli

sasivishnu:
Hi @MarcoRossignoli 

I can do it but it seems simple and also not an urgent issue to fix, may be can we mark this as good-first-issue? Just a thought. Otherwise, I can submit PR for this.

MarcoRossignoli:
Yes it's an enhancement, feel free to submit a PR

MarcoRossignoli:
fixed in https://github.com/tonerdo/coverlet/pull/415

PR #415, opened 2019-05-19
Using File.Exists for HitsFile creation/update validation

sasivishnu:
Related to #405

Changed to if-else condition instead of try-catch for handling hits file create/update scenario.


MarcoRossignoli:
@sasivishnu thank's!

Issue #212, opened 2018-10-12
Coverlet.Core as NuGet package.


p10tyr:
Is it possible to offer Coverlet.Core as a NuGet package please.

This will help in development of other extensions, tools and just general consumption of your amazing tool.

As you know, I would like to include automatic code coverage in my Extension but the only way I could to that is to clone your repo and then build the Core NuGet my self and consume that.  This creates a problem that I will be missing out on updates.

Also as discussed in this PR https://github.com/tonerdo/coverlet/pull/105 it would be better if this functionality consumed Coverlet.Core -- instead of adding that functionality INTO Coverlet.Core -- It is the same as if I wanted to add VSIX into Coverlet.Core.

Please consider this.
Thanks again! 🥇

tonerdo:
@ppumkin what about the alternative of just wrapping `coverlet.console` instead? Opening up `Coverlet.Core` will require changing things around a little bit as all the important stuff is either internal/private

ppumkin:
I did try this to use the Coverlet.Console package but I get the error

`Package 'coverlet.console 1.2.0' has a package type 'DotnetTool' that is not supported by project 'PrestoCoverage'.				

This is why I downloaded your branch and pushed it into my AzureDevops and built the Coverelt.Core package hosted in my private feed. I just need to do a subset of things you are doing in the Console

So in the extension I can just use this

```
Coverlet.Core.Coverage coverage = new Coverlet.Core.Coverage(fullPath, new string[0], new string[0], new string[0], string.Empty);

coverage.PrepareModules();

var result = coverage.GetCoverageResult();
```

I am not sure what changes you would need to do to  "open it up" ? It is fine jsut as it is now, I do not need anything else then to tell it where the dll is and what the coverage it.

I will automate the test execution using VS IDE API and then I can use Coverlet to get the Coverage directly from memory.

I think that building and publishing `Coverlet.Core` will just promote your package even more since you are allowing it to be used in other applications.

I am trying to understand why you would not want to publish the package so the options that I can see now are: (in order of preference) 

- *1 - You publish Coverlet.Core onto NuGet and allow people to consume it
- (99999) - I build and publish it to NuGet on your behalf so that who ever build my Extension can download the dependency
- (100000000) - I download the source project and include it inside my extension project so that it is built together

The only reason I ask for you to consider integrating that into your CI/CD pipeline is to avoid partitioning of your Package.

Thanks again.

singhsarab:
[...]

PR #420, opened 2019-05-23
Improve instrumentation tests

MarcoRossignoli:

@tonerdo @petli with this PR we've a way to test the "reality" of instrumentation and coverage calculation.

What we have:
1) a new helper class `RemoteExecutor.Invoke` that allow us to run a "lambda" in remote process to avoid issue with statics and file locks 🚀 🚀 🚀 
2) new possible assertions on line/branches coverage
3) we can "debug" all process, no more repro local build and reference+Debug.Launch...we can add "case" on test code and reproduce, fix and keep and run on every CI, I think this will improve the quality and keep control on new "compiler versions" crazyness autogencode.
4) to allow this I added the infrastructure for a DI pattern, I needed only RetryHelper and IProcessExitHandler injection for now...but with this in place we can improve testability of all components that will become injectables ❤️ ❤️ ❤️ !


src/coverlet.core/DependencyInjection.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System;
```
MarcoRossignoli: @tonerdo this is a key class, setup a "container" with all injectable services(for now only retry helper).
We have some "static" code(methods) so we need a domain wide model `DependencyInjection.Current` provide access to all services.
I used DI framework used by ASP.NET Code inside `Microsoft.Extensions.DependencyInjection` https://github.com/aspnet/Extensions

src/coverlet.core/DependencyInjection.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        public static IServiceProvider Current
+        {
+            get
+            {
+                return _serviceProvider.Value;
+            }
+        }
+
+        public static void Set(IServiceProvider serviceProvider)
```
MarcoRossignoli: this method needed to "plug new services collection" used by testing to pass mocked IRetryHelper aware of process exit and file locking.

src/coverlet.core/Helpers/InstrumentationHelper.cs
```diff
@@ -7,8 +7,9 @@
 using System.Reflection;
```
MarcoRossignoli: this file is the only file changed to use "container" low cost change.

src/coverlet.core/Helpers/RetryHelper.cs
```diff
@@ -1,3 +1,4 @@
+using Coverlet.Core.Abstracts;
```
MarcoRossignoli: Here we simply make RetryHelper an "instance" type that implements IRetryHelper.

src/coverlet.core/coverlet.core.csproj
```diff
@@ -7,15 +7,17 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="2.2.0" />
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="2.2.0" />
+    <PackageReference Include="Microsoft.Extensions.FileSystemGlobbing" Version="2.2.0" />
     <PackageReference Include="Mono.Cecil" Version="0.10.1" />
-    <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
+    <PackageReference Include="Newtonsoft.Json" Version="10.0.1" />
```
MarcoRossignoli: I change this to 10 because Newtosolft has got a bug on 9.0.1 and cannot use TypeConverter in a correct way, this lead to exception during serialization of `CoveragePrepareResult` in testing.
MSbuild does not load this version but version 10.0.1 this is the reason why it works today(unexpected :smile:).

test/coverlet.core.tests/CoverageTests.cs
```diff
@@ -54,5 +60,48 @@ public void TestCoverageWithTestAssembly()
 
             directory.Delete(true);
         }
+
+        [Fact]
+        public void Condition_If()
```
MarcoRossignoli: @tonerdo @petli start from here!
This is the code we need to write to test!

test/coverlet.core.tests/Samples/Instrumentation.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System.Threading.Tasks;
+
+namespace Coverlet.Core.Samples.Tests
+{
+    public class Conditions
```
MarcoRossignoli: Added new file for new "class" of tests.

MarcoRossignoli:
I need to fix post build copy on linux, let me fix

MarcoRossignoli:
Fail on unix...I'm investigating.

MarcoRossignoli:
Found the issue, race between `RestoreOriginalModules` and `RestoreOriginalModule` on msbuild/console/collector are in same `AppDomain` on test nope so they race on `Copy/Delete` on locked files.
I'll fix it.

src/coverlet.core/Abstracts/IProcessExitHandler.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System;
+
+namespace Coverlet.Core.Abstracts
+{
+    public interface IProcessExitHandler
```
MarcoRossignoli: We need this service to plug a fake in child process to avoid race condition on restore module.
In normal condition(no exception) we restore modules during coverage calculation, after  @ViktorHofer update we restore also other instrumented module on exit, on msbuild/console/collector we use a shared dictionary but on test we instrument and run on child process so parent/child race on restore/delete files.
On child test we'll do nop.

MarcoRossignoli:
Bonuns, improved coverage
Pre
```
 +---------------+--------+--------+--------+
  | Module        | Line   | Branch | Method |
  +---------------+--------+--------+--------+
  | coverlet.core | 76,03% | 67,42% | 83,03% |
  +---------------+--------+--------+--------+

  | Total   | 76,03% | 67,42% | 83,03% |
  | Average | 76,03% | 67,42% | 83,03% |
```
Post 
```
  +---------------+--------+--------+--------+
  | Module        | Line   | Branch | Method |
  +---------------+--------+--------+--------+
  | coverlet.core | 80,98% | 71,1%  | 88,82% |
  +---------------+--------+--------+--------+

  | Total   | 80,98% | 71,1%  | 88,82% |
  | Average | 80,98% | 71,1%  | 88,82% |

```
@tonerdo we lost codecov report.

MarcoRossignoli:
@tonerdo  great thank's for WIP!
Ready to review!

MarcoRossignoli:
@tonerdo can you review this?So we can go on with ["core instrumentation"](https://github.com/tonerdo/coverlet/issues?q=is%3Aissue+is%3Aopen+label%3Atenet-coverage) fix and improve test coverage .

src/coverlet.core/Abstracts/IRetryHelper.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System;
+
+namespace Coverlet.Core.Abstracts
+{
+    public interface IRetryHelper
```
ViktorHofer: Are you sure you want to expose such a type in coverlet.core? Have you thought about other alternatives which don't need to expose this, ie internalvisibleto, reflection?

MarcoRossignoli: Good catch...I prefer internalvisibleto

MarcoRossignoli: Btw we should review all visibility of core lib to allow in future to expose core directly as package for advanced scenario, this is a first step.

ViktorHofer: Why would you need such a package?

MarcoRossignoli: Some guys wanted it https://github.com/tonerdo/coverlet/issues/212

PR #422, opened 2019-05-24
Update Documentation

tonerdo:
Updating documentation ahead of `coverlet.collector` NuGet package release

Issue #395, opened 2019-05-03
Integrate coverlet with VSTest (Microsoft Test Platform)


PBoraMSFT:
What would integrating Coverlet with Microsoft Test Platform mean: 
1.	Coverlet based coverage solution is available as a data collector that does the instrumentation of the necessary modules before test execution and appropriate restore after.
2.	**Authoring** – When new test projects are created (dotnet mstest), references to the data collector package are added by default. This reduces adoption friction for customers. 
3.	**Test execution** - Today Coverlet is invoked as a msbuild target and as such only works with dotnet test (and requires a csproj). With the data collector in bullet 1 becoming available, coverage can also be collected when tests are run on built binaries (dotnet vstest)

Edit: Fixed markdown typos. I was referring to bullet 1 in the above description, not the issue number.

MarcoRossignoli:
> With the data collector in #1 becoming available,

Can we see the implementation?
Could this integration help with this issue https://github.com/tonerdo/coverlet/pull/329 ?We're trying to find a way to integrate with VSTest but with a good user experience(`RunSettings` file is not the best maybe)

> Today Coverlet is invoked as a msbuild target and as such only works with dotnet test (and requires a csproj). 

We have also dotnet tool, btw integration with `dotnet vstest` should be great.

We're working on [improving/reorganize "internals"](https://github.com/tonerdo/coverlet/issues/387), this should lead to "stable"(cross finger) interface/contracts usable "outside" as standalone api.
Maybe to integrate with VSTest we need to "refactor"(separate reports and engine) even more and ship `coverlet.core.dll` as nuget package, that is in line with current refactor plan.

vagisha-nidhi:
@MarcoRossignoli Adding implementation details.

Scenario Explanation :

1. Integration of coverlet such that it works with dotnet test.
    a. Command line invocation will be : **dotnet test --collect:"Xplat Code Coverage"**
2. There will be an outproc [datacollector](https://github.com/Microsoft/vstest-docs/blob/master/docs/extensions/datacollector.md)[**CoverletDataCollector**] with friendly name **"Xplat    Code Coverage"** which will have a dependency on **coverlet.core** to get coverage data by means of coverlet.core APIs such as coverage.PrepareModules(), coverage.GetCoverageResult(). 
[...]

PR #423, opened 2019-05-25
Tmp draft PR for Nerdbank.GitVersioning question

MarcoRossignoli:
Tmp PR to understand issue with Nerdbank.GitVersioning for nighly build versioning.

### DIFF
--- a/src/coverlet.collector/version.json
+++ b/src/coverlet.collector/version.json
@@ -1,7 +1,7 @@
 {
-    "$schema": "https://raw.githubusercontent.com/AArnott/Nerdbank.GitVersioning/master/src/NerdBank.GitVersioning/version.schema.json",
-    "version": "1.0.0-build.{height}",
-    "publicReleaseRefSpec": [
-      "^refs/heads/master$"
-    ]
-  }
+  "$schema": "https://raw.githubusercontent.com/AArnott/Nerdbank.GitVersioning/master/src/NerdBank.GitVersioning/version.schema.json",
+  "version": "1.0.0",
+  "publicReleaseRefSpec": [
+    "^refs/heads/master$"
+  ]
+}
--- a/src/coverlet.collector/version.json
+++ b/src/coverlet.collector/version.json
@@ -1,6 +1,6 @@
 {
     "$schema": "https://raw.githubusercontent.com/AArnott/Nerdbank.GitVersioning/master/src/NerdBank.GitVersioning/version.schema.json",
-    "version": "1.0.0",
+    "version": "1.0.0-build.{height}",
     "publicReleaseRefSpec": [
       "^refs/heads/master$"
     ]
--- a/src/coverlet.collector/version.json
+++ b/src/coverlet.collector/version.json
@@ -1,0 +1,7 @@
+{
+    "$schema": "https://raw.githubusercontent.com/AArnott/Nerdbank.GitVersioning/master/src/NerdBank.GitVersioning/version.schema.json",
+    "version": "1.0.0",
+    "publicReleaseRefSpec": [
+      "^refs/heads/master$"
+    ]
+  }
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project>
+  <ItemGroup>
+    <PackageReference Include="Nerdbank.GitVersioning" Version="2.3.38" PrivateAssets="all" />
+  </ItemGroup>
+</Project>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,5 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project>
+  <PropertyGroup>
+    <PackageOutputPath>$(MSBuildThisFileDirectory)build\$(Configuration)\</PackageOutputPath>
+  </PropertyGroup>
+
   <ItemGroup>
     <PackageReference Include="Nerdbank.GitVersioning" Version="2.3.38" PrivateAssets="all" />
   </ItemGroup>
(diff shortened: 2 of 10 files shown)


src/coverlet.collector/version.json
```diff
@@ -0,0 +1,7 @@
+{
+    "$schema": "https://raw.githubusercontent.com/AArnott/Nerdbank.GitVersioning/master/src/NerdBank.GitVersioning/version.schema.json",
+    "version": "1.0.0",
```
onovotny: This is using a fixed version of 1.0.0. If you want NB.GV to work, you should specify either `1.0` or `1.0.0-build.{height}`

onovotny:
By specifying the three part version, NB.GV won't do anything automatically. It increments the patch version and you've specified it. If you specify only the Major.Minor, NB.GV will set the patch version.

src/coverlet.collector/version.json
MarcoRossignoli: Did it, but no luck 
image

src/coverlet.collector/version.json
```diff
@@ -0,0 +1,7 @@
+{
+    "$schema": "https://raw.githubusercontent.com/AArnott/Nerdbank.GitVersioning/master/src/NerdBank.GitVersioning/version.schema.json",
+    "version": "1.0.0-build.{height}",
```
MarcoRossignoli: @onovotny did it but no luck
image

AArnott:
I don't see your nuspec file in this PR. How do you set its version tag? It should be set to `$version$` I believe.

AArnott:
Ya, here is your problem :

https://github.com/tonerdo/coverlet/blob/0271f5eef162788b56fb7707b0475c7288fd8d62/src/coverlet.collector/coverlet.collector.nuspec#L5

MarcoRossignoli:
@AArnott no lock with `$version$`
image
I think `pack` command expect version param as command input...and `$version$` is only a placeholder https://docs.microsoft.com/en-us/nuget/reference/nuspec#replacement-tokens

AArnott:
You're closer. Now per [this doc](https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#packing-using-a-nuspec) you should set a project property like this:

```xml
<PropertyGroup>
  <NuspecProperties>$(NuspecProperties);version=$(NuGetPackageVersion)</NuspecProperties>
</PropertyGroup>
```

MarcoRossignoli:
@AArnott closed but not enough seem that prop is empty 
image
image

AArnott:
Ah, ok. Try adding `InitialTargets="GetBuildVersion"` to your `<Project>` tag at the top of the file.

AArnott:
Oh, that won't work. Hang on... Do this:

```xml
<Target Name="SetNuspecPackageVersion" DependsOnTargets="GetBuildVersion" BeforeTargets="Pack">
  <PropertyGroup>
    <NuspecProperties>$(NuspecProperties);version=$(NuGetPackageVersion)</NuspecProperties>
  </PropertyGroup>
</Target>
```

MarcoRossignoli:
it does no trigger target maybe because `GetBuildVersion` run inside `Pack`
image

AArnott:
How are you building this? 

The build authoring in this repo looks needlessly complex. Why does build.proj manually run all these Exec tasks rather than simply invoking the targets on each project? As it is, the same targets get repeatedly built at each of a few steps. 

When I use `dotnet pack` (which *ought* to be what's used here, probably), all the versions are perfect. If coverlet.collector needs its own nuspec file, this should be specified in the project file itself, not the build.proj as a special parameter.

Anyway, [here's the minimum](https://github.com/AArnott/coverlet/commit/85e5f3445807bea1d78f72b09a79ab6c5267eb23) I had to add to get the package version for coverlet.collector correct. I had to make one change to what I had prescribed earlier.

AArnott:
You might want to take a few more commits from [my nightly branch](https://github.com/AArnott/coverlet/commits/nightly). Which is based on yours.

MarcoRossignoli:
> The build authoring in this repo looks needlessly complex. Why does build.proj manually run all these Exec tasks rather than simply invoking the targets on each project? As it is, the same targets get repeatedly built at each of a few steps.

I agree, we're working on this direction.

> When I use dotnet pack (which ought to be what's used here, probably), all the versions are perfect. If coverlet.collector needs its own nuspec file, this should be specified in the project file itself, not the build.proj as a special parameter.

I agree will do

> You might want to take a few more commits from my nightly branch. Which is based on yours.

Sure thank you very much for your fundamental help and for your great tool!

AArnott:
You're welcome. I'm actually preparing a few more commits to clean up the build authoring. Are you interested, or would you rather take over here?

MarcoRossignoli:
Feel free to commit your idea in you repo and we'll pick up, maybe not immediatly, but sure in next PR also because we will deploy new visual studio agnostic build version https://github.com/tonerdo/coverlet/pull/396 and we'll need to fix some path.
We need to deploy ASAP a nighly build because we will release VSTest official cross coverage collector 
https://github.com/tonerdo/coverlet/issues/395 
https://github.com/tonerdo/coverlet/issues/395#issuecomment-495602245
and we would have a way to provide nighly build to fix possible new bugs, we expect some "load" if MS will advertise this new `Xplat Code Coverage` and we want to be "ready" to provide best experience we can.

One more question on versioning...I would like to have coherent package name but for collector I get different output 
image
But version.json are the same.

MarcoRossignoli:
> Are you interested, 

We are always interested in valuable enhancement FYI here you can take a look on new docs with great explanation of coverlet features https://github.com/tonerdo/coverlet/pull/422 by @tonerdo

MarcoRossignoli:
It works!Close this draft I'll re-open official clean PR thank's @AArnott @onovotny

PR #424, opened 2019-05-25
Move common build steps into template

AArnott:
I've got a lot of simplifications in mind for the build authoring in this repo. But as I see you have a couple other PRs in flight that would result in merge conflicts with what I have in mind, here's my first offering which is highly localized but sets us up to make things simpler going forward.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -11,8 +11,7 @@ jobs:
   pool:
     vmImage: 'windows-2019'
   steps:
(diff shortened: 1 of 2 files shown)


AArnott:
Great. I have a few more branches that build on this one. I'll wait for this PR to complete so the next PRs don't repeat the same diffs.

MarcoRossignoli:
We wait for @tonerdo's double check and approval he is the father of coverlet, I'm only an extra hand!

PR #425, opened 2019-05-26
Setup nightly build

MarcoRossignoli:
This PR setup nightly build, what we have:
* added `Directory.Build.props` with `Nerdbank.GitVersioning` ref for git versioning and setup of `PackageOutputPath`
* cleaned up `coverlet.collector.csproj` to use custom nuspec so we can also cleanup `build.proj`
* added simple ps script for upload packages `eng/nightly.ps1` I know that we could use AZP to do all works(build artifacts and release)...but for now I prefer use what we have on master...and in future step by step we'll refactor build with @AArnott advices, pinned sdk for generic build and test with multi SDK locally testable.
Script accept 2 params `-apiKey xxx and -source https://...` I tested all workflow on https://nugetservercoverlet.azurewebsites.net/ and it works well.
@tonerdo you must setup variable with myget credentials like 
image
Here you can see tests log https://marcorossignoli.visualstudio.com/githubazp/_build/results?buildId=765
YAML ignore errors, because if we don't have new commit on master `donet nuget push` fails for "package already present".
* pinned sdk on `global.json` fist step of @AArnott advice, this update does not conflicts with https://github.com/tonerdo/coverlet/pull/396 because there we use `DOTNET_MULTILEVEL_LOOKUP=0`
BTW after this merge we'll have to override `build.proj` with new one, but I look forward to fix that

@toni after merge you need to create a new "Scheduled" pipeline(`eng/azure-pipelines-nightly.yml`) by hand like
image
Scheduled trigger is not yet supported with YAML https://docs.microsoft.com/en-us/azure/devops/pipelines/build/triggers?view=azure-devops&tabs=yaml#scheduled-triggers

To local test 
```
dotnet msbuild build.proj // like prev
dotnet msbuild build.proj /p:Configuration=Release // like prev
dotnet msbuild build.proj /p:Configuration=Release /p:PublicRelease=false // generate gitversioning nuget package
```
Format will be `coverlet.collector.1.0.0-g2478511111.nupkg` where `2478511111` is the first n part of last commit. Using scripts https://github.com/AArnott/Nerdbank.GitVersioning/tree/master/src/Nerdbank.GitVersioning.Tasks/tools we can translate version to commit.

cc: @vagisha-nidhi 

PS. A mega-huge thank's to @AArnott and @onovotny for the tool and help with setup, very appreciated!


MarcoRossignoli:
@tonerdo after merge of this and new documentation I'll add "Consume nightly build" docs

tonerdo:
@MarcoRossignoli just tried to schedule a build. There's no way to specify `eng/azure-pipelines-nightly.yml` as the config to use

PR #432, opened 2019-05-27
Replace checked in .nuspec file with generated one

### LINKED ISSUE #431: Coverlet.Collector.nuspec puts files in the wrong folder
The checked in Coverlet.Collector.nuspec file is missing this tag:
```xml
    <developmentDependency>true</developmentDependency>
```
which I'm guessing *should* be there since in fact it is present in the .csproj file, but not in the checked in nuspec file.

Also: all the files included in the package are placed under the `build\netcoreapp2.0` folder, which will make the NuGet package only able to install into other .NET Core 2.0+ projects. But since this is a build authoring package, it probably can (and probably should?) apply to all kinds of projects. So the folder used should be `build\netstandard1.0`.

AArnott:
All the redundancy between the .nuspec file and the .csproj file (both express and implied) is now removed. With it, I fix a few bugs:

1. The generated nuspec file includes the developmentDependency tag that was in the project file but not the checked in file.
1. The package now applies to *all* projects regardless of their target framework (well, so long as they're .netstandard1.0 compatible) instead of all projects rejecting this package unless they targeted .NET Core 2.0 or higher.

I also moved the .targets file into a folder structure within the project that resembles where it will appear in the package. This makes it more obvious that it belongs to the package when looking at the source code, and makes it easier to maintain the build/ folder in the package going forward because all files in it are included.

Fixes #431


src/coverlet.collector/coverlet.collector.csproj
```diff
@@ -61,12 +63,11 @@
...
-  <Target Name="SetNuspecPackageVersion" DependsOnTargets="GetBuildVersion" BeforeTargets="GenerateNuspec">
-    <PropertyGroup>
-      <NuspecProperties>$(NuspecProperties);version=$(NuGetPackageVersion);configuration=$(Configuration)
-    </NuspecProperties>
-  </PropertyGroup>
-  </Target>
 
+  <Target Name="PackBuildOutputs">
+    <ItemGroup>
+      <TfmSpecificPackageFile Include="$(TargetDir)\*.dll" PackagePath="build\netstandard1.0\" />
```
MarcoRossignoli: This was my missing piece :)

MarcoRossignoli: @vagisha-nidhi do you have concerns about `netstandard1.0`?

vagisha-nidhi: The nupkg will contain all dlls and targets in `netstandard1.0` folder right? That shouldn't be a problem.

MarcoRossignoli: yup

PR #434, opened 2019-05-27
 Fix `dotnet build` of a clean repo

AArnott:
For a perfectly clean repo (git clean -fdx), `dotnet build` fails because some tests require that other projects have been built using build.proj first.

This change adds the necessary build ordering instructions so that the build tasks are always built before the tests that need them, and the tests are always run against the *latest* source code being tested rather than the last version that happened to be built with build.proj.

### DIFF
--- a/build.proj
+++ b/build.proj
@@ -13,19 +13,7 @@
     <Exec Command="dotnet build &quot;$(MSBuildThisFileDirectory)src\coverlet.collector\coverlet.collector.csproj&quot; -c $(Configuration)" />
   </Target>
 
-  <Target Name="PublishMSBuildTaskProject" AfterTargets="BuildAllProjects">
-    <Exec Command="dotnet publish &quot;$(MSBuildThisFileDirectory)src\coverlet.msbuild.tasks\coverlet.msbuild.tasks.csproj&quot; -c $(Configuration) -o &quot;$(OutputPath)&quot;" />
-  </Target>
-
-  <Target Name="CopyMSBuildScripts" AfterTargets="PublishMSBuildTaskProject">
-    <ItemGroup>
-      <BuildScript Include="$(MSBuildThisFileDirectory)src\coverlet.msbuild.tasks\coverlet.msbuild.props" />
-      <BuildScript Include="$(MSBuildThisFileDirectory)src\coverlet.msbuild.tasks\coverlet.msbuild.targets" />
-    </ItemGroup>
-    <Copy SourceFiles="@(BuildScript)" DestinationFolder="$(OutputPath)" />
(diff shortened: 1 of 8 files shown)


build.proj
```diff
@@ -13,19 +13,7 @@
     <Exec Command="dotnet build &quot;$(MSBuildThisFileDirectory)src\coverlet.collector\coverlet.collector.csproj&quot; -c $(Configuration)" />
   </Target>
 
-  <Target Name="PublishMSBuildTaskProject" AfterTargets="BuildAllProjects">
```
MarcoRossignoli: No sure about that, @tonerdo we never talked about, why you build all in a `build` directory?For possible attach Debug?
With this update under build we'll have only generated packages when we build repo in release.

MarcoRossignoli: @AArnott if I understood well we can pass `CoverletToolsPath` in "repro" issue(for intance) and attach local?

test/Directory.Build.targets
```diff
@@ -0,0 +1,20 @@
+<Project>
```
MarcoRossignoli: @AArnott this is needed if I run only `RunTests` target with a clean repo, right?

build.proj
AArnott: Yes, that would work. But you don't *need* to do that, since the `CoverletToolsPath` property will be set to that local build path automatically when you build your tests.

test/Directory.Build.targets
AArnott: By "this" are you referring to this file? 
It's independent of the `RunTests` target in your `build.proj` file (which I'm working to make less and less necessary).
This file teaches your test projects how to build the msbuild tasks first and then consume the output of that build immediately as part of the test project's own build. It's what sets the `CoverletToolsPath` property so that you don't have to use `/p:CoverletToolsPath` manually on the command line.

build.proj
MarcoRossignoli: > Yes, that would work. But you don't need to do that, since the CoverletToolsPath property will be set to that local build path automatically when you build your tests.

I mean sometimes when it's not possible understand bugs using tests I attach to user repro by coping targes/props inside provided test csproj. Now I can attach and "live debug" without update csproj because using `CoverletToolsPath` I'm redirecting task load to "dev" binaries under my coverlet dev solution, right?

AArnott: Correct

MarcoRossignoli: very good.

test/Directory.Build.targets
MarcoRossignoli: ok thank's.

Issue #403, opened 2019-05-07
Provide private feed with official nightly build


MarcoRossignoli:
We should provide official nightly build to allow users to consume fix/features ASAP and avoid to release official nuget packages, this help for some reason to me:

* New features fix could be "tested" in real project, for instance corefx codebase could pickup nuget package from private feed and do feedback very quick, corefx is used every day by a lot of dev(@ViktorHofer).

* https://github.com/tonerdo/coverlet/pull/402 We are about to become `The cross platform solution for Code Coverage generation for .NET Core projects(in a consistent manner).` likely a lot of user will use coverlet from non win system.
`DataCollector` will be picked up as nuget package, with private feed they can simply update `PackageReference` add private feed and consume fix/features in a fast way(@vagisha-nidhi).

 We could start using:
* myget service for open source https://www.myget.org/opensource
* New github package registry https://github.com/features/package-registry

BTW I don't know if foundation could help with better choice cc: @onovotny
Maybe azure blob storage?

@tonerdo I propose also to move CI to Azure Pipelines, I found command trigger very useful to re-run pipeline https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/github?view=azure-devops&tabs=yaml#comment-triggers at the moment I cannot re-run appveyor CI from PR, I don't know if it's supported.

/cc @tonerdo @petli

MarcoRossignoli:
Per offline discussion...I take this for first part, generate packages and upload to some nuget server, I'll try to setup private service for dev. cc: @tonerdo

tonerdo:
Nightly feed: https://www.myget.org/feed/Packages/coverlet-dev

MarcoRossignoli:
Index path https://www.myget.org/F/coverlet-dev/api/v3/index.json

MarcoRossignoli:
closed in https://github.com/tonerdo/coverlet/pull/425

MarcoRossignoli:
Sorry re-open we'll close when documentation will be up-to-da

MarcoRossignoli:
@AArnott I did some check on nighly build and it works...there is one problem we would have a correct order inside vs on `Latest prerelease` at the moment are ordered in an alphabetic way.
Do we need to use git height?
I would like to have something like `1.0.0-{height}-gxxxxxxxxxx` in this way we should have a correct order and we update every offical release semver of nightly.
What do you think?Can you share your experience?

AArnott:
[...]

PR #439, opened 2019-05-27
[WIP]Fix nightly upload nuget package order

MarcoRossignoli:
Fix nightly upload nuget package order

Now we've order i.e. `coverlet.msbuild.2.6.1-preview-0011-gfbf1f3beb8.nupkg`

cc: @tonerdo


MarcoRossignoli:
@AArnott here https://github.com/tonerdo/coverlet/issues/403#issuecomment-496303004 you said "moving your versioning to a major.minor.height pattern?" but I cannot have this behaviour, in this PR I fixed order issue and I like it we release 3 package with hash and height and after some usage tests I found ok the user experience, clear the order(last master first) and simple way to link with master commit with hash.
Could be also better if we generate something like `coverlet.msbuild.2.6.11-preview-gfbf1f3beb8.nupkg` instead of `coverlet.msbuild.2.6.1-preview-0011-gfbf1f3beb8.nupkg` but `1.0.{height}-preview` schema is not supported, what I'm missing?

AArnott:
> here #403 (comment) you said "moving your versioning to a major.minor.height pattern?" but I cannot have this behaviour,

It's up to you. But I'm a bit confused because later in your above comment it sounds like you're asking for exactly what I had prescribed:

> Could be also better if we generate something like coverlet.msbuild.2.6.11-preview-gfbf1f3beb8.nupkg instead of coverlet.msbuild.2.6.1-preview-0011-gfbf1f3beb8.nupkg

Yes. I'm happy to help you do that.

> but 1.0.{height}-preview schema is not supported, what I'm missing?

That's not the right syntax. Instead, use `1.0-preview`. The `{height}` macro is only used when inside a prerelease tag. Otherwise, the height is implied as the next integer in the version that you haven't specified. So if you specify major.minor, then height goes into PATCH, and if you specify major.minor.patch then height goes into REVISION (4th integer).

AArnott:
Now regarding the `-preview` tag and the hash that you're struggling with... I'm writing up a doc to explain it. I've had to explain it a few times so I figured a doc would save me time down the road.

MarcoRossignoli:
Close and reopen I did a mess with rebase 😞

MarcoRossignoli:
BTW @AArnott seem that works as expected now
image

I'll open a new PR with where I'll publish directly with CI, we'll go on on that thread!

MarcoRossignoli:
> Now regarding the -preview tag and the hash that you're struggling with... I'm writing up a doc to explain it. I've had to explain it a few times so I figured a doc would save me time down the road.

@AArnott point me you doc when done!

AArnott:
Will do. I hope to have it done by EOD.

MarcoRossignoli:
Oh ready for my breakfast :smile:

AArnott:
OK, here is [the first draft](https://github.com/AArnott/Nerdbank.GitVersioning/blob/publicReleaseDoc/doc/public_vs_stable.md). I'd be interested in your feedback and questions before I merge this into master.

MarcoRossignoli:
Doc are very clear...open a draft if you want and I'll review also with some simple advice like newlines

AArnott:
Done: https://github.com/AArnott/Nerdbank.GitVersioning/pull/345

PR #441, opened 2019-05-29
Fix PackageOutputPath in local builds

AArnott:
Define Configuration when undefined before using it. Otherwise packages build to just build\ instead of build\debug.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -2,6 +2,7 @@
 <Project>
   <PropertyGroup>
     <RepoRoot>$(MSBuildThisFileDirectory)</RepoRoot>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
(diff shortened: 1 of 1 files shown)


Directory.Build.props
```diff
@@ -2,6 +2,7 @@
 <Project>
   <PropertyGroup>
     <RepoRoot>$(MSBuildThisFileDirectory)</RepoRoot>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
```
tonerdo: Nit:
```suggestion
    <Configuration Condition="'$(Configuration)' == ''">Debug</Configuration>
```

AArnott: We _can_ do that, but coming from the MSBuild team myself (several years ago) we learned it was easier to read the condition expression when the double and single quotes were separated by a space. So I'd _rather_ not make this change.

PR #442, opened 2019-05-29
Index sources into PDBs

AArnott:
Use SourceLink so that debugging into sources is easy.
Also rely on thet .NET SDK to set the RepositoryUrl property. This allows the generated nuspec file to contain the actual commit ID that produced the package.


AArnott:
The build failure appears to be unrelated and is due to #444

AArnott:
Never mind. It's not unrelated. The build failure is due to the the new snupkg symbols package being somehow empty.

AArnott:
> how to "step into" code in case of issue using one of our way to instrument:

I do it by adding a check for an unlikely environment variable, then calling Debugger.Launch()

MarcoRossignoli:
> I do it by adding a check for an unlikely environment variable, then calling Debugger.Launch()

So you release with `Debugger.Launch()` under if?I understood well?

MarcoRossignoli:
ping @tonerdo

AArnott:
ping

MarcoRossignoli:
ping @tonerdo

PR #443, opened 2019-05-29
Build to bin\<config>\Packages instead of build\ 

AArnott:
Includes #441 and goes a step further:

The "bin" folder name is more common for .NET based projects such as this one. And it allows the .gitignore file to ignore the "bin" folder while allowing the "build" folder to appear under project directories with a build folder in source code that contains package payload.


Directory.Build.props
```diff
@@ -2,10 +2,11 @@
 <Project>
   <PropertyGroup>
     <RepoRoot>$(MSBuildThisFileDirectory)</RepoRoot>
-    <PackageOutputPath>$(MSBuildThisFileDirectory)build\$(Configuration)\</PackageOutputPath>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <PackageOutputPath>$(MSBuildThisFileDirectory)bin\$(Configuration)\Packages\</PackageOutputPath>
```
MarcoRossignoli: I was experimenting something with this and could great have a separation between nightly and official for produced pacakge something like https://github.com/MarcoRossignoli/coverlet/blob/fixversion/Directory.Build.props#L4 (maybe `official` instead of `live`)
and produce artifacts like https://github.com/MarcoRossignoli/coverlet/blob/fixversion/azure-pipelines-artifacts.yml
We can also add same job to `azure-pipelines.yml` on Windows Release config, build is fast
After that we can run a scheduled build in the night https://marcorossignoli.visualstudio.com/githubazp/_build/results?buildId=784
publish with release https://marcorossignoli.visualstudio.com/githubazp/_release?definitionId=2&view=mine&_a=releases(fail are ok because if package already exists nuget throw...no -ignoreIfExists yet)

As long as you're on it can you also fix `version[s].json`?
https://github.com/MarcoRossignoli/coverlet/blob/fixversion/src/coverlet.collector/version.json
https://github.com/MarcoRossignoli/coverlet/blob/fixversion/src/coverlet.console/version.json
https://github.com/MarcoRossignoli/coverlet/blob/fixversion/src/coverlet.msbuild.tasks/version.json

AArnott: I don't think we should build `publicrelease=true` and `publicrelease=false` packages from the same branch regularly. I don't see why we need both. Can you elaborate on that?

What is it about version.json you need fixed?

I don't yet see why we need a nightly build since the CI build produces everything we need. A release pipeline can pick up the latest CI build and deploy it to the nuget feed on a schedule if that's what you want.

MarcoRossignoli: > I don't yet see why we need a nightly build since the CI build produces everything we need. A release pipeline can pick up the latest CI build and deploy it to the nuget feed on a schedule if that's what you want.

I agree it's unuseful.

> I don't think we should build publicrelease=true and publicrelease=false packages from the same branch regularly. I don't see why we need both. Can you elaborate on that?

I thought about a way to deploy on nightly and on official nuget throught release pipeline, one scheduled(nightly) and the other manual(that we use only on public release).

> What is it about version.json you need fixed?

At the moment version are all fixed MAJOR.MINOR.PATCH(i.e. https://github.com/tonerdo/coverlet/blob/master/src/coverlet.collector/version.json)  I would like to start to let to nbgv generate PATCH.

AArnott: > I thought about a way to deploy on nightly and on official nuget throught release pipeline, one scheduled(nightly) and the other manual(that we use only on public release).

It sounds like your goal is to include the git commit ID in the package versions of your myget feed pushes. Is that right, and if so, why? The package carries metadata so we can recover the commit ID that built it if we need to.

> I would like to start to let to nbgv generate PATCH.

Great. I'll send a separate PR for that.

MarcoRossignoli: > It sounds like your goal is to include the git commit ID in the package versions of your myget feed pushes. Is that right, and if so, why? The package carries metadata so we can recover the commit ID that built it if we need to.

Only to be clear that package is a "pre-release" and it's recognized as is on tools, for instance visual studio.

MarcoRossignoli: I mean if we check cproj it's clear that we're referencing a pre-release package.

AArnott: OK, I think that's misusing the "public release" flag as a _prerelease_ flag. If you want to signify a prerelease, we can add `-beta` or something to the version.json files. We'll just have to remove that flag leading up to the quarterly release.

Alternatively if you feel like the product is already stable, then my recommendation is leave the prerelease tag off and push to your myget feed what looks like stable packages, since they are in fact stable.

MarcoRossignoli: Product is already stable, go for push to myget stable package, I've got a different way to think about versioning.
To me release is "released to public and official advertised as to use" and "-beta or -commit or -prerelease-commit" is "maybe unstable". 
After a release to me it's "unstable" until next release that will be done when product is know as "stable".
But I want to start to think like you because I don't like the idea to strip out `-beta` only to release, I need to start to evaluate if product is stable or not, not if we have something stable but not released yet
```
The unstable nature of a product might be in functional resilience, or that its API isn't finalized, or lack of adequate testing. 
Any and all of these are based on the assessment of the software engineers responsible for the project.
```
I'll embrace this paradigm.

AArnott: OK, it sounds like we shouldn't have a `-beta` tag in the version.json files because they're actually "stable".
Now as to the question of pushing packages to myget with a -gCommitId prerelease tag, that's up to you. If you want to proceed with that, we'll just remove all `publicReleaseRefSpec` entries from the version.json file so that they always are seen as non-public. Then when it's time to push to nuget.org, you can queue a build with a pipeline queue-time variable of `PublicRelease=true` set to get the versions you need to push to nuget.org.

MarcoRossignoli: Maybe if we use 1 pipeline build for 1 pipeline release it's better remove all `publicReleaseRefSpec` so we will sure that we're not releasing result of in progress PR(could be hard understand it due to automatic PATCH).
We'll enqueue `PublicRelease=true` and we check artifacts before "push" on ui.

AArnott: PR builds would not show up as a public release because they always build within context of a branch named `pulls/*` which doesn't match any of the branches under `publicReleaseRefSpec`.
So if we simply remove all references to `publicrelease` everywhere, then all builds will be non-public unless you explicitly set it at queue time on the day you intend to ship.

MarcoRossignoli: Mmm...it's not so clear

 > PR builds would not show up as a public release because they always build within context of a branch named pulls/* which doesn't match any of the branches under publicReleaseRefSpec.

If so how we can publish artifacts from CI? Before you said "A release pipeline can pick up the latest CI build and deploy it to the nuget feed on a schedule if that's what you want.", do we need to create a new build pipeline after merge to master?One for PR and one for after merge? And link release to this build?What do you mean for "lastest CI build"?

Can you explain what NBGV do in case of empty `publicReleaseRefSpec`?

https://github.com/AArnott/Nerdbank.GitVersioning/wiki/PublicReleaseRefSpec do you mean that you recognize if you're on build server(azp for instance) and implicity apply `PublicRelease=true`?
And if is empty on build server it doesn't find "master"(regular exp) record so it does not apply `PublicRelease=true`?

AArnott: When `publicReleaseRefSpec` is empty or absent, it just means the checked out branch can't match it, and thus publicrelease=true won't be the default --- it will always be false.

> do you mean that you recognize if you're on build server(azp for instance) and implicity apply PublicRelease=true?

No. The default is set to true when you build with HEAD checked out to a ref in the `publicReleaseRefSpec`. This is true locally or in cloud builds. But cloud builds are very intentional about which commit is being built so they don't ever actually check out a *branch*. They always check out a "detached head" by specifying the commit to check out explicitly. But they *do* usually have a branch in mind that they claim to be building, and therein lines the need for NB.GV to detect cloud builds -- we have to rely on environment variables that these cloud builders set to determine which branch or tag is conceptually being built in order to determine what the default value of PublicRelease should be.

> If so how we can publish artifacts from CI?

You should only need one build Pipeline, which builds PR and CI builds. And it will default to building whatever kind of packages you want your release pipeline to pick up and push to myget on a regular and frequent basis. It sounds like that would be publicrelease=false packages. 
Your build pipeline can be configured to take an optional PublicRelease variable at queue time, which lets you fill in `true` when you want to force it to build packages without the git commit in their versions in preparation to publish to nuget.org. After that build completes, you queue a release of that build and push to the nuget.org environment.

MarcoRossignoli: Ok all clear thank's!Sorry to bother you too kind!

> So if we simply remove all references to publicrelease everywhere, then all builds will be non-public unless you explicitly set it at queue time on the day you intend to ship.

Try with this for now!We'll see after some usage with @tonerdo if it fit well our needs!

MarcoRossignoli: > After that build completes, you queue a release of that build and push to the nuget.org environment.

Ah and if I'm not mistaken after release we need to run CI build with `false` to avoid to release package without commit on myget.

AArnott: > if I'm not mistaken after release we need to run CI build with false to avoid to release package without commit on myget.

I wouldn't worry about that one bit. If myget happens to get pushed a package that was already on nuget.org, no big deal, right?

MarcoRossignoli: Yes not a great issue.

MarcoRossignoli:
I agree with this refactor @AArnott also here when you've time could you resolve? cc: @tonerdo

PR #454, opened 2019-06-06
Add "Consume nightly build" docs

MarcoRossignoli:
Added docs on nightly build usage

closes https://github.com/tonerdo/coverlet/issues/403

cc: @tonerdo


MarcoRossignoli:
@tonerdo we could merge this we had been a first user!

Issue #463, opened 2019-06-11
Wrong percentages within ReportGenerator


boldt:
# Setup

* nUnit (3.12.0)
* coverlet.msbuild (2.6.2)
* ReportGenerator (4.1.8)

## My toolchain

```bash
dotnet clean
dotnet build
dotnet test --no-build /p:CollectCoverage=true /p:CoverletOutputFormat=\"cobertura\"
packages\ReportGenerator.4.1.10\tools\net47\ReportGenerator.exe "-reporttypes:Html;HtmlChart" "-reports:Test\coverage.cobertura.xml" "-targetdir:report/" "-filters:-AddInPlayground*"
```

# Issue

In ReportGenerator I see 966600% instead of 96%:

![grafik](https://user-images.githubusercontent.com/750793/59027791-3d656480-885a-11e9-96e3-663553bbd94e.png)

This is how it looks in the `coverage.cobertura.xml`:

```xml
<method name="VariableChanged" signature="()" line-rate="1" branch-rate="0,9666">
```
## Related bug report

I already created a bug report at ReportGenerator: https://github.com/danielpalme/ReportGenerator/issues/252

They say:

> The problem is, that I would expect `0.9666`. The `,` makes the difference.
> 
> There are several tools that generate _Cobertura_ files. _coverlet_ is the only one that uses `,`.

# Solution

Report the rate in dot-format instead of comma-format.

MarcoRossignoli:
Thank's for reporting

boldt:
Thanks for the very quick fixing. Please let me know, when it's released officially.

MarcoRossignoli:
@boldt after merge you can use nightly build with fix, at the moment I don't know if we plan to release official so quick cc: @tonerdo

PR #464, opened 2019-06-11
[Bug fix]Fix culture for cobertura xml report

MarcoRossignoli:
Fixes https://github.com/tonerdo/coverlet/issues/463

@boldt comes from Germany where culture uses comma as decimal separator(like Italy).
This PR fix conversion using `InvariantCulture` that is defined as 
`The invariant culture is culture-insensitive; it is associated with the English language but not with any country/region. ` 
doc https://docs.microsoft.com/en-us/dotnet/api/system.globalization.cultureinfo.invariantculture?view=netcore-2.2

Found this useful guide on formats...english languages uses dot. https://docs.oracle.com/cd/E19455-01/806-0169/overview-9/index.html

cc: @tonerdo

### DIFF
--- a/src/coverlet.core/Reporters/CoberturaReporter.cs
+++ b/src/coverlet.core/Reporters/CoberturaReporter.cs
@@ -1,9 +1,8 @@
 using System;
-using System.Collections.Generic;
+using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Text;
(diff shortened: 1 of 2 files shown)


MarcoRossignoli:
@tonerdo remember to change the merge strategy to "squash merge" to make easy possible reverts.

Issue #466, opened 2019-06-12
Test run fails on netcoreapp1.0 targeted tests


AArnott:
When running tests on linux (in Azure Pipelines at least) with code coverage turned on...

```
dotnet test src --no-build -c Release --filter "TestCategory!=FailsInCloudTest" -v n -l trx /p:CollectCoverage=true /p:CoverletOutputFormat=cobertura
```

The [build log](https://dev.azure.com/andrewarnott/OSS/_build/results?buildId=1384&view=logs&j=2d2b3007-3c5c-5840-9bb0-2b1ea49925f3) shows this failure when the tests run on .NET Core 1.0:

```
[xUnit.net 00:00:01.64]       System.IO.FileNotFoundException : Could not load file or assembly 'netstandard, Version=2.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'. The system cannot find the file specified.
[xUnit.net 00:00:01.64]       Stack Trace:
[xUnit.net 00:00:01.64]            at Coverlet.Core.Instrumentation.Tracker.Nerdbank.Streams_42ec0704-f64b-48c0-a903-a8b9be93bc98.RecordHit(Int32 )
[xUnit.net 00:00:01.64]         D:\a\1\s\src\Nerdbank.Streams\Sequence`1.cs(43,0): at Nerdbank.Streams.Sequence`1..ctor()
[xUnit.net 00:00:01.64]         D:\a\1\s\src\Nerdbank.Streams.Tests\BufferWriterStreamTests.cs(25,0): at BufferWriterStreamTests..ctor(ITestOutputHelper logger)
```

MarcoRossignoli:
Yes I did some check and the issue seem mainly 2 packages 
`Mono.Cecil 0.10.1`
`Microsoft.Extensions.FileSystemGlobbing 2.0.1`

that are supported only on `netstandard2.0` and collector tfm is `netcoreapp2.0`

@vagisha-nidhi is there are reason to not build collector lib for `netstandard2.0`?
If so we could update guide to report that minimal support is `netstandard2.0` runtimes.

vagisha-nidhi:
@MarcoRossignoli There seems no particular reason that I can recall. Can you try making it `netstandard2.0` and see if everything works?

vagisha-nidhi:
@MarcoRossignoli Although the coverlet integration for now will be working for only .NET Core projects, that may be the reason of not making this `netstandard` targeted.

[...]

PR #467, opened 2019-06-12
Update requirements on docs

MarcoRossignoli:
Contributes https://github.com/tonerdo/coverlet/issues/466

cc: @tonerdo @AArnott @vagisha-nidhi

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,9 @@ After the above command is run, a `coverage.json` file containing the results wi
 
 See [documentation](Documentation/MSBuildIntegration.md) for advanced usage.
 
+#### Requirements
+Requires a runtime that support _.NET Standard 2.0 and above_
+
 ### Global Tool
 
 The `coverlet` tool is invoked by specifying the path to the assembly that contains the unit tests. You also need to specify the test runner and the arguments to pass to the test runner using the `--target` and `--targetargs` options respectively. The invocation of the test runner with the supplied arguments **must not** involve a recompilation of the unit test assembly or no coverage result will be generated.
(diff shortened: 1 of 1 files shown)


AArnott:
The docs look fine, but I don't think this should close #466 because a doc on this README is neither discoverable nor alleviates the fault-*looking* error when the test run fails. 

IMO we need to update docs where users look. Personally, the docs I was basing my work on were here: https://aka.ms/vstest-collect. They don't even mention coverlet, but apparently coverlet is an implementation detail of how code coverage is collected. Can/should we get those docs updated to indicate the restriction?

Also, can we produce a better error than the one mentioned in #466 to educate the user as to what they should do?

This limitation required that I break up my otherwise monolithic `dotnet test` command into individual `dotnet test -f TFM1` commands (6 of them, I think) just so I could avoid adding the code coverage switch to the `netcoreapp1.x` targeted test target frameworks. I'd rather keep just the one command in my Azure Pipeline and have the non-supporting targets emit a warning to the effect of "coverage won't be collected on .NET Core 1.x" but otherwise let the tests run their course.

MarcoRossignoli:
> The docs look fine, but I don't think this should close #466 because a doc on this README is neither discoverable nor alleviates the fault-looking error when the test run fails.

I understood that issue was to only update our side with the info.

> IMO we need to update docs where users look. Personally, the docs I was basing my work on were here: https://aka.ms/vstest-collect. They don't even mention coverlet, but apparently coverlet is an implementation detail of how code coverage is collected. Can/should we get those docs updated to indicate the restriction?

Should be keep updated by coverlet dev members? cc: @vagisha-nidhi 

> Also, can we produce a better error than the one mentioned in #466 to educate the user as to what they should do?
This limitation required that I break up my otherwise monolithic dotnet test command into individual dotnet test -f TFM1 commands (6 of them, I think) just so I could avoid adding the code coverage switch to the netcoreapp1.x targeted test target frameworks. I'd rather keep just the one command in my Azure Pipeline and have the non-supporting targets emit a warning to the effect of "coverage won't be collected on .NET Core 1.x" but otherwise let the tests run their course.

I agree

PR #470, opened 2019-06-14
VSTest collector: target netstandard

iainteractive-iain:
Instead of netcoreapp.

This allows our VSTest collector to be used within Visual Studio (e.g. v16.1.3). Without this, said collector only seems to work from the VSTest console application.

### DIFF
--- a/src/coverlet.collector/coverlet.collector.csproj
+++ b/src/coverlet.collector/coverlet.collector.csproj
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk" >
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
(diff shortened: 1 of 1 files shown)


iainteractive-iain:
Once this is merged (assuming it is merged), it would be great if a new NuGet package could be published.

That will allow PrestoCoverage to then use coverlet's VSTest collector.

MarcoRossignoli:
We talked about it https://github.com/tonerdo/coverlet/issues/466#issuecomment-501158913 cc: @vagisha-nidhi  thoughts?

MarcoRossignoli:
@iainteractive-iain thank's for PR but for the moment we won't change tfm for collector, as @vagisha-nidhi said at the moment collector are well supported by VSTest only for .NET Core project, in this way we can also use full core features.

Issue #81, opened 2018-05-03
[F#]Auto-properties are not implicitly covered


SteveGilham:
Writing unit tests for auto-properties is a futile exercise; they are essentially noise (as is the coverage or otherwise of such properties).

Auto-properties in code exercised by coverlet are marked as uncovered by default be they C#
```
    public class Class1
    {
        string Property { get; set; }
    }
```
or F#
```
type Class1 () =
  member val Property= String.Empty with get, set
```

lucasbflopes:
Any updates on this feature?

MarcoRossignoli:
@tonerdo do you agree on adding `-skipautoprops`(opencover name)?
I don't have a strong opinion on "exclude always" vs "switch"

MarcoRossignoli:
At the moment we can achieve this using [`ExcludeByAttribute`](https://github.com/tonerdo/coverlet/blob/master/Documentation/MSBuildIntegration.md#excluding-from-coverage) thank's to https://github.com/tonerdo/coverlet/pull/477#issuecomment-512515625

MarcoRossignoli:
Closing because we can use `/p:ExcludeByAttribute=\"CompilerGeneratedAttribute\"`

MarcoRossignoli:
Ugh...sorry to you for my ignorance...didn't know, I'm not familiar with F# compiler.
I'm thinking to add a F# lib on tests to handle the cases.
Thank's Steve.

PR #477, opened 2019-06-26
Fix property attribute detection

amweiss:
I realized my initial work on #233 and #232 overlooked custom attributes on properties and checking them against each of the generated methods. This resolves that with a few updated tests and may achieve the desired result for #81 if you add `CompilerGenerated` to the list of excluded attributes.

I didn't create a new issue for this per the contribution guidelines since it was technically related to an existing one, but let me know if you want one there.

### DIFF
--- a/src/coverlet.core/Instrumentation/Instrumenter.cs
+++ b/src/coverlet.core/Instrumentation/Instrumenter.cs
@@ -273,10 +273,20 @@ namespace Coverlet.Core.Instrumentation
             foreach (var method in methods)
             {
                 MethodDefinition actualMethod = method;
+                IEnumerable<CustomAttribute> customAttributes = method.CustomAttributes;
                 if (InstrumentationHelper.IsLocalMethod(method.Name))
                     actualMethod = methods.FirstOrDefault(m => m.Name == method.Name.Split('>')[0].Substring(1)) ?? method;
 
-                if (!actualMethod.CustomAttributes.Any(IsExcludeAttribute))
+                if (actualMethod.IsGetter || actualMethod.IsSetter)
+                {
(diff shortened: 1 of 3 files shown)


src/coverlet.core/Instrumentation/Instrumenter.cs
```diff
@@ -273,10 +273,18 @@ private void InstrumentType(TypeDefinition type)
...
             {
                 MethodDefinition actualMethod = method;
+                IEnumerable<CustomAttribute> customAttributes = method.CustomAttributes;
                 if (InstrumentationHelper.IsLocalMethod(method.Name))
                     actualMethod = methods.FirstOrDefault(m => m.Name == method.Name.Split('>')[0].Substring(1)) ?? method;
 
-                if (!actualMethod.CustomAttributes.Any(IsExcludeAttribute))
+                if (actualMethod.IsGetter || actualMethod.IsSetter)
+                {
+                    var prop = type.Properties.FirstOrDefault(p => p.GetMethod.FullName.Equals(actualMethod.FullName));
```
MarcoRossignoli: nit: please use full name `PropertyDefinition` and not `var` when assignment expression does not decribe in a clear way a type, it's unreadable.

amweiss: sure it's an easy one to change, fixed.

ardalis:
Question: Can someone confirm whether this resolves #81 and if so, which version of coverlet this bug fix is in/will be in? Thanks!

amweiss:
@ardalis I can confirm if run the nightly build and exclude `CompilerGenerated` attribute, it does exclude them from my reports (dropping about 80 lines off my coverable lines). I think that achieves the goal of #81 .

image

PR #479, opened 2019-06-27
Changed to calculate based on the average coverage of the modules

dlplenin:
Fixes https://github.com/tonerdo/coverlet/issues/346

Minor changes to calculate Average (console's table) based on average from all projects. Im not sure if it is ok only to modify CalculateLineCoverage(Modules modules) functions.
Also a change in CoverageResult.cs just to keep same logic, because its switch-case (ThresholdStatistic.Average)  was working well (average based on average from all projects).

### DIFF
--- a/src/coverlet.console/Program.cs
+++ b/src/coverlet.console/Program.cs
@@ -172,9 +172,9 @@ namespace Coverlet.Console
                 var totalBranchPercent = summary.CalculateBranchCoverage(result.Modules).Percent;
                 var totalMethodPercent = summary.CalculateMethodCoverage(result.Modules).Percent;
 
-                var averageLinePercent = summary.CalculateLineCoverage(result.Modules).AveragePercent;
-                var averageBranchPercent = summary.CalculateBranchCoverage(result.Modules).AveragePercent;
-                var averageMethodPercent = summary.CalculateMethodCoverage(result.Modules).AveragePercent;
+                var averageLinePercent = summary.CalculateLineCoverage(result.Modules).AverageModulePercent;
+                var averageBranchPercent = summary.CalculateBranchCoverage(result.Modules).AverageModulePercent;
+                var averageMethodPercent = summary.CalculateMethodCoverage(result.Modules).AverageModulePercent;
 
                 foreach (var _module in result.Modules)
                 {
--- a/src/coverlet.console/Program.cs
+++ b/src/coverlet.console/Program.cs
@@ -172,13 +172,17 @@ namespace Coverlet.Console
                 var totalBranchPercent = summary.CalculateBranchCoverage(result.Modules).Percent;
                 var totalMethodPercent = summary.CalculateMethodCoverage(result.Modules).Percent;
 
+                var averageLinePercent = summary.CalculateLineCoverage(result.Modules).AveragePercent;
+                var averageBranchPercent = summary.CalculateBranchCoverage(result.Modules).AveragePercent;
+                var averageMethodPercent = summary.CalculateMethodCoverage(result.Modules).AveragePercent;
+
                 foreach (var _module in result.Modules)
                 {
                     var linePercent = summary.CalculateLineCoverage(_module.Value).Percent;
                     var branchPercent = summary.CalculateBranchCoverage(_module.Value).Percent;
                     var methodPercent = summary.CalculateMethodCoverage(_module.Value).Percent;
 
-                    coverageTable.AddRow(Path.GetFileNameWithoutExtension(_module.Key), $"{linePercent}%", $"{branchPercent}%", $"{methodPercent}%");
+                    coverageTable.AddRow(Path.GetFileNameWithoutExtension(_module.Key), $"{linePercent.ToString("0.00")}%", $"{branchPercent.ToString("0.00")}%", $"{methodPercent.ToString("0.00")}%");
                 }
 
                 logger.LogInformation(coverageTable.ToStringAlternative());
@@ -187,8 +191,8 @@ namespace Coverlet.Console
                 coverageTable.Rows.Clear();
 
                 coverageTable.AddColumn(new[] { "", "Line", "Branch", "Method" });
-                coverageTable.AddRow("Total", $"{totalLinePercent}%", $"{totalBranchPercent}%", $"{totalMethodPercent}%");
-                coverageTable.AddRow("Average", $"{totalLinePercent / numModules}%", $"{totalBranchPercent / numModules}%", $"{totalMethodPercent / numModules}%");
+                coverageTable.AddRow("Total", $"{totalLinePercent.ToString("0.00")}%", $"{totalBranchPercent.ToString("0.00")}%", $"{totalMethodPercent.ToString("0.00")}%");
+                coverageTable.AddRow("Average", $"{averageLinePercent.ToString("0.00")}%", $"{averageBranchPercent.ToString("0.00")}%", $"{averageMethodPercent.ToString("0.00")}%");
 
                 logger.LogInformation(coverageTable.ToStringAlternative());
 
--- a/src/coverlet.console/Program.cs
+++ b/src/coverlet.console/Program.cs
@@ -168,13 +168,17 @@ namespace Coverlet.Console
                 var summary = new CoverageSummary();
                 int numModules = result.Modules.Count;
 
-                var totalLinePercent = summary.CalculateLineCoverage(result.Modules).Percent;
-                var totalBranchPercent = summary.CalculateBranchCoverage(result.Modules).Percent;
-                var totalMethodPercent = summary.CalculateMethodCoverage(result.Modules).Percent;
+                var linePercentCalculation = summary.CalculateLineCoverage(result.Modules);
+                var branchPercentCalculation = summary.CalculateBranchCoverage(result.Modules);
+                var methodPercentCalculation = summary.CalculateMethodCoverage(result.Modules);
 
-                var averageLinePercent = summary.CalculateLineCoverage(result.Modules).AverageModulePercent;
-                var averageBranchPercent = summary.CalculateBranchCoverage(result.Modules).AverageModulePercent;
-                var averageMethodPercent = summary.CalculateMethodCoverage(result.Modules).AverageModulePercent;
+                var totalLinePercent = linePercentCalculation.Percent;
+                var totalBranchPercent = branchPercentCalculation.Percent;
+                var totalMethodPercent = methodPercentCalculation.Percent;
+
+                var averageLinePercent = linePercentCalculation.AverageModulePercent;
+                var averageBranchPercent = branchPercentCalculation.AverageModulePercent;
+                var averageMethodPercent = methodPercentCalculation.AverageModulePercent;
 
                 foreach (var _module in result.Modules)
                 {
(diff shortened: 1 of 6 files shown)


MarcoRossignoli:
Diego can you add some tests here https://github.com/tonerdo/coverlet/blob/master/test/coverlet.core.tests/CoverageSummaryTests.cs?

sasivishnu:
Just a suggestion... could you use `modules.Count` instead of `modules.Count()`? since property already available to give the value. Linq `Count()` method do use property `Count` when type is of `ICollection` I believe.

src/coverlet.core/CoverageResult.cs
```diff
@@ -196,33 +196,26 @@ public ThresholdTypeFlags GetThresholdTypesBelowThreshold(CoverageSummary summar
...
-                        {
-                            line += summary.CalculateLineCoverage(module.Value).Percent;
-                            branch += summary.CalculateBranchCoverage(module.Value).Percent;
-                            method += summary.CalculateMethodCoverage(module.Value).Percent;
-                        }
-
                         if ((thresholdTypes & ThresholdTypeFlags.Line) != ThresholdTypeFlags.None)
                         {
-                            if ((line / numModules) < threshold)
+                            if (line < threshold)
```
sasivishnu: Since we aren't not rounding down average value of `line`, there is chance that we will output coverage failed for average threshold but display value indicate it shouldn't have failed.

Ex: threshold value is "85.25" and lets say `line` value is "85.2463", then if condition will evaluate to true and we output coverage failed for average threshold. But while displaying, we print "85.25" because we do `ToString("0.00")` which will do normal rounding. 

Same applies to below calculation for branch & method.

src/coverlet.core/CoverageSummary.cs
```diff
@@ -53,12 +53,15 @@ public CoverageDetails CalculateLineCoverage(Documents documents)
         public CoverageDetails CalculateLineCoverage(Modules modules)
         {
             var details = new CoverageDetails();
+            var accumPercent = 0.0;
```
sasivishnu: Just to make it more obvious, can you do `var accumPercent = 0.0D`

src/coverlet.console/Program.cs
```diff
@@ -172,13 +172,17 @@ static int Main(string[] args)
                 var totalBranchPercent = summary.CalculateBranchCoverage(result.Modules).Percent;
                 var totalMethodPercent = summary.CalculateMethodCoverage(result.Modules).Percent;
 
+                var averageLinePercent = summary.CalculateLineCoverage(result.Modules).AverageModulePercent;
+                var averageBranchPercent = summary.CalculateBranchCoverage(result.Modules).AverageModulePercent;
+                var averageMethodPercent = summary.CalculateMethodCoverage(result.Modules).AverageModulePercent;
```
sasivishnu: We can reuse objects from line# 171, 172 and 173, then get total percent and average percent from those `CoverageDetails` object for each type line, branch and module respectively. This way we avoid doing recalculation and get some perf benefit. What you think?

sasivishnu:
@MarcoRossignoli I did review since I worked on this section last time. Let me know if you want to remove my comments. I am doing review first time in Github so I am not sure whether I did it right or messed up something.

src/coverlet.core/CoverageDetails.cs
```diff
@@ -6,6 +6,7 @@ public class CoverageDetails
     {
         public double Covered { get; internal set; }
         public int Total { get; internal set; }
+        public double AverageModulePercent { get; internal set; }
```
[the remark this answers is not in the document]
dlplenin: I agree with u, at first try I wrote with this approach, just more verbose than `ToString("0.00")`, but as u said we have different outputs if only use string format:
> Ex: threshold value is "85.25" and lets say line value is "85.2463", then if condition will evaluate to true and we output coverage failed for average threshold. But while displaying, we print "85.25" because we do ToString("0.00") which will do normal rounding.

I realized about it, but I needed another oint of view, tks! I'll work on it

src/coverlet.console/Program.cs
dlplenin: yep, it need a refactoring

MarcoRossignoli:
@sasivishnu absolutely NOT!
This project is OSS every idea review discussion on code are welcome!
Feel free to comment everything!
We won't remove any comments if constructive!

dlplenin:
@MarcoRossignoli Tests added.
@sasivishnu I refactored according your suggestions, also I change the aligment for numbers (just visual thing), let me knot if it is ok or if it add complexity in vane.

test/coverlet.core.tests/CoverageSummaryTests.cs
```diff
@@ -75,8 +75,52 @@ private void SetupData()
             _modules.Add("module", documents);
         }
 
+        private void SetupAditionalData()
```
MarcoRossignoli: nit: rename `SetupDataForAverageCalculation` and call inside constructor `CoverageSummaryTests` one time.

MarcoRossignoli:
> also I change the aligment for numbers (just visual thing), let me knot if it is ok or if it add complexity in vane.

Were there issues here?I don't think we should touch that code in this PR, better keep the focus on issue topic.
BTW can you attach screenshot with before/after to understand the difference?

src/coverlet.core/CoverageDetails.cs
```diff
@@ -6,6 +6,13 @@ public class CoverageDetails
     {
         public double Covered { get; internal set; }
         public int Total { get; internal set; }
+        private double averageModulePercent;
+        public double AverageModulePercent
+        {
+            get { return Math.Round(averageModulePercent, 2); ; }
```
MarcoRossignoli: nit: double token `;`

sasivishnu:
> 
> 
> > also I change the aligment for numbers (just visual thing), let me knot if it is ok or if it add complexity in vane.
> 
> Were there issues here?I don't think we should touch that code in this PR, better keep the focus on issue topic.
> BTW can you attach screenshot with before/after to understand the difference?

Yep, agree with @MarcoRossignoli here. No need to change any alignment as part of this PR. Please open new issue to discuss regarding your alignment proposal.

src/coverlet.core/CoverageDetails.cs
sasivishnu: Also, can you do `return Math.Floor(averageModulePercent * 100)/100;`

To keep consistency with the way Coverlet handles rounding. We round down using floor. You can refer `Percent `property below this line.

test/coverlet.core.tests/CoverageSummaryTests.cs
dlplenin: Rename is done,.... but if SetupDataForAverageCalculation is called into the constructor, previous tests will fail because initial data will have two modules.
I want to have two scnearios to test, one with one module(TestCalculateBranchCoverage_OneModule), and other with multiple modules (TestCalculateBranchCoverage_Modules).... what do u think?

src/coverlet.core/CoverageDetails.cs
dlplenin: Done

dlplenin:
> > also I change the aligment for numbers (just visual thing), let me knot if it is ok or if it add complexity in vane.
> 
> Were there issues here?I don't think we should touch that code in this PR, better keep the focus on issue topic.
> BTW can you attach screenshot with before/after to understand the difference?

There isnt a issue at all, it is just a visual aligment for number. It is not really important so I reverted the code. 
Just: "My client asked me why number are aligned to left in CI report, ...I like that my clients value the use of tests"

sasivishnu:
LGTM

test/coverlet.core.tests/CoverageSummaryTests.cs
MarcoRossignoli: I prefer custom private fields with custom setup for tests for instance

```cs
private Modules _averageCalculationSingleModule;
private Modules _averageCalculationMultiModule;
```
and load in constructor.

dlplenin: Changed... pls take a look at line 123 (LMK if its ok) => I want to reuse data from SetupDataSingleModule

src/coverlet.core/CoverageDetails.cs
```diff
@@ -6,6 +6,13 @@ public class CoverageDetails
     {
         public double Covered { get; internal set; }
         public int Total { get; internal set; }
+        private double averageModulePercent;
```
MarcoRossignoli: nit: move this field on top of class and change in `_averageModulePercent`
```cs
public class CoverageDetails
    {
        private double _averageModulePercent;
        public double Covered { get; internal set; }
        public int Total { get; internal set; }
        public double AverageModulePercent
        {
            get { return Math.Floor(_averageModulePercent * 100) / 100; }
            internal set { _averageModulePercent = value; }
        }

        public double Percent => Total == 0 ? 100D : Math.Floor((Covered / Total) * 10000) / 100;
    }
```

src/coverlet.core/CoverageResult.cs
MarcoRossignoli: Vishnu is this comment out of date?

sasivishnu: Yes.. Since we are rounding it in property `AverageModulePercent ` itself, we are good now.

src/coverlet.core/CoverageResult.cs
```diff
@@ -196,33 +196,26 @@ public ThresholdTypeFlags GetThresholdTypesBelowThreshold(CoverageSummary summar
                     break;
                 case ThresholdStatistic.Average:
                     {
-                        double line = 0;
-                        double branch = 0;
-                        double method = 0;
+                        var line = summary.CalculateLineCoverage(Modules).AverageModulePercent;
```
MarcoRossignoli: nit: `double`

test/coverlet.core.tests/CoverageSummaryTests.cs
```diff
@@ -71,58 +73,148 @@ private void SetupData()
...
+            Lines lines = new Lines
+            {
+                { 1, 1 },//covered
+                { 2, 0 },//not covered
+                { 3, 0 }//not covered
+            };
+
+            Branches branches = new Branches
+            {
+                new BranchInfo { Line = 1, Hits = 1, Offset = 1, Path = 0, Ordinal = 1 },//covered
```
MarcoRossignoli: nit: add space before comments ` // covered` everywhere

test/coverlet.core.tests/CoverageSummaryTests.cs
MarcoRossignoli: I see it's ok, some nits and LGTM

src/coverlet.core/CoverageDetails.cs
dlplenin: Done

src/coverlet.core/CoverageResult.cs
dlplenin: Done... just one thing: lines 173 to 175 & 225 to 227 are using `var`, LMK if var o double is better, I can do a quick push if needed

MarcoRossignoli: yes please push fast update to keep style.

MarcoRossignoli: The rule I follow is "if the type is clear from assignment it's ok var but if it's not clear better make clean".

dlplenin: Done!

MarcoRossignoli:
@tonerdo I merged this stale bug fix.
Feel free to revert if not in agreement.

PR #487, opened 2019-07-12
Add a major roll-forward policy

ViktorHofer:
Unfortunately there is no way today to specify a roll-forward policy when invoking a repo tool with `dotnet tool run ...`. This is disappointing. Meanwhile this should help people trying to run coverlet.console without having a 2.x SDK installed.

@MarcoRossignoli I would highly appreciate if we could upload a new version to nuget with this fix. If that's not possible I would be willing to use the private feed but usually we try to avoid that in corefx.

### DIFF
--- a/src/coverlet.console/runtimeconfig.template.json
+++ b/src/coverlet.console/runtimeconfig.template.json
@@ -1,0 +1,3 @@
+{
+  "rollForwardOnNoCandidateFx": 2
+}


MarcoRossignoli:
Viktor confirm my doubt...you need this to run coverage on machine with only 3.0 preview installed?ie CI etc...?

ViktorHofer:
CI is one scenario but that also affects every dev machine where only a runtime >2.x is installed. ReportGenerator made the same change some months ago because of this

ViktorHofer:
Ready to merge? :)

MarcoRossignoli:
> Ready to merge? :)

Our merge policy expects 2 approval cc: @tonerdo

ViktorHofer:
@tonerdo any chance to upload a new package to nuget?

PR #495, opened 2019-07-22
Fix setter only property instrumentation

ViktorHofer:
Properties can have a setter without a getter which causes the current type instrumentation code to crash. Fixing this by using the SetMethod if GetMethod is null.

cc @MarcoRossignoli @tonerdo


MarcoRossignoli:
We should add a test to cover this, I'll open a new issue

PR #498, opened 2019-07-24
Revert "Using File.Exists for HitsFile creation/update validation"

MarcoRossignoli:
Reverts tonerdo/coverlet#415

This change broke System.Private.CoreLib code coverage measurement as System.IO.File is not available in it.


MarcoRossignoli:
cc: @ViktorHofer @tonerdo

Issue #476, opened 2019-06-26
Coverlet assumes coverage of 3rd party files with embedded ppdb


onovotny:
Hi,

Looking here: https://dev.azure.com/dotnet/ReactiveUI/_build/results?buildId=18689&view=codecoverage-tab

I see that it's trying to measure coverage of System.Reactive. Rx has embedded ppdb's in it. I'm assuming Coverlet assumes to measure coverage if it finds a pdb, but many 3rd party libraries have embedded ones. 

Is there a way to better detect this so people don't have to have large amounts of config by defualt?

MarcoRossignoli:
AFAIK we instrument also with embedded pdb https://github.com/tonerdo/coverlet/blob/f64337e74ef605b3ad3d9a7d304e02274c4383a6/src/coverlet.core/Helpers/InstrumentationHelper.cs#L82
When coverlet is not able to instrument you should see a warning in console https://github.com/tonerdo/coverlet/blob/09f91f2afa623dbd011acac8a10f6023d9513812/src/coverlet.core/Instrumentation/Instrumenter.cs#L61
Can you enale verbose logs to understand which is o isn't instrumented? https://github.com/tonerdo/coverlet/blob/master/Documentation/Troubleshooting.md

onovotny:
My point is that I don't want it to instrument/include coverage for those third parties.

MarcoRossignoli:
Mmm how do coverlet know that are "third parties"?

MarcoRossignoli:
Maybe there is a way to filter out 0% asm cc: @danielpalme

onovotny:
Is there a way for it to detect if the source files are present for it (based on the pdb data)? That could be the most fool-proof method.

MarcoRossignoli:
We could take a look...maybe could be some interaction with source link switch cc: @tonerdo

onovotny:
Where in the code base does that processing happen?

MarcoRossignoli:
Instrument check https://github.com/tonerdo/coverlet/blob/5a139b200f7bf51c21660b93f41906f4571c971b/src/coverlet.core/Coverage.cs#L96
Source link data gathering https://github.com/tonerdo/coverlet/blob/09f91f2afa623dbd011acac8a10f6023d9513812/src/coverlet.core/Instrumentation/Instrumenter.cs#L108
[...]

PR #510, opened 2019-08-05
Skip instrumentation of module with embedded ppbd without local sources

MarcoRossignoli:
closes https://github.com/tonerdo/coverlet/issues/476

cc: @tonerdo

### DIFF
--- a/src/coverlet.core/Helpers/InstrumentationHelper.cs
+++ b/src/coverlet.core/Helpers/InstrumentationHelper.cs
@@ -110,15 +110,22 @@ namespace Coverlet.Core.Helpers
                                 Document document = metadataReader.GetDocument(docHandle);
                                 string docName = metadataReader.GetString(document.Name);
 
-                                // We stop at first document we need only to verify if module has got local source
-                                return File.Exists(docName);
+                                // We verify all docs and return false if not all are present in local
+                                // We could have false negative if doc is not a source
+                                // Btw check for all possible extension could be weak approach
+                                if (!File.Exists(docName))
+                                {
+                                    return false;
+                                }
                             }
                         }
                     }
                 }
             }
 
-            return false;
+            // If we don't have EmbeddedPortablePdb entry return true, for instance empty dll
+            // We should call this method only on embedded pdb module
+            return true;
         }
 
         public static void BackupOriginalModule(string module, string identifier)
--- a/src/coverlet.core/Helpers/InstrumentationHelper.cs
+++ b/src/coverlet.core/Helpers/InstrumentationHelper.cs
@@ -5,6 +5,7 @@ using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using System.Reflection.Metadata;
 using System.Reflection.PortableExecutable;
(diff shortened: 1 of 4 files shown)


src/coverlet.core/Helpers/InstrumentationHelper.cs
```diff
@@ -90,6 +93,34 @@ public static bool HasPdb(string module)
...
+                        using (MetadataReaderProvider embeddedMetadataProvider = peReader.ReadEmbeddedPortablePdbDebugDirectoryData(entry))
+                        {
+                            MetadataReader metadataReader = embeddedMetadataProvider.GetMetadataReader();
+                            foreach (DocumentHandle docHandle in metadataReader.Documents)
+                            {
+                                Document document = metadataReader.GetDocument(docHandle);
+                                string docName = metadataReader.GetString(document.Name);
+
+                                // We stop at first document we need only to verify if module has got local source
+                                return File.Exists(docName);
```
MarcoRossignoli: @onovotny does it make sense?
You've more familiarity than me with ppdb, I mean is it possible some type of document different than local source code that could return false negative?

MarcoRossignoli:
@ViktorHofer does it work with corelib?

src/coverlet.core/Helpers/InstrumentationHelper.cs
onovotny: I'm not sure, sorry.

MarcoRossignoli: Thanks for quick response, I'll refine with only cs sources.

onovotny:
Will this also skip a module with an external PDB that doesn't have local source? That's another possible scenario that should probably be handled the same way.

MarcoRossignoli:
> Will this also skip a module with an external PDB that doesn't have local source? That's another possible scenario that should probably be handled the same way.

No this PR handle only embedded PDB...do you've some case with that issue?

onovotny:
If a nuget package includes the pdb's in it, default behavior will put them in the output directory, which would lead to the same issue.

MarcoRossignoli:
> If a nuget package includes the pdb's in it, default behavior will put them in the output directory, which would lead to the same issue.

Can you point me to a package that behave like that, so I can write a test.

onovotny:
I'd have to make one as I don't know one off the top of my head.

ViktorHofer:
> @ViktorHofer does it work with corelib?

I didn't test this but it should work fine as only changes to the ModuleTrackerTemplate are dangerous for CoreLib coverage.

MarcoRossignoli:
@tonerdo can you take a look?
With this update I think that we could also remove fixed xunit filters...maybe in future.

tonerdo:
@MarcoRossignoli we can go ahead an remove xunit filters before I make another release this Friday

MarcoRossignoli:
> @MarcoRossignoli we can go ahead an remove xunit filters before I make another release this Friday

@tonerdo I though about a bit more...I don't know if could lead to same problem in future, I mean what if for some reason xunit will remove embedded pdb?
Maybe we could keep filter as fallback. What do you think?

PR #518, opened 2019-08-14
Add ReportGenerator helper

MarcoRossignoli:

Added report generator helper, in this way we can generate during debugging and testing full report to better understand if results are ok.

cc: @tonerdo @petli 

@danielpalme I saw that report generator supports plugin...I would like "run generator" from my code base(adding core package) can you point me to some code sample?
I don't wan to create a custom report, only run, for instance `ReportGenerator.Run(reportfile,reportdirectory)`
BTW code below works well, but we could avoid global tool install.

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -12,6 +12,7 @@
     <PackageReference Update="Moq" Version="4.10.1" />
     <!-- Do not upgrade this version or we won't support old SDK -->
     <PackageReference Update="Newtonsoft.Json" Version="9.0.1" />
+    <PackageReference Update="ReportGenerator.Core" Version="4.2.15" />
     <!--
         Do not change System.Reflection.Metadata version since we need to support VSTest DataCollectors. Goto https://www.nuget.org/packages/System.Reflection.Metadata to check versions.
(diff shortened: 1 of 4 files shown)


danielpalme:
@MarcoRossignoli 
The following approach should work for you:
1. Install package [ReportGenerator.Core](https://www.nuget.org/packages/ReportGenerator.core)
2. Use one of the following code snippets:
```csharp
Palmmedia.ReportGenerator.Core.Program.Main($"-reports:\"{reportFile}\" \"-targetdir:{dir.FullName}\" {(string.IsNullOrEmpty(sourceFileFilter) ? string.Empty : $" -filefilters:{sourceFileFilter}")}");
```

```csharp
new Palmmedia.ReportGenerator.Core.Generator().GenerateReport(new Palmmedia.ReportGenerator.Core.ReportConfiguration(
    new[] { reportFile },
    dir.FullName,
    new string[0],
    null,
    new string[0],
    new string[0],
    new string[0],
    new string[0],
    string.IsNullOrEmpty(sourceFileFilter) ? new string[0] : new[] { sourceFileFilter },
    null,
    null));
```

MarcoRossignoli:
Great thank's!

PR #462, opened 2019-06-10
Add `[xunit*]*` to default excluded modules filter if not specified

MarcoRossignoli:
closes https://github.com/tonerdo/coverlet/issues/461

cc: @tonerdo @vagisha-nidhi @onovotny

PR #519, opened 2019-08-14
Revert "Add `[xunit*]*` to default excluded modules filter if not specified"

MarcoRossignoli:
Reverts tonerdo/coverlet#462

With https://github.com/tonerdo/coverlet/pull/510 now we exclude from instrumentation module with embedded pdb without local source.
For this reason we could think to remove `xunit` filter from code as @tonerdo told here https://github.com/tonerdo/coverlet/pull/510#issuecomment-521247960
We could keep filter as fallback in case that xunit for some reason will move embedded pdb to local folder(@onovotny talked about it here https://github.com/tonerdo/coverlet/pull/510#issuecomment-518243308)

Personally I would keep filter ***for the moment***, but yes could be unuseful/dead code.

Thoughts? @tonerdo @vagisha-nidhi @onovotny @ViktorHofer 
Maybe @bradwilson could also be useful here to confirm or not something.

Ah one more thing, if someone know the name of nuget package that download pdb on output directory please let me know...so we can write code/test to skip also that case.

### DIFF
--- a/src/coverlet.collector/DataCollection/CoverletSettingsParser.cs
+++ b/src/coverlet.collector/DataCollection/CoverletSettingsParser.cs
@@ -131,13 +131,6 @@ namespace Coverlet.Collector.DataCollection
                 }
             }
 
-            // if we've only one element mean that we only added CoverletConstants.DefaultExcludeFilter
-            // so add default exclusions
-            if (excludeFilters.Count == 1)
-            {
-                excludeFilters.Add("[xunit*]*");
-            }
-
             return excludeFilters.ToArray();
         }
 
--- a/src/coverlet.console/Program.cs
+++ b/src/coverlet.console/Program.cs
@@ -58,12 +58,6 @@ namespace Coverlet.Console
                     logger.Level = verbosity.ParsedValue;
                 }
 
-                // We add default exclusion filter if no specified
-                if (excludeFilters.Values.Count == 0)
-                {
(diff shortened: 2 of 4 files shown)


tonerdo:
@MarcoRossignoli what do you mean by

> We could keep filter as fallback in case that xunit for some reason will move embedded pdb to local 

This change seems legit to me as it's unlikely for the xunit source files to be on a user's machine except they work on xunit itself

MarcoRossignoli:
> This change seems legit to me as it's unlikely for the xunit source files to be on a user's machine except they work on xunit itself

@tonerdo I mean that for the moment we check for source only for embedded pdb, so if for some reason(unlikely I think) move pdb to local folder we won't check for source but we'll try to instrument

tonerdo:
Oh I see. Well we should update the code to check for local sources for both embedded and standalone pdb

MarcoRossignoli:
> Oh I see. Well we should update the code to check for local sources for both embedded and standalone pdb

I agree as I've asked to Oren I would like to write a test on it, but  I need a "sample" nuget package that behave like that (before build one tailored made)...@tonerdo do you know one?

MarcoRossignoli:
Great thanks Oren! I'm on vacancy until next week, I'll give it a try asap!

MarcoRossignoli:
Thanks @bradwilson for India.
@tonerdo we can merge for now...I'll update with external pdb's on new PR

PR #526, opened 2019-08-20
Add switch test

MarcoRossignoli:
Improve "real" instrumentation testing

image

cc: @tonerdo @petli


test/coverlet.core.tests/CoverageTests.cs
```diff
@@ -106,5 +106,35 @@ public void Condition_If()
...
+                        return Task.CompletedTask;
+                    }, pathSerialize);
+                    return 0;
+                }, path).Dispose();
+
+                CoverageResult result = TestInstrumentationHelper.GetCoverageResult(path);
+
+                result.Document("Instrumentation.SelectionStatements.cs")
+                      .AssertLinesCovered(BuildConfiguration.Release, (24, 1), (26, 0), (28, 0))
+                      .AssertBranchesCovered(BuildConfiguration.Release, (24, 1, 1))
```
petli: This looks like it make the test very dependent on the optimisations in the compiler, so they could start breaking on a different SDK version. Since these tests are about matching specific lines it seems better to just use Debug runs and not include any expected outcome for Release runs.

test/coverlet.core.tests/CoverageTests.cs
```diff
@@ -106,5 +106,35 @@ public void Condition_If()
                 File.Delete(path);
             }
         }
+
+        [Fact]
```
petli: To get more mileage out of these tests they can perhaps be theories (at least when only using Debug builds), e.g.:

    [Theory]
    [InlineData(24, 1)]
    [InlineData(26, 2)]
    public void SelectionStatements_Switch(int expectedReturnLineHit, int expectedBranchTaken)

test/coverlet.core.tests/CoverageTests.cs
MarcoRossignoli: > This looks like it make the test very dependent on the optimisations in the compiler,

Yep it's true...talked also with Andrew https://github.com/tonerdo/coverlet/pull/396#issuecomment-495944424 and we had same issue in past https://github.com/tonerdo/coverlet/issues/389
Yes we could test only for debug config...but why if some bug happen on release and not on debug config for some reason related to "instrumentation" bug?If in future we put some conditional code related to some new roslyn behaviour?Maybe I'm too concern about something that won't never happen 😅 , but could be a good "doorbell", and maybe we could add the possibility to run test only for specific SDK like a custom xunit attribute `[RunOnSdks("2.2.301","..")]`, to be very rigorous.
Another thing that I'm not sure is if also "debug" config will be stable...I mean also that config could depend on `different SDK version`, less likely but I'm not 100% sure.
**BTW I don't have a super strong opinion and we could test only debug config**...I tried to be complete as possible...
Thank's for review Peter!!!
@tonerdo thoughts?

test/coverlet.core.tests/CoverageTests.cs
MarcoRossignoli: If I'm not mistaken with [`Theory`] we run test n times where n is number of `[InlineData]` so in this case could be heavy because every test clone dll and instrument in external process, I mean more load for same result, because inline expectedReturnLineHit,expectedBranchTaken are "strictly related" each other(source code will be the same).
This is the reason I created new custom "asserts" helper.

petli: Yes, each InlineData becomes one unit test from the point of running the test. I guess it's difficult to use a test fixture where the test DLL is cloned and instrumented once and then used by all tests and still having it as an external process...

MarcoRossignoli: >  I guess it's difficult to use a test fixture

Yes...because we could also "instrument" with different parameters(not exposed at the moment https://github.com/tonerdo/coverlet/blob/master/test/coverlet.core.tests/InstrumenterHelper.cs#L168)

MarcoRossignoli:
@tonerdo @petli what do you think about this PR?
I'd like at least another approval, open question is remove or not release asserts.

MarcoRossignoli:
I'll merge this, low risk I need test for next PRs.

Issue #529, opened 2019-08-28
Skip instrumentation for external pdb with no local sources


MarcoRossignoli:
Follow up https://github.com/tonerdo/coverlet/pull/519#issuecomment-525854352

cc @tonerdo

MarcoRossignoli:
@onovotny seem that external pdb isn't an issue because pdb path is mapped to build drive, so also if it's in local won't be instrumented.

![image](https://user-images.githubusercontent.com/7894084/64184155-55558000-ce6b-11e9-84be-33858750fef5.png)

PR #531, opened 2019-09-04
Inject InstrumentationHelper

MarcoRossignoli:
I'm working on https://github.com/tonerdo/coverlet/issues/529 and I need to inject an IFilesystem service to instrumentation helper so simulate a missing source files.
To do that I need to move InstrumentationHelper class from static to instance, in this way we can also test a core part o coverlet, the instrumentation helper.

contributes also to https://github.com/tonerdo/coverlet/issues/387

cc: @tonerdo @petli


src/coverlet.core/DependencyInjection.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace Coverlet.Core
 {
-    internal static class DependencyInjection
+    public static class DependencyInjection
```
MarcoRossignoli: The only thing I need to do is to expose container to be used by console/task/collector without expose internal implementation.

src/coverlet.core/Helpers/InstrumentationHelper.cs
```diff
@@ -10,21 +10,22 @@
...
     {
-        private static readonly ConcurrentDictionary<string, string> _backupList = new ConcurrentDictionary<string, string>();
+        private readonly ConcurrentDictionary<string, string> _backupList = new ConcurrentDictionary<string, string>();
+        private IRetryHelper _retryHelper;
 
-        static InstrumentationHelper()
+        public InstrumentationHelper(IProcessExitHandler processExitHandler, IRetryHelper retryHelper)
         {
-            DependencyInjection.Current.GetService<IProcessExitHandler>().Add((s, e) => RestoreOriginalModules());
+            processExitHandler.Add((s, e) => RestoreOriginalModules());
```
MarcoRossignoli: @ViktorHofer you did this change, now a single instance of `InstrumentationHelper`  will be returned by ["static" container](https://github.com/tonerdo/coverlet/pull/531/files#diff-7358e9f0a54c6af47045bebb661bb44eR33)(singleton) so the delegate will keep object alive for "restore".
Is it ok for you or do you see some issue with it?

PR #532, opened 2019-09-05
fix hits file share

teh13th:
Exclude the possibility of modifying or deleting a hits file while reading.

Probably, will fix 'Unable to read beyond end of stream' problem. 
We encountered this annoying problem in our project

### DIFF
--- a/src/coverlet.core/Coverage.cs
+++ b/src/coverlet.core/Coverage.cs
@@ -247,7 +247,7 @@ namespace Coverlet.Core
                     }
                 }
 
-                using (var fs = new FileStream(result.HitsFilePath, FileMode.Open))
+                using (var fs = new FileStream(result.HitsFilePath, FileMode.Open, FileAccess.Read, FileShare.Read))
                 using (var br = new BinaryReader(fs))
                 {
                     int hitCandidatesCount = br.ReadInt32();


MarcoRossignoli:
Mmm no sure to understand your issue, what happen?
Hit file should be read/write only in a sequential way...the workflow is:
1) instrument il
2) test run and we collect the hits
3) we write the hits to file
4) we read file

3/4 are sequential...
Are you running coverage for different project that share same dll?

teh13th:
> Are you running coverage for different project that share same dll?

No.

Our build system deletes hits file after it's been opened for reading by Coverlet. Because of this, Coverlet crashes on one of `br.ReadInt32()` calls with '`Unable to read beyond end of stream`'.
I know that problem is in our build system (and we'll fix it), 
but I thought it would be more reliable for Coverlet if it would open hits file with an explicit `FileShare.Read`. 
In this case, no one can modify/delete hits file until Coverlet reads it.

MarcoRossignoli:
> Unable to read beyond end of stream'.

We know that we've an historical issue on this resolved by collectors(vs test infrastructure kill process due to fixed timeout during hit file writing)

> In this case, no one can modify/delete hits file until Coverlet reads it.

Understood your idea, btw I'm think it's not correct allow overlap, this could hide other issues/bug and that "exlusive" access is a guarantee that our workflow works as expected.

> I know that problem is in our build system (and we'll fix it),

I propose to you to fork coverlet and run your patched version until fix...let me know if you need some help for publish packet and use local file system nuget source.
What do you think?Does it help?

teh13th:
OK, thanks for answer.

PR #524, opened 2019-08-20
Fix exclude by files

MarcoRossignoli:
Fixes https://github.com/tonerdo/coverlet/issues/122

At the moment the issue is that files to exclude are searched on current directory and this is the problem.
I moved source file exclusion evaluation inside instrumentation for every instrumented dll/source.

NB. At the moment we allow rule with root drive defined...but this doesn't work well with globbing pattern matching(d:\file and c:\file will be matched like same file), btw this implementation allow the use, in future we could think about forbid. I don't think make a lot of sense use full path on rules.

I run this test
win

```
dotnet test /p:CollectCoverage=true /p:Include=\"[coverlet.core*]*\" /bl  /p:ExcludeByFile=\"**\Instrumenter.cs,**\ModuleTrackerTemplate.cs\,**\InstrumenterResult.cs"
```
with this result 
Capture

ubuntu
```
 dotnet test /p:CollectCoverage=true /p:Include=\"[coverlet.core*]*\" /bl  /p:ExcludeByFile=\"**\\Instrumenter.cs,**\\ModuleTrackerTemplate.cs\,**\\InstrumenterResult.cs\"
```
image

Report without `Instrumenter.cs, InstrumenterResult.cs`  [coverage.json.txt](https://github.com/tonerdo/coverlet/files/3520521/coverage.json.txt)
Normal report without filter `Instrumenter.cs, InstrumenterResult.cs` are present [coverage.json.txt](https://github.com/tonerdo/coverlet/files/3520544/coverage.json.txt)

cc: @tonerdo @petli 
@ido-namely you did first implementation, if you've some time let me know what do you think.
@SlowLogicBoy @orkenstein  @hack2root if you find some time could clone this branch and do tests, not obligated, you can use your local clone following this guide https://github.com/tonerdo/coverlet/blob/master/Documentation/Troubleshooting.md#use-local-build

PR #558, opened 2019-09-19
Fix source file check for non portable pdb

MarcoRossignoli:
Thank's to @adrianhara https://github.com/tonerdo/coverlet/pull/524#issuecomment-532807597 I found an issue on new local source file check during instrumentation.
If pdf is not a portable pdb `metadataReaderProvider.GetMetadataReader()` throws exception `Invalid COR20 header signature.` because cecil can read only portable pdb.

65247903-87b0df80-daf1-11e9-9187-74e256b94283

For now in case of exception for this reason we'll skip check so we'll instument module.
Return false would block user and in case of issue we'll can skip bad module with exclusion filter, so I think it's a good solution.

cc: @tonerdo @petli 

Thank's a lot Adrian for the repro!

NB: minor change, I fixed redundant verbosity log.
New verbose output

image

### DIFF
--- a/src/coverlet.core/Coverage.cs
+++ b/src/coverlet.core/Coverage.cs
@@ -81,6 +81,7 @@ namespace Coverlet.Core
 
             Array.ForEach(_excludeFilters ?? Array.Empty<string>(), filter => _logger.LogVerbose($"Excluded module filter '{filter}'"));
             Array.ForEach(_includeFilters ?? Array.Empty<string>(), filter => _logger.LogVerbose($"Included module filter '{filter}'"));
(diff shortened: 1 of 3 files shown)


MarcoRossignoli:
@adrianhara could be great if tomorrow you give it a try(with nightly) with non portable pdb and tell us if it works as expected!

adrianhara:
@MarcoRossignoli thanks for the quick fix. I'll give it a try and let you know.

adrianhara:
I tried the latest nightly and it works fine. Thanks @MarcoRossignoli ! 👍

MarcoRossignoli:
@adrianhara we released new package https://www.nuget.org/packages/coverlet.msbuild/2.7.0 you can switch to production!

Issue #555, opened 2019-09-18
We should strong name coverlet.collector.dll


MarcoRossignoli:
Now that we've in place injection it's easy that some internal types that lives in core have to be used(injected) in collectors code.
To do that we have 2 way, expose type or apply internals visible to attribute.
I prefer second way but to be clean/safe we should add signed assemblies.

Is there a reason to not sign coverlet.collector.dll?

cc: @vagisha-nidhi for vs test side

tonerdo:
I believe coverlet.msbuild.tasks.dll is already strongly named

MarcoRossignoli:
Uh you're right...

vagisha-nidhi:
@MarcoRossignoli You can sign the dll. For the inproc collector to be picked up, we'll have to make a small change similar to https://github.com/microsoft/vstest/pull/2020/files on the vstest side.

PR #566, opened 2019-09-23
Sign collector and update visibility

MarcoRossignoli:
contributes to https://github.com/tonerdo/coverlet/issues/555

I've signed collector and updated all visibility of object.
This is a first step to expose coverage.core.dll as a lib, we'll expose only public pertinent object through interfaces/factories to be free to update internal implementation.

AFTER MERGE WE MUST UPDATE VSTEST RUNTIME AND DO SOME TESTS https://github.com/microsoft/vstest/pull/2020/files

cc: @tonerdo @vagisha-nidhi @petli


MarcoRossignoli:
coverlet.collector.snk keys
```
C:\git\coverlet\src\coverlet.collector (signcollector -> origin)
λ sn -tp token.snk

Microsoft (R) .NET Framework Strong Name Utility  Version 4.0.30319.0
Copyright (c) Microsoft Corporation.  All rights reserved.

Public key (hash algorithm: sha1):
00240000048000009400000006020000002400005253413100040000010001003d23b9ef372215
da7c81af920b919db5799fd021a1ca10b2e9e0ddac71237a29f8f6361a805a747457e561a7d616
417f1870cda099486df25d580a4e11a0738293342881566254d7840e42f42fb9bfd8e8dca354df
7dc68db14b2d0cd79bb2bf7afdbd62bd948d81b534cba7a326cf6ee840a1aee5dba0a1c660b308
13ca99e5

Public key token is 4beb1ad7f05a1c71

C:\git\coverlet\src\coverlet.collector (signcollector -> origin)
λ sn -t token.snk

Microsoft (R) .NET Framework Strong Name Utility  Version 4.0.30319.0
Copyright (c) Microsoft Corporation.  All rights reserved.

Public key token is 4beb1ad7f05a1c71
```

src/coverlet.core/Reporters/IReporter.cs
```diff
@@ -1,14 +1,14 @@
 namespace Coverlet.Core.Reporters
 {
-    public interface IReporter
+    internal interface IReporter
```
petli: I'm not sure in what situations coverlet would be used as a lib, but could one be that the user want to provide their own IReporter?  Or would they just get CoverageResult directly from coverlet and do whatever they need to do with that without going through the reporting interface?

MarcoRossignoli: In past we had a discussion on it https://github.com/tonerdo/coverlet/issues/212 some guys wants integrate the feature as native.

MarcoRossignoli: Do you agree?What do you think?

petli: Ok, so for now hide everything, then in a separate issue start designing externally viable APIs?  Sounds good.

MarcoRossignoli: Yep

MarcoRossignoli:
> I have reservations about exposing coverlet.core publicly

@tonerdo I don't have a strong opinion about it...I don't know if this scenario makes sense.
If you remember we talked about it some time ago https://github.com/tonerdo/coverlet/issues/212 and also in there I wasn't so "happy" about.
We can decide to wait for more community pressure. 
BTW do you agree in hiding all implementation to be more clean?
I like the idea that if someone add reference directly to coverlet libs doesn't see anything.
@petli thought?

MarcoRossignoli:
@vagisha-nidhi merged FYI

PR #569, opened 2019-09-26
Improve exception message

### LINKED ISSUE #459: Improve exception message for sdk that doesn't support collectors
Moved from https://github.com/tonerdo/coverlet/issues/456#issuecomment-500184354

 @vagisha-nidhi's comment
```
Data collector 'XPlat code coverage' message: System.MissingMethodException: Method not found: '!!0 Microsoft.VisualStudio.TestPlatform.ObjectModel.DataCollection.SessionStartEventArgs.GetPropertyValue(System.String)'.
   at Coverlet.Collector.DataCollection.CoverletCoverageCollector.GetTestModules(SessionStartEventArgs sessionStartEventArgs)
   at Coverlet.Collector.DataCollection.CoverletCoverageCollector.OnSessionStart(Object sender, SessionStartEventArgs sessionStartEventArgs).
```
Should this message be replaced with something more informative (like using the correct sdk version)?

cc: @cltshivash @vagisha-nidhi @tonerdo

daveMueller:
closes #459


daveMueller:
image

src/coverlet.collector/DataCollection/CoverletCoverageCollector.cs
```diff
@@ -165,15 +165,28 @@ private void OnSessionEnd(object sender, SessionEndEventArgs e)
...
+            }
+            catch (MissingMethodException ex)
+            {
+                throw new Exception("Method not found, make sure to use .NET core SDK Version >= 2.2.300", ex);
             }
-
             return testModules;
         }
+
+        private static IEnumerable<string> GetTestSourcesValues(SessionStartEventArgs sessionStartEventArgs)
```
MarcoRossignoli: why this helper?If I'm not mistaken is used only in the above `GetModules` method

src/coverlet.collector/DataCollection/CoverletCoverageCollector.cs
```diff
@@ -165,15 +165,28 @@ private void OnSessionEnd(object sender, SessionEndEventArgs e)
...
+                if (_eqtTrace.IsInfoEnabled)
+                {
+                    _eqtTrace.Info("{0}: TestModules: '{1}'",
+                        CoverletConstants.DataCollectorName,
+                        string.Join(",", testModules ?? Enumerable.Empty<string>()));
+                }
+            }
+            catch (MissingMethodException ex)
+            {
+                throw new Exception("Method not found, make sure to use .NET core SDK Version >= 2.2.300", ex);
```
MarcoRossignoli: we should retrow same exception `MissingMethodException`

src/coverlet.collector/DataCollection/CoverletCoverageCollector.cs
daveMueller: I need this helper to catch the exception. In this case the ```MissingMethodException``` is thrown during the JITting phase and not during normal code execution. Without the helper it's impossible to catch it.

src/coverlet.collector/DataCollection/CoverletCoverageCollector.cs
daveMueller: I've now put the original exception message in the inner exception part and used our piece in the real exception message without concatinating it. I tried your formatting but it looked wired because the whole namespace down to the missing function was printed.

image

With the way it is now it looks like this:

image

src/coverlet.collector/DataCollection/CoverletCoverageCollector.cs
MarcoRossignoli: ah ok change the name to `GetTestSourcesValuesWrapper` and add a comment above like "We need to wrap  GetPropertyValue to catch possible MissingMethodException on unsupported runtime".

We want to be sure that nobody remove wrapper...and as is seem unuseful.

src/coverlet.collector/DataCollection/CoverletCoverageCollector.cs
MarcoRossignoli: Can you zoom on second image?I want to be sure to see the "real missing method" `GetPropertyValue`.

daveMueller: Yes this screenshot should be better:

image

src/coverlet.collector/DataCollection/CoverletCoverageCollector.cs
daveMueller: Done

src/coverlet.collector/DataCollection/CoverletCoverageCollector.cs
```diff
@@ -165,15 +165,33 @@ private void OnSessionEnd(object sender, SessionEndEventArgs e)
         /// <returns>Test modules list</returns>
         private IEnumerable<string> GetTestModules(SessionStartEventArgs sessionStartEventArgs)
         {
-            var testModules = sessionStartEventArgs.GetPropertyValue<IEnumerable<string>>(CoverletConstants.TestSourcesPropertyName);
-            if (_eqtTrace.IsInfoEnabled)
+            IEnumerable<string> testModules;
```
MarcoRossignoli: why not
```cs
 private IEnumerable<string> GetTestModules(SessionStartEventArgs sessionStartEventArgs)
        {
            try
            {
                IEnumerable<string> testModules = GetPropertyValueWrapper(sessionStartEventArgs);
                if (_eqtTrace.IsInfoEnabled)
                {
                    _eqtTrace.Info("{0}: TestModules: '{1}'",
                        CoverletConstants.DataCollectorName,
                        string.Join(",", testModules ?? Enumerable.Empty<string>()));
                }
                return testModules;
            }
            catch (MissingMethodException ex)
            {
                throw new MissingMethodException("Make sure to use .NET core SDK Version >= 2.2.300", ex);
            }
        }
```

Issue #296, opened 2019-01-15
Viewing instrumented assemblies


sharwell:
I would like to inspect the instrumented assembly for a project, e.g. with ILSpy. Is there an easy way to create the instrumented assembly for this purpose?

ViktorHofer:
If you add a reference to coverlet.msbuild you can just the required properties to invoke this target: https://github.com/tonerdo/coverlet/blob/master/src/coverlet.msbuild/coverlet.msbuild.targets#L20

MarcoRossignoli:
@sharwell can we close this?Does @ViktorHofer's idea works?
Usually similar to @ViktorHofer's idea I copy targets on "tests" and point `<UsingTask>` to debug build path, after with Debug.Launch() I attach to VS inside coverlet(and rebuild), in this way I stop code before "restore" dll and I can analyze instrumented dll.

sharwell:
@MarcoRossignoli There isn't a good approach today but I'm hoping we can implement it after the project PRs are merged.

MarcoRossignoli:
@sharwell I found a way:
```
dotnet msbuild /t:Restore,Build,InstrumentModulesAfterBuild /p:CollectCoverage=true "C:\git\coverlet-issue-343\DemoLibraryTests\DemoLibraryTests.csproj" /v:m
```

MarcoRossignoli:
Feel free to re-open if needed

PR #577, opened 2019-10-03
Allow to prevent RestoreOriginalModule when GetCoverageResult is called

shargon:
There are specific situations when you want to preserve the Instrumentation. For example fuzzing techniques, we are working on an fuzzer using this library for the coverage and we need this change.

Thanks for your amazing work

### DIFF
--- a/src/coverlet.core/Coverage.cs
+++ b/src/coverlet.core/Coverage.cs
@@ -128,7 +128,7 @@ namespace Coverlet.Core
             };
         }
 
-        public CoverageResult GetCoverageResult(bool revertModules = true)
+        public CoverageResult GetCoverageResult(bool restoreModules = true)
         {
             CalculateCoverage();
 
@@ -214,7 +214,7 @@ namespace Coverlet.Core
                 }
 
                 modules.Add(Path.GetFileName(result.ModulePath), documents);
-                if (revertModules) _instrumentationHelper.RestoreOriginalModule(result.ModulePath, _identifier);
+                if (restoreModules) _instrumentationHelper.RestoreOriginalModule(result.ModulePath, _identifier);
             }
 
             var coverageResult = new CoverageResult { Identifier = _identifier, Modules = modules, InstrumentedResults = _results };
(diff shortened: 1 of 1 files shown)


shargon:
For prevent the revert after the process die we use this

```csharp

                var helper = (IInstrumentationHelper)DependencyInjection.Current.GetService(typeof(IInstrumentationHelper));

                var field = (ConcurrentDictionary<string, string>)helper
                    .GetType()
                    .GetField("_backupList", BindingFlags.Instance | BindingFlags.NonPublic)
                    .GetValue(helper);

                field.Clear();
```

There are an easier way?

MarcoRossignoli:
Hi @shargon before go on with this we need to re-open discussion with @tonerdo @petli on future of core lib.
As you can see on this PR https://github.com/tonerdo/coverlet/pull/566 we're discussing if has got sense expose a stable "core" functionality lib.
You can read some history on linked issue https://github.com/tonerdo/coverlet/pull/566#issuecomment-536053774 https://github.com/tonerdo/coverlet/issues/212

The issue with this kind of update is that are not "well structured" and "planned". I mean at the moment we don't guarantee core lib types stability(reason for hide everything) so we're concerned about take direct dependency on classes in these assembly.
We could break in future.

My idea was hide all and only public pertinent object through interfaces/factories and be free to update internal implementation.

MarcoRossignoli:
@shargon to view instrumented asm you could use also msbuild https://github.com/tonerdo/coverlet/issues/296#issuecomment-470527068

shargon:
If you don't expose a stable "core" nuget, you are killing some good ideas. Is possible with a fork yes, but have sense to do that?

MarcoRossignoli:
> If you don't expose a stable "core" nuget, you are killing some good ideas

cc: @tonerdo @petli

MarcoRossignoli:
@shargon close this, because we'll start to shape public api.
Until that moment you'll need to clone and add a signature to your asm and adding to friendly asm on core lib. Let me know if you're in trouble.

PR #583, opened 2019-10-08
Improve lambda scenario coverage

MarcoRossignoli:
contributes to https://github.com/tonerdo/coverlet/issues/343

This code fix the issue of "false covered lines/branches" for lambda and async lambda bit complex scenario.

Capture

We've again some issue with unexpected reported branch, but for now is better than nothing.
Other solution I found trim out valid branches so for now I prefer false positive uncovered branch than covered.

cc: @driseley @tonerdo @petli @SteveGilham 

@daveMueller take a look if you're interested in the intrumentation part, you can start from test, are debuggable pass true to `TestInstrumentationHelper.Run(.,disableRestoreModule` and `RemoteExecutor.Invoke(...,invokeLocal`), if you're in throuble let me know.

### DIFF
--- a/test/coverlet.core.tests/InstrumenterHelper.cs
+++ b/test/coverlet.core.tests/InstrumenterHelper.cs
@@ -4,6 +4,7 @@ using System.IO;
 using System.Linq;
 using System.Reflection;
 using System.Runtime.CompilerServices;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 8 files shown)


test/coverlet.core.tests/InstrumenterHelper.cs
```diff
@@ -363,4 +432,23 @@ public void LogWarning(string message)
             File.AppendAllText(_logFile, message + Environment.NewLine);
         }
     }
+
+    class InstrumentationHelperForDebugging : InstrumentationHelper
```
MarcoRossignoli: @shargon  I did this to avoid restore and make instrumentation completely debuggable

PR #596, opened 2019-10-18
Add extension for getservice and replace usages

### LINKED ISSUE #554: Add extension `GetService<T>` to DependencyInjection for better readability
Add extension to `Coverlet.Core.DependencyInjection`ot  cleanup a bit the code, something like 
```cs
    public static class DependencyInjectionExtensions
    {
        public static T GetService<T>(this IServiceProvider serviceProvider)
        {
            return (T)serviceProvider.GetService(typeof(T));
        }
    }
```
And update the callers.
NB. We could use `Microsoft.Extensions.DependencyInjection.Abstractions` but we want avoid to have too much transitive/reference.

flerka:
Fix #554 this issue.
Add extension `GetService<T>` to DependencyInjection for better readability and replaced all usages.

### DIFF
--- a/src/coverlet.core/Extensions/DependencyInjectionExtensions.cs
+++ b/src/coverlet.core/Extensions/DependencyInjectionExtensions.cs
@@ -1,0 +1,14 @@
+﻿using System;
+using Coverlet.Core.Attributes;
+
+namespace coverlet.core.Extensions
+{
+    public static class DependencyInjectionExtensions
+    {
+        [ExcludeFromCoverage]
+        public static T GetService<T>(this IServiceProvider serviceProvider)
+        {
+            return (T)serviceProvider.GetService(typeof(T));
+        }
+    }
+}
--- a/src/coverlet.core/Extensions/DependencyInjectionExtensions.cs
+++ b/src/coverlet.core/Extensions/DependencyInjectionExtensions.cs
@@ -1,11 +1,9 @@
 ﻿using System;
-using Coverlet.Core.Attributes;
 
-namespace coverlet.core.Extensions
+namespace Coverlet.Core.Extensions
 {
     public static class DependencyInjectionExtensions
     {
-        [ExcludeFromCoverage]
         public static T GetService<T>(this IServiceProvider serviceProvider)
         {
             return (T)serviceProvider.GetService(typeof(T));
--- a/test/coverlet.collector.tests/CoverletCoverageDataCollectorTests.cs
+++ b/test/coverlet.collector.tests/CoverletCoverageDataCollectorTests.cs
@@ -15,6 +15,7 @@ using Xunit;
 using Coverlet.Collector.DataCollection;
 using Coverlet.Core.Reporters;
 using Coverlet.Core.Abstracts;
+using coverlet.core.Extensions;
 
 namespace Coverlet.Collector.Tests
 {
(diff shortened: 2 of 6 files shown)


src/coverlet.core/Extensions/DependencyInjectionExtensions.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using Coverlet.Core.Attributes;
+
+namespace coverlet.core.Extensions
```
MarcoRossignoli: nit: namespace casing `Coverlet.Core.Extensions`

src/coverlet.core/Extensions/DependencyInjectionExtensions.cs
```diff
@@ -0,0 +1,14 @@
...
+{
+    public static class DependencyInjectionExtensions
+    {
+        [ExcludeFromCoverage]
+        public static T GetService<T>(this IServiceProvider serviceProvider)
+        {
+            return (T)serviceProvider.GetService(typeof(T));
```
MarcoRossignoli: Why `[ExcludeFromCoverage]`? core lib won't be instrumented so we don't need to exclude, you can remove it.

test/coverlet.collector.tests/CoverletCoverageDataCollectorTests.cs
```diff
@@ -74,7 +75,10 @@ public void OnSessionStartShouldPrepareModulesForCoverage()
                     null,
                     _context);
             IDictionary<string, object> sessionStartProperties = new Dictionary<string, object>();
-            Coverage coverage = new Coverage("abc.dll", null, null, null, null, null, true, true, "abc.json", true, It.IsAny<ILogger>(), (IInstrumentationHelper)DependencyInjection.Current.GetService(typeof(IInstrumentationHelper)), (IFileSystem)DependencyInjection.Current.GetService(typeof(IFileSystem)));
+            Coverage coverage = new Coverage("abc.dll", null, null, null, null, null, true, true, "abc.json", true,
```
MarcoRossignoli: please keep declaration in one line...it's not so long at the moment, I prefer all in one line or all stacked.

src/coverlet.core/Extensions/DependencyInjectionExtensions.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using Coverlet.Core.Attributes;
+
+namespace coverlet.core.Extensions
+{
+    public static class DependencyInjectionExtensions
```
MarcoRossignoli: when we'll merge https://github.com/tonerdo/coverlet/pull/566 we'll hide as internals...for now it's ok.

flerka: Ok, for now it will stay public.

src/coverlet.core/Extensions/DependencyInjectionExtensions.cs
flerka: Fixed, thanks!

test/coverlet.collector.tests/CoverletCoverageDataCollectorTests.cs
flerka: You are right consistency is better, fixed.

src/coverlet.core/Extensions/DependencyInjectionExtensions.cs
flerka: Fixed, I added it as it was on `HelperExtensions`. Maybe if core lib won't be instrumented I should remove `[ExcludeFromCoverage]` from ``HelperExtensions` too?

MarcoRossignoli: Ah ok...understood...I think it was placed to remove from "autocoverage" that we do on build.

Issue #605, opened 2019-11-06
Import exclusions from file


thviQit:
We got the idea, that it would be nice to have a file with all your source file exclusions, instead of providing a long list to the build task.
In this way, you can have the exclusions under version control and you do not have to edit your build every time you need to exclude a file.

I missed the CONTRIBUTION.md (even though it is upper case ...), so I already have an implementation, but I have refrained from creating a pull request.

I hope it is something you would be interested in.

thviQit:
You can see it here: https://github.com/thviQit/coverlet/tree/exclusions-from-file

MarcoRossignoli:
Hi @thviQit thank's for the idea...you could open a PR without problem when you want and we can discuss on it.

First thing I'm not so "happy" is that we've at the moment 3 types of exclusion
-by types
-by file
-by attributes
so we should include all types of filters and not favour one, I mean I used a lot types filtering in my experience but never used files...depends on "usage".

the second one is that we've also the opposite concept that is "include" so to be symmetric in design we should also add inclusion filter file.

I wonder if someone can reach same goal on CI simply composing run script with a "script" that take exclusion from file(versionable).

MarcoRossignoli:
Close for too few thumbs up/interest from community.

PR #607, opened 2019-11-06
Exclusions from file

thviQit:
PR for issue #605 

Updated coverlet.console and coverlet.msbuild.task to take a new parameter (a file path).
The file path is for a file with a list of source files to be excluded.
Updated documentation to reflect the use


MarcoRossignoli:
Hi @thviQit thank's for the idea...you could open a PR without problem when you want and we can discuss on it.

First thing I'm not so "happy" is that we've at the moment 3 types of exclusion
-by types
-by file
-by attributes
so we should include all types of filters and not favour one, I mean I used a lot types filtering in my experience but never used files one...depends on "usage".

the second one is that we've also the opposite concept that is "include" so to be symmetric in design we should also add inclusion filter file.

I wonder if someone can reach same goal on CI simply composing run script with a "script" that take exclusion from file(versionable).

thviQit:
I understand the concerns about symmetry, but it should be relative easy to have an include version of this.

I guess you could create a script to do the same, if you use coverlet as a dotnet global tool. It could still make a quite unreadable command line though.
I didn't consider it, as we use it as a msbuild task.

If someone (me/you/?) invested a little more time in it, a file structure could be created together with a parser, which would allow for all exclusions and inclusions to be handled from one file.
That would keep the interface almost in its current state, so it wouldn't be necessary to introduce parsing and commands for all use cases, that is include/exclude attributes|source files|types ...

MarcoRossignoli:
Some question to better understand:

1) How many filter you have?
2) How many time you update your filter?

thviQit:
We don't have a lot right now, as it is pretty recent that we added coverlet. However we do have a task to go through our projects to find files that should be excluded, e.g. program.cs, startup.cs, static classes, auto generated stuff etc.

So from my point of view it would just be easier to add a line in a text file when in Visual Studio, rather than having to edit the build in Azure Devops

MarcoRossignoli:
> rather than having to edit the build in Azure Devops

You should use yaml file to describe your pipeline and not using remote devops ui...in that way you'll update from VS and will be "commited" to repo.
https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=schema

thviQit:
We are not quite there yet. We have a bunch of custom build tasks that have to be converted to steps first, and that is going to take some time.

MarcoRossignoli:
> We are not quite there yet. We have a bunch of custom build tasks that have to be converted to steps first, and that is going to take some time.

Understood, so you agree with me that this feature would be "useful" only for this particular case...I mean I'm not sure that is correct add a feature(adding complexity to maintain and possible bugs) that is useful only for single case where anyway there is already an elegant solution.
We could keep issue/PR open to understand if it's a common case and eventually go on with feat.

BTW this is my opinion, wait for @tonerdo @petli one.

thviQit:
Well I do think it is nicer to keep this stuff outside the yaml files to keep them as simple as possible.

In the case that you import a step template with the coverlet command from another repository, then you can parameterize the excludes easier by passing a file path to the --exclude-from-file command.

I just think that using a file as input provides some extra flexibility.

No hard feelings so far ;-)

MarcoRossignoli:
> In the case that you import a step template with the coverlet command from another repository, then you can parameterize the excludes easier by passing a file path to the --exclude-from-file command.

FYI You can do it also with yaml https://docs.microsoft.com/en-us/azure/devops/pipelines/process/templates?view=azure-devops#variable-re-use

thviQit:
Cool - I did not know that.
So we use the `dotnet test` command right now (via the ui), but we have a template like this:
```yaml
- task: DotNetCoreCLI@2
  displayName: 'dotnet test'
  inputs:
    command: test
    projects: ${{ parameters.testProjects }}
    arguments: '--configuration ${{ parameters.buildConfiguration }} --no-build /p:CollectCoverage=true /p:CoverletOutputFormat=cobertura /p:Exclude="${{ parameters.excludedTypes }}" /p:ExcludeByFile="${{ parameters.excludedFiles }}" /p:ExcludeByAttribute="${{ parameters.excludedAttributes }}"'
```
So in our local repositories, where we define the pipeline, we could have a dedicated yaml fille with the excludes like this:
```yaml
# coverage-settings.yaml
variables:
  excludedSourceFiles: '../dir1/*.cs,../**/Program.cs'
  excludedAttributes: ''
  excludedTypes: ''
  threshold: ''
  ...
```
And import it like this:
```yaml
#jobtemplate.yaml
variables:
- template: coverage-settings.yml
```
and use it like `/p:ExcludedByFile="${{excludedSourceFiles}}"`.
The only thing I don't like with this is that it is still one long string instead of a list, which would make it easier to read.

MarcoRossignoli:
> The only thing I don't like with this is that it is still one long string instead of a list, which would make it easier to read.

BTW I don't expect too much stuff because otherwise you should negate the concept and use include.

thviQit:
That might be better.
If you reject the pull request based on this, it is completely understandable (and might be for the best?). 
I'll just write of my wasted hours as education 😁

MarcoRossignoli:
> I'll just write off my wasted hours as education 😁

Ideas to improve coverlet value are always welcome, we discuss everything.
Usually we add features based on the number of devs ask for it(or grouping similar issue) and also after that we need to be "critical" much as possible to keep the product stable and maintainable, so my first approach is to understand if there is a solution with "no code".
In this case seem that there is a better/elegant and less invasive solution to problem.

May I suggest next time to open an issue with the "feature idea" before work on PR, you did a great work on this and it's a shame if won't be merged 😞 
As a contributor of other projects I know the feeling :expressionless:

BTW for the moment I'll leave PR and issue open we'll see next months and if we'll get more thumbs up or interest.

Thank's you for the effort here!

MarcoRossignoli:
No thumbs up after a month, closing for now, we'll re-open if needed.

Issue #257, opened 2018-12-03
Support Source Link


ViktorHofer:
If an assembly supports Source Link it generates the following (sample) code in the pdb:

```
{"documents": { "E:\\A\\_work\\104\\s\\*" : "https://raw.githubusercontent.com/dotnet/coreclr/5feb0389f1e1ba715fba5560c979ac6362b1bcf8/*" }}
```

As we write the source code paths into the report we should add a switch, i.e. -sourcelink (true/false) which replaces the local file paths with the source link ones. A corresponding issue in the ReportGenerator repo was opened to support fetching remote files: https://github.com/danielpalme/ReportGenerator/issues/180

ViktorHofer:
ping @tonerdo

tonerdo:
@ViktorHofer looking into this over the weekend. Work's been a bit crazy for me this week

ViktorHofer:
No prob. Take your time :)

PR #614, opened 2019-11-10
Fix cobertura Jenkins reporter  + source link support

### LINKED ISSUE #197: Wrong source path in cobertura report with similar project name starts
Hi there,

I have 3 project in my solution:

- DAL
- DAL.MySql
- DAL.MsSql

When I generate a cobertura report I get a report with the following `sources` part.

```
<sources>
    <source>{Path to my solution}\DAL</source>
</sources>
```

I think this is a mistake since this is not a common parent folder for each 3 project. I think the following line of code is responsible for that because it is checking that every modules' path  starting with the same string even though in this case it is true it is not a common parent folder for the modules.

https://github.com/tonerdo/coverlet/blob/5217033c744f27f63db3fd1339ed3d8faa5fb34d/src/coverlet.core/Reporters/CoberturaReporter.cs#L149

### LINKED ISSUE #295: Difference in cobertura report between coverlet 2.4.0 and 2.5.0
### Issue description
When I run tests using command:
```
dotnet test /p:CopyLocalLockFileAssemblies=true /p:Exclude=\"[System.*]*\" /p:CollectCoverage=true /p:CoverletOutputFormat=cobertura
```

in my sample project https://github.com/rwasala/CodeCoverage I get different output in reports.
For version 2.4.0 it looks like:
```
<?xml version="1.0" encoding="UTF-8"?>
<coverage line-rate="1" branch-rate="1" version="1.9" timestamp="1547558846" lines-covered="26" lines-valid="26" branches-covered="4" branches-valid="4">
  <sources>
    <source>/var/lib/jenkins/workspace/Build_CodeCoverage_master/SampleLib/</source>
  </sources>
  <packages>
    <package name="SampleLib" line-rate="1" branch-rate="1" complexity="10">
      <classes>
        <class name="SampleLib.Author" filename="Author.cs" line-rate="1" branch-rate="0" complexity="3">
          <methods>
            <method name="get_Id" signature="()" line-rate="1" branch-rate="0">
```
For version 2.5.0 it looks like:
```
<?xml version="1.0" encoding="UTF-8"?>
<coverage line-rate="1" branch-rate="1" version="1.9" timestamp="1547558132" lines-covered="26" lines-valid="26" branches-covered="4" branches-valid="4">
  <sources/>
  <packages>
    <package name="SampleLib" line-rate="1" branch-rate="1" complexity="10">
      <classes>
        <class name="SampleLib.Author" filename="/var/lib/jenkins/workspace/Build_CodeCoverage_master/SampleLib/Author.cs" line-rate="1" branch-rate="1" complexity="3">
          <methods>
            <method name="get_Id" signature="()" line-rate="1" branch-rate="1">
```

Basically the difference is that in version 2.5.0 node source is empty which causes problems with cobertura Jenkins plugin - it is not showing the source code with marked what is covered and what is not.

Is it possible to make report from `coverlet.msbuild` in version 2.5.0 the same like in 2.4.0?

### LINKED ISSUE #408: Cobertura reporter breaks on 2.6.0
@tonerdo I noticed similar issue reported in https://github.com/tonerdo/coverlet/issues/295
I recently upgraded coverlet version from 2.3.1 to 2.6.0 (some of my team members upgraded from 2.5.0 to 2.6.0) and our Cobertura reporter started failing.

Below is the log:
```
[Cobertura] Publishing Cobertura coverage report...

[Cobertura] No coverage results were found using the pattern '**/TestResults/coverage.*.xml' relative to '/home/ec2-user/workspace/'.  Did you enter a pattern relative to the correct directory?  Did you generate the XML report(s) for Cobertura?
```
This is our jenkin pipeline step that is failing right now with 2.6.0:
```
step([$class             : 'CoberturaPublisher',
                      autoUpdateHealth   : false,
                      autoUpdateStability: false,
                      coberturaReportFile: '**/TestResults/coverage.*.xml',
                      failUnhealthy      : false,
                      failUnstable       : false,
                      maxNumberOfBuilds  : 0,
                      onlyStable         : false,
                      sourceEncoding     : 'ASCII',
                      zoomCoverageChart  : false])
```
Same reporter works fine with 2.3.1 and my team members reported worked until 2.5.0.
Please can you look into this? If you need anything from my side please let me know.

daveMueller:
closes #197, closes #408 and really closes #295 and maintains #257


daveMueller:
Filename attribute is now relative to system root and the reports can properly be displayed in jenkins running on linux or windows.
SourceFile

SourceFileLinux

MarcoRossignoli:
Thanks David bit busy next week I'll take a look asap!

src/coverlet.core/CoverageResult.cs
```diff
@@ -41,6 +41,7 @@ public class CoverageResult
     {
         public string Identifier;
         public Modules Modules;
+        public bool IsSourceLinkUsed;
```
MarcoRossignoli: nit: I prefer to maintain switch name `UseSourceLink`

src/coverlet.core/Reporters/CoberturaReporter.cs
```diff
@@ -1,7 +1,9 @@
 using System;
+using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
 using System.Linq;
+using System.Runtime.InteropServices;
```
MarcoRossignoli: Is this necessary?

src/coverlet.core/Reporters/CoberturaReporter.cs
```diff
@@ -129,5 +132,37 @@ public string Report(CoverageResult result)
 
             return Encoding.UTF8.GetString(stream.ToArray());
         }
+
+        private static IEnumerable<string> GetRootDirs(Modules modules, bool isSourceLinkUsed)
```
MarcoRossignoli: why not
```cs
if (isSourceLinkUsed)
{
	return new[] { string.Empty };
}
else
{
	return modules.Values.SelectMany(k => k.Keys).Select(Directory.GetDirectoryRoot).Distinct();
}
```
Does it work?

src/coverlet.core/Reporters/CoberturaReporter.cs
```diff
@@ -129,5 +132,37 @@ public string Report(CoverageResult result)
...
+
+                sources.AddRange(
+                    module.Value.Select(d => Path.GetDirectoryName(d.Key)));
+            }
+
+            sources = sources.Distinct().ToList();
+            return sources.Select(Directory.GetDirectoryRoot).Distinct();
+        }
+
+        private static string GetRelativePathFromBase(IEnumerable<string> basePaths, string path, bool isSourceLinkUsed)
```
MarcoRossignoli: rename `rootPaths`

src/coverlet.core/Reporters/CoberturaReporter.cs
```diff
@@ -129,5 +132,37 @@ public string Report(CoverageResult result)
...
+        {
+            if (isSourceLinkUsed) return path;
+
+            string relativePath = path;
+            basePaths.ToList().ForEach(basePath =>
+            {
+                if (path.StartsWith(basePath))
+                {
+                    relativePath = path.Substring(basePath.Length);
```
MarcoRossignoli: what do you think about
```cs
            if (isSourceLinkUsed)
            {
                return path;
            }

            foreach (var root in basePaths)
            {
                if (path.StartsWith(root))
                {
                    return path.Substring(root.Length);
                }
            }

            Debug.Assert(false, "Unexpected, we should find at least one path starts with one pre-build root");

            return path;
```
it seems more logic to me...when we found one valid root we can return immediatly, if we don't found any valid root(unexpected) we return original path, so we avoid exception and we'll generate a file(better than nothing), I added also an assertion so on debug we'll catch possible problem.

MarcoRossignoli:
Thank's @daveMueller for the effort here and sorry for the delay.

I'd like to add more tests on this
1) one test with 2 file from two different windows roots i.e. c:\proj\file.cs and e:\proj\file.cs 
2) one test with source link parameter true

MarcoRossignoli:
@daveMueller let me know if you want complete this or if can finish this PR on top of your precious work!

daveMueller:
> 
> 
> @daveMueller let me know if you want complete this or if can finish this PR on top of your precious work!

@MarcoRossignoli Yes I will complete it. I already did most of the things you mentioned locally. My daytime job has been keeping me busy last week but I think I will find some free time on the weekend.

MarcoRossignoli:
No worries no rush, only asked to know!

src/coverlet.core/CoverageResult.cs
daveMueller: Done

src/coverlet.core/Reporters/CoberturaReporter.cs
daveMueller: Done

src/coverlet.core/Reporters/CoberturaReporter.cs
daveMueller: Yes I tested it. Done.

src/coverlet.core/Reporters/CoberturaReporter.cs
daveMueller: Done

src/coverlet.core/Reporters/CoberturaReporter.cs
daveMueller: Yes, I agree.

daveMueller:
@MarcoRossignoli I still want to refactore the tests. Had some problems to get them running on X plat and so they are still not really clean. If you have some ideas what could be changed, please let me know.

test/coverlet.core.tests/Reporters/CoberturaReporterTests.cs
```diff
@@ -120,5 +134,67 @@ public void TestReport()
...
+
+            var classes = new Classes { { "Class", new Methods() } };
+            var documents = new Documents { { absolutePath1, classes }, { absolutePath2, classes } };
+
+            result.Modules = new Modules { { "Module", documents } };
+
+            CoberturaReporter reporter = new CoberturaReporter();
+            string report = reporter.Report(result);
+
+            var doc = XDocument.Load(new MemoryStream(Encoding.UTF8.GetBytes(report)));
```
MarcoRossignoli: What do you think about 
```cs
List<string> rootPaths = doc.Element("coverage").Element("sources").Elements().Select(e => e.Value).ToList();
List<string> relativePaths = doc.Element("coverage").Element("packages").Element("package").Element("classes").Elements().Select(e => e.Attribute("filename").Value).ToList();
List<string> possiblePaths = new List<string>();
foreach (var root in rootPaths)
{
    foreach (string relativePath in relativePaths)
    {
        possiblePaths.Add(Path.Combine(root, relativePath));
    }
}
            
Assert.Contains(absolutePath1, possiblePaths);
Assert.Contains(absolutePath2, possiblePaths);
```
This should be the expected alg of every reporter that can read a cobertura format...we try every permutation root+path and check if exists

test/coverlet.core.tests/Reporters/CoberturaReporterTests.cs
```diff
@@ -120,5 +134,67 @@ public void TestReport()
...
+
+            var classes = new Classes { { "Class", new Methods() } };
+            var documents = new Documents { { absolutePath, classes } };
+
+            result.Modules = new Modules { { "Module", documents } };
+
+            CoberturaReporter reporter = new CoberturaReporter();
+            string report = reporter.Report(result);
+
+            var doc = XDocument.Load(new MemoryStream(Encoding.UTF8.GetBytes(report)));
```
MarcoRossignoli: ok maybe cleanup a bit with `doc.Element("coverage").Element("packages").Element("package").Element("classes").Elements().Select(e => e.Attribute("filename").Value).Single();`

MarcoRossignoli:
@daveMueller can you test and attach also a reportgenerator result?Only to be sure to be compatible with it(most used tool).

daveMueller:
@MarcoRossignoli are the attached screenshots sufficient?

Here is the relevant part of a cobertura report:
image

And this is the reportgenerator output for the cobertura report.
image

test/coverlet.core.tests/Reporters/CoberturaReporterTests.cs
daveMueller: Yes I changed it that way.

test/coverlet.core.tests/Reporters/CoberturaReporterTests.cs
daveMueller: Done

MarcoRossignoli:
yep thank you

MarcoRossignoli:
@danielpalme can you take a look at report generated at https://github.com/tonerdo/coverlet/pull/614#issuecomment-562889735 and double confirm that it's ok?

danielpalme:
Your code looks fine to me.
_ReportGenerator_ iterates through the `source` elements and combines the value with the `filename` attribute. The first combined file path that exists on disk is used.
Another option would be to just use the absolute path for the the `filename` attribute. Then no "guessing" is required. 
 Advantages:
- Faster
- More reliable (example a class with same name exists in several directories/namespaces)

See:
https://github.com/danielpalme/ReportGenerator/blob/master/src/ReportGenerator.Core/Parser/Preprocessing/CoberturaReportPreprocessor.cs

MarcoRossignoli:
> Another option would be to just use the absolute path for the the filename attribute. Then no "guessing" is required

This is the old behavior, but we need to split root+relative due to Jenkins reporter, it doesn't work well without `sources`

MarcoRossignoli:
Thank's Dave!

PR #615, opened 2019-11-11
Allow coverlet integration with other MSBuild test strategies

sharwell:
This change exposes the InstrumentModules and GenerateCoverageResult targets which execute unconditionally, and can then be tied into build steps using conditional logic. The previous logic for supporting the VSTest target is retained through conditional targets.


MarcoRossignoli:
@sharwell where do you need it?

sharwell:
@MarcoRossignoli I had to duplicate most of the targets file in dotnet/roslyn-analyzers#3028. With this change, I'll be able to just duplicate the `InstrumentModulesNoBuild` and `GenerateCoverageResultAfterTest` targets.

MarcoRossignoli:
Understood...I don't have idea of release at the moment we're discussing off line...if you want to use this you can dogfood it tomorrow https://github.com/tonerdo/coverlet/blob/master/Documentation/ConsumeNightlyBuild.md

MarcoRossignoli:
Here the unreleased feat https://github.com/tonerdo/coverlet/blob/master/Documentation/Changelog.md

sharwell:
@MarcoRossignoli Thanks. I'll test out the nightly but there's no rush on the release itself.

PR #628, opened 2019-12-02
[Experiment]Expose coverlet.core.dll as nuget package

MarcoRossignoli:
Tackle https://github.com/tonerdo/coverlet/issues/212

At the moment we can use coverlet in two main scenario:
* Classic coverage tool, workflow :
1) run instrumentation
2) run code
3) get coverage data

* Real time coverage analysis 
1) run instrumentation
2) inject coverlet inprocess and read real time coverage data
3) get coverage data or simply close

You can checkout this branch and I did a complete working sample under `Documentation/Examples/StandaloneUsage` you can follow `Readme.md` in same folder to test by yourself.
Let me know if you're in trouble and if the api would be better.

cc possible interested guys @ppumkin @shargon @singhsarab @jnovick @vagisha-nidhi @acesiddhu @karanjitsingh @sharwell 

cc: @tonerdo @petli 

Video sample https://twitter.com/MarcoRossignoli/status/1201546945589841922


src/coverlet.core/Abstracts/IInstrumenter.cs
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    public interface ICoverageEngineFactory
+    {
+        ICoverageEngine CreateEngine(InstrumentationOptions options);
+        IReporter CreateReporter(string format);
+        IInProcessCoverageEngine CreateInProcessEngine(Stream instrumentationResultStream);
+    }
+
+    public interface ICoverageEngine
```
petli: Could this be IDisposable instead and clean up the instrumented files on dispose? It would let the calling code control when it happens instead of it being done on process exit.

MarcoRossignoli: > clean up the instrumented files

You mean "restore non instrumented ones"?

petli: Yes :)

MarcoRossignoli: Could be possible...I need to check how much code update...but I think that we could inject a slighly different instrumenterhelper that don't fire restore on process exits, I'll take a look, thank's.

petli: The latter is also a good tool to look at instrumented files themselves.  Could be an enum on `InstrumenterOptions` when calling `ICoverageEngineFactory.CreateEngine`:

    enum RestoreInstrumentedModules
    {
           OnProcessExit,
           OnEngineDispose,
           Never
     }

It would mean that `ICoverageEngine` still need to be `IDisposable`, even if it is only cleans up resources in some cases.

MarcoRossignoli: I'll try to add this feat.

MarcoRossignoli:
Closing for now, we'll take some inspiration in future when we'll decide to expose instrumentation engine.

Issue #623, opened 2019-11-23
Incorrect branch coverage with async/await test


tfadler:
If I am under the wrong impression then please let me know.

But I expected scenarios like the below to have full branch coverage.

Reproduced with .net core 3.0 and coverlet 2.7.0

**Code**

```

namespace ClassLibrary1
{
	public class Class1
	{
		public async Task<int> DoAsync()
		{
			await Task.Delay(100);
			return 42;
		}
	}
}
```

**Test Code**

```
namespace ClassLibrary1Test
{
	public class Class1Test
	{
		[Fact]
		public async Task Test1()
		{
			var objectToTest = new Class1();

			var result = await objectToTest.DoAsync().ConfigureAwait(false);

			Assert.Equal(42, result);
		}
	}
}
```

I run

dotnet test /p:CollectCoverage=true

And here is my contents of my coverage.json

I did not expect the branches.  

In our real code base we are using /p:CoverletOutputFormat=opencover  and reportgenerator and the results for branch is less than a 100%

I expected full branch coverage.

```
{
  "ClassLibrary1.dll": {
    "C:\\Users\\name1\\source\\repos\\ClassLibrary1\\ClassLibrary1\\Class1.cs": {
      "ClassLibrary1.Class1/<DoAsync>d__0": {
        "System.Void ClassLibrary1.Class1/<DoAsync>d__0::MoveNext()": {
          "Lines": {
            "8": 1,
            "9": 1,
            "10": 1,
            "11": 1
          },
          "Branches": [
            {
              "Line": 9,
              "Offset": 35,
              "EndOffset": 37,
              "Path": 0,
              "Ordinal": 0,
              "Hits": 1
            },
            {
              "Line": 9,
              "Offset": 35,
              "EndOffset": 101,
              "Path": 1,
              "Ordinal": 1,
              "Hits": 1
            }
          ]
        }
      }
    }
  }
}
```

MarcoRossignoli:
Hi @tfadler!

> I did not expect the branches.

Actually there is a generated state machine by c# compiler so that branch exists(c# lie)

We recently did some improvements to async/await to remove unexpected branches without risk too much https://github.com/tonerdo/coverlet/pull/549 and will be available on next release

BTW I did some test with your code and it's reported 100% to me on report generator(you can also check by yourself...lines and branches has got 1 in "Hits" element, so I expect full coverage(also with compiler generated branch noise).

[...]

PR #634, opened 2019-12-03
Fix ConfigureAwait state machine generated branches

MarcoRossignoli:
closes https://github.com/tonerdo/coverlet/issues/623

C# compiler use a different state machine class in case of `await task.ConfiguraAwait(bool)`, we added new type to async/await state machine branches filter.

@tfadler thanks for repro, day after merge of this you'll can try fix using nightly build https://github.com/tonerdo/coverlet/blob/master/Documentation/ConsumeNightlyBuild.md

cc @tonerdo @petli


tfadler:
Thanks @MarcoRossignoli !

Issue #650, opened 2019-12-09
[coverlet] Unable to instrument module for pdbs while launching tests in multiple threads


tempora-mutantur:
Hello.
We collect code coverage using coverlet.msbuild version 2.7.0. We do this on Windows Server VMs  and we launch our NUnit tests for a *.sln file using the "dotnet test MySolution.sln ..." command. This launches tests in parallel for every test project, and as I expect, this leads to such warnings:

_'[coverlet] Unable to instrument module: C:\Repositories\myproject.git\Tests\Output\UnitTests\TerrificMethods\Core.dll' because : The process cannot access the file 'C:\Repositories\myproject.git\src\Core\obj\debug\Core.pdb' because it is being used by another process._

Because of this code coverage can change from one tests launch to another.
Thank you.

MarcoRossignoli:
Is it a warning or an error?can you copy/paste full line?
Can you provide a small repro?

tempora-mutantur:
It's a warning,

C:\Users\savanchuk\.nuget\packages\coverlet.msbuild\2.7.0\build\coverlet.msbuild.targets(7,5): warning : [coverlet] Unable to instrument module: 'C:\Repositories\repo.git\Tests\Output\UnitTests\Server.UnitTests\Server.Views.dll' because : The process cannot access the file 'C:\Repositories\repo.git\src\Server\obj\Debug\netcoreapp2.1\Server.Views.pdb' because it is being used by another process. [C:\Repositories\repo.git\src\Server.UnitTests\Server.UnitTests.csproj]

Maybe I'm missing something, but it seems that it would just skip this PDB in this case.

https://github.com/tonerdo/coverlet/blob/57ec8d39a86af31defbd4ce6585f71775f9ae471/src/coverlet.core/Coverage.cs#L117

I think the repro would be a couple of test projects that use some common dll and launch tests for these projects in parallel.

I may prepare a little sample.

MarcoRossignoli:
Yep could be great a repro...because I did a check on code base and we open pdbs always in read mode share read so it's strange...I'd like to understand who is locking that guy.

tempora-mutantur:
Could this be the reason?
https://github.com/tonerdo/coverlet/blob/63dbda8190c505d293c7266445143b11b02f56e9/src/coverlet.core/Helpers/InstrumentationHelper.cs#L223

MarcoRossignoli:
[...]

PR #656, opened 2019-12-12
Fix pdb file locking during instrumentation

MarcoRossignoli:
fixes https://github.com/tonerdo/coverlet/issues/650

@tempora-mutantur thank's for the repro, this PR should fix the issue.

You can try tomorrow using nightly build https://github.com/tonerdo/coverlet/blob/master/Documentation/ConsumeNightlyBuild.md


tempora-mutantur:
@MarcoRossignoli 
Greatly appreciate, I will.

PR #671, opened 2019-12-27
Avoid to instrument compiler generated code when generated from excluded methods

### LINKED ISSUE #129: [ExcludeFromCodeCoverage] lets through lambdas, yield return and await
This code
```
    public class Class1
    {
        [ExcludeFromCodeCoverage]
        public int F1(string input)
        {
            Func<char, int> f = (c => { return (int)c; });

            return input.Select(x =>
            {
                var l = new List<char> { x };
                return l.Select(f).Last();
            }).Sum();
        }

        [ExcludeFromCodeCoverage]
        public IEnumerable<int> F2(string input)
        {
            foreach (char c in input)
            {
                yield return c;
            }
        }

        [ExcludeFromCodeCoverage]
        public async Task<string> F3(string input)
        {
            var f = new StreamReader(File.OpenRead(input));
            var r = await f.ReadToEndAsync();
            return r;
        }

        public static string F4(string input)
        {
            return input.ToUpperInvariant();
        }

        [ExcludeFromCodeCoverage]
        public string F5(string input)
        {
            return input.ToUpperInvariant();
        }
    }
```
with a test to exercise method `F4`.

Expect just method `F4` in the coverage file, with 100% coverage
Get 15.8% line coverage, with the lambda expressions in `F1` and generated `MoveNext` methods in the implementations of `F2` and `F3` showing uncovered.

### LINKED ISSUE #670: ExcludeFromCodeCoverage attribute is ignored in an static class in a ASP.NET Core 3.1 project
Hi all,
I'm facing an issue when using Coverlet in a ASP.NET Core 3.1 project. I can see that a static class, which has a static void method, is not being excluded from code coverage, although both the class and the method have the ExcludeFromCodeCoverage attribute applied, from the namespace `System.Diagnostics.CodeAnalysis`.

This static method is an extension method referenced by the Startup class in my ASP.NET Core 3.1 project, Startup and Program classes are also excluded from code coverage and they don't appear in the code coverage results, however my static class and its static void method appear in the results.

Is this a known issue with Coverlet, .NET Core 3.1 and static classes and methods? This is the command I'm running to get the results:

`dotnet test /p:CollectCoverage=true /p:CoverletOutputFormat=cobertura`

My unit test project references coverlet.msbuild with version 2.7.0. It also references Xunit, version 2.4.1, and my ASP.NET Core 3.1 project. The unit test project also targets .NET Core 3.1. This is extracted from my csproj file for the reference to coverlet.msbuild:

`<PackageReference Include="coverlet.msbuild" Version="2.7.0">
      <PrivateAssets>all</PrivateAssets>
      <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
    </PackageReference>`

I found this issue in my build pipeline in Azure DevOps too.

Finally, I'd like to mention that the code coverage tools included in Visual Studio 2019 Enterprise, version 16.4.2, do exclude the class and the method from code coverage, so I think this is an issue with Coverlet itself, or the commands I'm using with `dotnet test`.

Please advise.

Regards,
Jorge

matteoerigozzi:
Fix #129 
Fix #670

Change instrumenter to keeping track of methods with ExcludeFromCodeCoverage attribute and exclude compiler-generated code for lamda, yield, async/await and local functions.

Check here  
https://github.com/dotnet/roslyn/blob/1fcbc51/src/Compilers/CSharp/Portable/Symbols/Synthesized/GeneratedNames.cs
for compiler synthesized names.


MarcoRossignoli:
Some test fail, they seems a bit flaky, one time fails on debug and other time on release, maybe we should try to find something more stable.

matteoerigozzi:
Fix also #670

src/coverlet.core/Instrumentation/Instrumenter.cs
```diff
@@ -180,8 +182,21 @@ private void InstrumentModule()
...
                             && (!_isCoreLibrary || actualType.FullName != "System.Threading.Interlocked")
                             && !_instrumentationHelper.IsTypeExcluded(_module, actualType.FullName, _excludeFilters)
-                            && _instrumentationHelper.IsTypeIncluded(_module, actualType.FullName, _includeFilters))
-                            InstrumentType(type);
+                            && _instrumentationHelper.IsTypeIncluded(_module, actualType.FullName, _includeFilters)
+                            )
+                        {
+                            if (IsSynthesizedMemberToBeExcluded(type))
+                            {
+                                if (_excludedCompilerGeneratedTypes is null)
```
MarcoRossignoli: FYI 

```cs
_excludedCompilerGeneratedTypes &= new List<string>();
```

src/coverlet.core/Instrumentation/Instrumenter.cs
```diff
@@ -349,8 +365,23 @@ private void InstrumentType(TypeDefinition type)
...
+                    continue;
+                }
+
                 if (!customAttributes.Any(IsExcludeAttribute))
+                {
                     InstrumentMethod(method);
+                }
+                else
+                {
+                    if (_excludedMethods is null)
```
MarcoRossignoli: ditto `??=`

MarcoRossignoli:
@matteoerigozzi I did some small updates+add documentation can you take a look?

PR #679, opened 2020-01-08
Trim whitespace between values when reading from configuration from runsettings

### LINKED ISSUE #641: In coverlet.collector, allow repeating configuration elements in runsettings
When excluding multiple elements, the current format can make things a bit unwieldy (and makes the changes much harder to track
```
<Exclude>[lib]ns1.*,[lib]ns2.*</Exclude>
```
It would be nice if we could specify multiple items on different Exclude nodes:
```
<Exclude>[lib]ns1.*</Exclude>
<Exclude>[lib]ns2.*</Exclude>
```
Same goes for the other colon separated settings

EricStG:
Closes #641

### DIFF
--- a/src/coverlet.collector/DataCollection/CoverletSettingsParser.cs
+++ b/src/coverlet.collector/DataCollection/CoverletSettingsParser.cs
@@ -86,7 +86,7 @@ namespace Coverlet.Collector.DataCollection
             if (configurationElement != null)
             {
                 XmlElement reportFormatElement = configurationElement[CoverletConstants.ReportFormatElementName];
-                formats = reportFormatElement?.InnerText?.Split(',').Select(format => format.Trim())
+                formats = this.SplitElement(reportFormatElement)?
                     .Where(format => !string.IsNullOrEmpty(format)).ToArray();
             }
 
@@ -101,7 +101,7 @@ namespace Coverlet.Collector.DataCollection
         private string[] ParseIncludeFilters(XmlElement configurationElement)
         {
             XmlElement includeFiltersElement = configurationElement[CoverletConstants.IncludeFiltersElementName];
-            return includeFiltersElement?.InnerText?.Split(',');
+            return this.SplitElement(includeFiltersElement);
         }
 
         /// <summary>
@@ -112,7 +112,7 @@ namespace Coverlet.Collector.DataCollection
         private string[] ParseIncludeDirectories(XmlElement configurationElement)
         {
             XmlElement includeDirectoriesElement = configurationElement[CoverletConstants.IncludeDirectoriesElementName];
-            return includeDirectoriesElement?.InnerText?.Split(',');
+            return this.SplitElement(includeDirectoriesElement);
         }
 
         /// <summary>
@@ -127,7 +127,7 @@ namespace Coverlet.Collector.DataCollection
             if (configurationElement != null)
             {
                 XmlElement excludeFiltersElement = configurationElement[CoverletConstants.ExcludeFiltersElementName];
-                string[] filters = excludeFiltersElement?.InnerText?.Split(',');
+                string[] filters = this.SplitElement(excludeFiltersElement);
                 if (filters != null)
                 {
                     excludeFilters.AddRange(filters);
@@ -145,7 +145,7 @@ namespace Coverlet.Collector.DataCollection
         private string[] ParseExcludeSourceFiles(XmlElement configurationElement)
         {
             XmlElement excludeSourceFilesElement = configurationElement[CoverletConstants.ExcludeSourceFilesElementName];
-            return excludeSourceFilesElement?.InnerText?.Split(',');
+            return this.SplitElement(excludeSourceFilesElement);
         }
 
         /// <summary>
@@ -156,7 +156,7 @@ namespace Coverlet.Collector.DataCollection
         private string[] ParseExcludeAttributes(XmlElement configurationElement)
         {
             XmlElement excludeAttributesElement = configurationElement[CoverletConstants.ExcludeAttributesElementName];
-            return excludeAttributesElement?.InnerText?.Split(',');
+            return this.SplitElement(excludeAttributesElement);
         }
 
         /// <summary>
@@ -204,6 +204,16 @@ namespace Coverlet.Collector.DataCollection
             XmlElement includeTestAssemblyElement = configurationElement[CoverletConstants.IncludeTestAssemblyElementName];
             bool.TryParse(includeTestAssemblyElement?.InnerText, out bool includeTestAssembly);
             return includeTestAssembly;
+        }
+
+        /// <summary>
+        /// Splits a comma separated elements into an array
+        /// </summary>
+        /// <param name="element">The element to split</param>
+        /// <returns>An array of the values in the element</returns>
+        private string[] SplitElement(XmlElement element)
+        {
+            return element?.InnerText?.Split(',')?.Select(value => value.Trim())?.ToArray();
         }
     }
 }
--- a/src/coverlet.collector/DataCollection/CoverletSettingsParser.cs
+++ b/src/coverlet.collector/DataCollection/CoverletSettingsParser.cs
@@ -212,7 +212,7 @@ namespace Coverlet.Collector.DataCollection
         /// <returns>An array of the values in the element</returns>
         private string[] SplitElement(XmlElement element)
         {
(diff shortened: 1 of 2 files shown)


MarcoRossignoli:
Thank's I'll review asap!

src/coverlet.collector/DataCollection/CoverletSettingsParser.cs
```diff
@@ -205,5 +205,15 @@ private bool ParseIncludeTestAssembly(XmlElement configurationElement)
...
         }
+
+        /// <summary>
+        /// Splits a comma separated elements into an array
+        /// </summary>
+        /// <param name="element">The element to split</param>
+        /// <returns>An array of the values in the element</returns>
+        private string[] SplitElement(XmlElement element)
+        {
+            return element?.InnerText?.Split(',')?.Select(value => value.Trim())?.ToArray();
```
MarcoRossignoli: Maybe `return element?.InnerText?.Split(',', StringSplitOptions.RemoveEmptyEntries).Select(value => value.Trim()).ToArray();` why the null check after `InnerText` one?

test/coverlet.collector.tests/CoverletSettingsParserTests.cs
```diff
@@ -42,17 +42,24 @@ public void ParseShouldSelectFirstTestModuleFromTestModulesList()
             Assert.Equal("module1.dll", coverletSettings.TestModule);
         }
 
-        [Fact]
-        public void ParseShouldCorrectlyParseConfigurationElement()
+        [Theory]
+        [InlineData("[*]*,[coverlet]*", "[coverlet.*.tests?]*,[coverlet.*.tests.*]*", @"E:\temp,/var/tmp", "module1.cs,module2.cs", "Obsolete,GeneratedCodeAttribute,CompilerGeneratedAttribute")]
+        [InlineData("[*]*, [coverlet]*", "[coverlet.*.tests?]*, [coverlet.*.tests.*]*", @"E:\temp, /var/tmp", "module1.cs, module2.cs", "Obsolete, GeneratedCodeAttribute, CompilerGeneratedAttribute")]
+        [InlineData("[*]*,\r\n[coverlet]*", "[coverlet.*.tests?]*,\r\n[coverlet.*.tests.*]*", "E:\\temp,\r\n/var/tmp", "module1.cs,\r\nmodule2.cs", "Obsolete,\r\nGeneratedCodeAttribute,\r\nCompilerGeneratedAttribute")]
+        public void ParseShouldCorrectlyParseConfigurationElement(string includeFilters,
```
MarcoRossignoli: Can you add some test with empty space and null?

src/coverlet.collector/DataCollection/CoverletSettingsParser.cs
EricStG: I've updated it and added a coalesce operator at the end to return an empty array instead of null. The extra null checks weren't needed so they are gone.

test/coverlet.collector.tests/CoverletSettingsParserTests.cs
EricStG: Added a few cases for missing elements, null inner texts and whitespace

src/coverlet.collector/DataCollection/CoverletSettingsParser.cs
MarcoRossignoli: > I've updated it and added a coalesce operator at the end to return an empty array instead of null. The extra null checks weren't needed so they are gone.

I prefer to keep old semantic and return null, I'd like to avoid surprise in case of expected null of callers. Please remove the coalesce at the end, the rest is ok to me.

MarcoRossignoli: It's also a shame, maybe in future we'll update with elegant one https://github.com/dotnet/corefx/issues/41487#issuecomment-573576875

test/coverlet.collector.tests/CoverletSettingsParserTests.cs
MarcoRossignoli: thank's!

test/coverlet.collector.tests/CoverletSettingsParserTests.cs
```diff
@@ -42,17 +42,27 @@ public void ParseShouldSelectFirstTestModuleFromTestModulesList()
...
         }
 
-        [Fact]
-        public void ParseShouldCorrectlyParseConfigurationElement()
+        [Theory]
+        [InlineData("[*]*,[coverlet]*", "[coverlet.*.tests?]*,[coverlet.*.tests.*]*", @"E:\temp,/var/tmp", "module1.cs,module2.cs", "Obsolete,GeneratedCodeAttribute,CompilerGeneratedAttribute")]
+        [InlineData("[*]*,,[coverlet]*", "[coverlet.*.tests?]*,,[coverlet.*.tests.*]*", @"E:\temp,,/var/tmp", "module1.cs,,module2.cs", "Obsolete,,GeneratedCodeAttribute,,CompilerGeneratedAttribute")]
+        [InlineData("[*]*, ,[coverlet]*", "[coverlet.*.tests?]*, ,[coverlet.*.tests.*]*", @"E:\temp, ,/var/tmp", "module1.cs, ,module2.cs", "Obsolete, ,GeneratedCodeAttribute, ,CompilerGeneratedAttribute")]
+        [InlineData("[*]*, [coverlet]*", "[coverlet.*.tests?]*, [coverlet.*.tests.*]*", @"E:\temp, /var/tmp", "module1.cs, module2.cs", "Obsolete, GeneratedCodeAttribute, CompilerGeneratedAttribute")]
+        [InlineData("[*]*,\r\n[coverlet]*", "[coverlet.*.tests?]*,\r\n[coverlet.*.tests.*]*", "E:\\temp,\r\n/var/tmp", "module1.cs,\r\nmodule2.cs", "Obsolete,\r\nGeneratedCodeAttribute,\r\nCompilerGeneratedAttribute")]
```
MarcoRossignoli: Please add also some test with space before and after CRLF for instance

`"   [*]*,    \r\n[coverlet]*"   , "[coverlet.*.tests?]*   ,   \r\n   [coverlet.*.tests.*]*   "`

EricStG: Added those as well as a few tabs for good measure

src/coverlet.collector/DataCollection/CoverletSettingsParser.cs
EricStG: Coalesce has been removed

test/coverlet.collector.tests/CoverletSettingsParserTests.cs
MarcoRossignoli: Good catch...test fails on tabs

test/coverlet.collector.tests/CoverletSettingsParserTests.cs
```diff
@@ -62,22 +78,68 @@ public void ParseShouldCorrectlyParseConfigurationElement()
...
+            Assert.Null(coverletSettings.ExcludeSourceFiles);
+            Assert.Null(coverletSettings.ExcludeAttributes);
+            Assert.Single(coverletSettings.ExcludeFilters, "[coverlet.*]*");
+        }
+
         [Theory]
         [InlineData(" , json", 1, new[] { "json" })]
         [InlineData(" , json, ", 1, new[] { "json" })]
         [InlineData("json,cobertura", 2, new[] { "json", "cobertura" })]
+        [InlineData("json,\r\ncobertura", 2, new[] { "json", "cobertura" })]
```
MarcoRossignoli: Please add some space and tabs also here

test/coverlet.collector.tests/CoverletSettingsParserTests.cs
EricStG: That was an error in the test. `@"\t"` is not the same as `"\t"`. It's been fixed.

test/coverlet.collector.tests/CoverletSettingsParserTests.cs
EricStG: Added a few cases with space, tabs and a mix of everything

test/coverlet.collector.tests/CoverletSettingsParserTests.cs
MarcoRossignoli: I see

Issue #689, opened 2020-01-13
Inconsistent behaviour with nested types and function values


SteveGilham:
In upgrading to collector v1.2 from v1.1 I noticed that code like
```
module Whatever =
...
  do AppDomain.CurrentDomain.DomainUnload.Add(FlushCounter DomainUnload)
     AppDomain.CurrentDomain.ProcessExit.Add(FlushCounter ProcessExit)
```
started registering as uncovered. I'm assuming that this is part of the lambda aliasing change in PR #583, and it's triggering because each line compiles to a form with 4 extra classes, and one of those has an unattributed `Invoke()` method that executes only when the event is actually fired.

Proceeding to exclude this not readily testable behaviour by refactoring to
```
module Whatever =
...
  module private Uncoverlet =
    let AddDomainHandlers() =
      AppDomain.CurrentDomain.DomainUnload.Add(FlushCounter DomainUnload)
      AppDomain.CurrentDomain.ProcessExit.Add(FlushCounter ProcessExit)
...
  do Uncoverlet.AddDomainHandlers()
```
and adding an exclude entry `[*]*Uncoverlet` resulted in the uncovered line count from ReportGenerator via OpenCover format dropping by 2, as intended.

Looking at the coverage report, however, showed that rather than being excluded from consideration, the `AddDomainHandlers` method was showing as covered.

There seem to be two problems here.  First, exclude by type name doesn't seem to be honoured for nested types, and second, the new behaviour of detecting multiple items on a line doesn't seem to have propagated down this extra level either.

I haven't explicitly tried the C# equivalent
```
static class Whatever {
  static class Uncoverlet {
    ...
  }
}
```
but the IL in each case will be the same.

MarcoRossignoli:
Thanks for reporting this

SteveGilham:
In a similar C# case  the same behaviour is observed.

Example code
[CoverletNested.zip](https://github.com/tonerdo/coverlet/files/4058276/CoverletNested.zip)

To repro
`dotnet test  --settings ./coverletArgs.runsettings --configuration Debug --framework netcoreapp3.1`
in the `TestNested` folder

MarcoRossignoli:
Thank's a lot for the repro Steve!

PR #694, opened 2020-01-18
Check nested types for exclude filter

MarcoRossignoli:
Fixes https://github.com/tonerdo/coverlet/issues/689

Report [ExcludeFilterNestedAutogeneratedTypes.zip](https://github.com/tonerdo/coverlet/files/4081333/ExcludeFilterNestedAutogeneratedTypes.zip)

cc: @tonerdo @petli @SteveGilham

*This fix also an old bug if type was nested we checked for parent type exclusion only, so if we have more nesting children of nested excluded type wasn't excluded, maybe it's a rare condition never reported.


src/coverlet.core/Instrumentation/Instrumenter.cs
```diff
@@ -177,18 +177,25 @@ private void InstrumentModule()
...
+                            !actualType.CustomAttributes.Any(IsExcludeAttribute)
+
                             // Instrumenting Interlocked which is used for recording hits would cause an infinite loop.
                             && (!_isCoreLibrary || actualType.FullName != "System.Threading.Interlocked")
-                            && !_instrumentationHelper.IsTypeExcluded(_module, actualType.FullName, _excludeFilters)
+
+                            && (
+                                 !_instrumentationHelper.IsTypeExcluded(_module, actualType.FullName, _excludeFilters) &&
+                                 // Check exclusion for nested types
+                                 (type.IsNested ? !_instrumentationHelper.IsTypeExcluded(_module, type.FullName, _excludeFilters) : true)
```
MarcoRossignoli: @SteveGilham this is the core update, nested types weren't checked for exclusion filtering.

MarcoRossignoli: @tonerdo do you remember the reason?

test/coverlet.core.tests/Coverage/InstrumenterHelper.cs
```diff
@@ -279,7 +301,7 @@ public static void GenerateHtmlReport(CoverageResult coverageResult, IReporter r
             reportFile = Path.Combine(dir.FullName, Path.ChangeExtension("report", reporter.Extension));
             File.WriteAllText(reportFile, reporter.Report(coverageResult));
             // i.e. reportgenerator -reports:"C:\git\coverlet\test\coverlet.core.tests\bin\Debug\netcoreapp2.0\Condition_If\report.cobertura.xml" -targetdir:"C:\git\coverlet\test\coverlet.core.tests\bin\Debug\netcoreapp2.0\Condition_If" -filefilters:+**\Samples\Instrumentation.cs
-            new Generator().GenerateReport(new ReportConfiguration(
+            Assert.True(new Generator().GenerateReport(new ReportConfiguration(
```
MarcoRossignoli: @Danielpalme after upgrade to 3.1 started to fail to generate reports, is there a way to understand what's wrong?I get only bool result.

MarcoRossignoli:
Checking flaky tests.

src/coverlet.core/Instrumentation/Instrumenter.cs
SteveGilham: Wouldn't it be simpler just to write the check as an iteration up the declaration chain like
```
    var excludedByName = false;
    for (var checkType = type; checkType != null && !excludedByName; checkType = checkType.DeclaringType) 
        excludedByName = _instrumentationHelper.IsTypeExcluded(_module, checkType.FullName, _excludeFilters)
```

MarcoRossignoli: You mean if current type or one of his parent type are excluded exclude it. Is this the correct semantic?I mean suppose deep nesting A->B->C and I want to skip only B, if I understood well in this case we'll skip also C, right?

MarcoRossignoli: Actually at the moment the behavior is what you expect with only the first direct parent, so the semantic should be if I'm out also all my children and children of my children will be out. I'll give it a try with your idea Steve thanks!

test/coverlet.core.tests/Coverage/InstrumenterHelper.cs
danielpalme: Could you please try again with version 4.4.4 of `ReportGenerator.Core`.

Until 4.4.3 the structure was:
```
lib\netstandard2.0
lib\netcoreapp
```

With 4.4.4 I added:
```
lib\netcoreapp3.0
```

.NET Core 3.x seems to ignore DLLs in `lib\netcoreapp`.

src/coverlet.core/Instrumentation/Instrumenter.cs
SteveGilham: Of course, as `type.DeclaringType.FullName` is always a leading substring of `type.FullName`, simply taking a string `"*MatchEndOfClassName"` to regex `".*MatchEndOfClassName\/?" (as Cecil indicates nesting by `/``), the match could be simply done without looking for declaring types or being nested.

Iteration would still be required to pick up things like exclusions by attribute, though

test/coverlet.core.tests/Coverage/InstrumenterHelper.cs
MarcoRossignoli: I'll try asap thanks

MarcoRossignoli: Got same error but the dll seem ok(new one)
image

danielpalme: Will debug this myself tomorrow evening.

danielpalme: Will debug this myself tomorrow evening.

danielpalme: Will debug this myself tomorrow evening.

MarcoRossignoli: No worries no rush thank you!

src/coverlet.core/Instrumentation/Instrumenter.cs
```diff
@@ -141,6 +141,37 @@ public InstrumenterResult Instrument()
             return _result;
         }
 
+        // If current type or one of his parent is excluded we'll exclude it
+        // If I'm out every my children and every children of my children will be out
+        private bool IsTypeExcluded(TypeDefinition type)
```
MarcoRossignoli: @SteveGilham moved all on helper and added some tests also for attributes.

src/coverlet.core/Instrumentation/Instrumenter.cs
```diff
@@ -141,6 +141,37 @@ public InstrumenterResult Instrument()
...
+                }
+            }
+
+            return false;
+        }
+
+        // Instrumenting Interlocked which is used for recording hits would cause an infinite loop.
+        private bool Is_System_Threading_Interlocked_CoreLib_Type(TypeDefinition type)
+        {
+            return _isCoreLibrary && type.FullName == "System.Threading.Interlocked";
```
MarcoRossignoli: @ViktorHofer I moved core lib check on helper to make code clearer

test/coverlet.core.tests/Instrumentation/InstrumenterTests.cs
```diff
@@ -23,14 +23,10 @@ public class InstrumenterTests
         private readonly InstrumentationHelper _instrumentationHelper = new InstrumentationHelper(new ProcessExitHandler(), new RetryHelper(), new FileSystem());
         private readonly Mock<ILogger> _mockLogger = new Mock<ILogger>();
 
-        [Fact(Skip = "To be used only validating System.Private.CoreLib instrumentation")]
+        [Fact]
```
MarcoRossignoli: @ViktorHofer added `System.Private.CoreLib.dll` tests on CI to validate interlocked class(was manual).

test/coverlet.core.tests/Coverage/InstrumenterHelper.cs
danielpalme: Does the problem still exist? You resolved this issue and merged the change into master, that's why I'm asking.

MarcoRossignoli: Yes still exists I merged because I use automatic generation of report only for debugging purpouse. So to move stuff forward I merged this fix.

MarcoRossignoli: You can clone repo
```
dotnet pack
dotnet build
dotnet test
```
And you can try with this test https://github.com/tonerdo/coverlet/blob/master/test/coverlet.core.tests/Coverage/CoverageTests.cs#L114 enabling that row and it will generate report during test.

MarcoRossignoli: Feel free to ping me if you're in throuble.

danielpalme: I've sent a PR #696 which finally resolves the problem.
The Nuget behavior seems to have changed in .NET Core 3.x. Not all files in `lib\netcoreapp3.0` get copied during building the project.

PR #697, opened 2020-01-21
Flow ILogger to InstrumentationHelper

### LINKED ISSUE #559: Flow `ILogger` to InstrumentationHelper
We should flow the logger to instrumentation helper and emit important log  warning here

https://github.com/tonerdo/coverlet/blob/82a920864a157bfaa9417540a20a4c03b336f1d0/src/coverlet.core/Helpers/InstrumentationHelper.cs#L155-L159

cc: @tonerdo @petli

daveMueller:
closes #559


daveMueller:
@MarcoRossignoli Ok there is no instance of ```ILogger``` registered in the container. Mhh I see a ```MSBuildLogger```, ```ConsoleLogger``` and ```TestPlatformLogger```. Should one of these be used regarding to the driver? 
But ```TestPlatformLogger``` currently is no instance of ```ILogger```?
Any help on how I should proceed?

MarcoRossignoli:
> Ok I got it. But I think we could/should keep a IServiceCollection in Coverlet.Core with services that are used in every driver. I would just add the specific services in Services.cs of the driver and then build the ServiceProvider there.

Not sure, I prefer a replication of builder, because today we need only to exchange ILogger, but tomorrow?If we embed part o services in core in future we'll need to update also core lib. I prefer something like asp.net core startup...built in there is no middleware added to pipeline also if we know that 99% of people will use MVC but we need to specify `AddMvc`, they don't add Mvc by design to core.
It's more clean and a small replication is expected here, we're building-up services to work well in that context.
Does it make sense for you?

daveMueller:
> Does it make sense for you?

Yes absolutely, I will do it like that.

daveMueller:
@MarcoRossignoli There are tests in ```Coverlet.Core.Tests``` that are using the static ```Coverlet.Core.DependencyInjection``` class. So now there isn't a service provider in core namespace anymore. How should I replace this when I remove ```DependencyInjection```. just new the classes up?

MarcoRossignoli:
Yes we should provide another container for tests with same services

MarcoRossignoli:
@daveMueller if you're in trouble with test part feel free to update this PR and we'll do some pair programming and I'll help to understand how to fix that part.

daveMueller:
Ok thanks I pushed everything and would be grateful for any help. I'm currently facing two problems. 

1. In ```CoverletCoverageCollector``` in the overridden initialize I can't resolve the ```TestPlatformLogger``` because the ```DataCollectionLogger``` can't be resolved which in the end leads to problem that I can't resolve a ```CoverletLogger```. 
2. As I said before the tests that referenced the container in core. Currently I  just added providers to the project so that it can compile. Actually I didn't spend much time with this issue because the problem with the ```CoverletLogger``` occupied me.

MarcoRossignoli:
> In CoverletCoverageCollector in the overridden initialize I can't resolve the TestPlatformLogger because the DataCollectionLogger can't be resolved which in the end leads to problem that I can't resolve a CoverletLogger.

Don't worry about `CoverletCoverageCollector` constructor, our goal is to flow `ILogger` inside `Coverage`  not create services for all, so you should change `CoverageWrapper.CreateCoverage` in `CoverageWrapper.CreateCoverage(CoverletSettings,ILogger,IInstrumentationHelper)`, we should remove static call to `Current`, this statics was useful in the beginning but now we have to flow all instance and not reference directly the container. 

I don't like too much shared container...are comfortable because we don't need add parameter to method and constructors but make all classes less testable and also if one day I want to change my "container system" or remove it code become ugly.

You should build service at the end `Initialize` where you have all service available, on `OnSessionStart` you can ask for needed services, this should be always possible because this class is the "entry point"(like when we setup services in a `Program.cs`) for collectors.

You can avoid custom file for services, a private helper "IServiceProvider GetServiceProvider(...)" inside `CoverletCoverageCollector.cs` is enough, you can set it to a private field and use it where you need.

> As I said before the tests that referenced the container in core. Currently I just added providers to the project so that it can compile. Actually I didn't spend much time with this issue because the problem with the CoverletLogger occupied me.

Here it's the same we only have to get 2 object from service container `IInstrumentationHelper/IFileSystem` , and you should add only services "needed" for that context, allow to simulate real instrumentation.

https://github.com/tonerdo/coverlet/blob/a2b0373d60b0893240997f4a3d86c4f460ab9dae/test/coverlet.core.tests/Coverage/InstrumenterHelper.cs#L356-L367

For now here statics is ok because I use it on test verification so I need to get [`IInstrumentationHelper`](https://github.com/tonerdo/coverlet/blob/master/test/coverlet.core.tests/Coverage/InstrumenterHelper.cs#L345), but I'll change in future with an instance.

Statics should be avoided like plague 😄 slowly I'll remove from everywhere.

src/coverlet.msbuild.tasks/Services.cs
```diff
@@ -8,24 +8,9 @@
 
 namespace coverlet.msbuild.tasks
 {
-    internal static class Services
+    public class Services
```
MarcoRossignoli: Internal

src/coverlet.msbuild.tasks/CoverageResultTask.cs
```diff
@@ -76,16 +77,18 @@ public string CoverletMultiTargetFrameworksCurrentTFM
 
         public CoverageResultTask()
         {
-            _logger = Services.Current.GetService<ILogger>();
+           _serviceProvider = new Services().GetServiceProvider();
```
MarcoRossignoli: MsbuildLogger needs TaskLogginHelper so you need to pass it to `GetServiceProvider`  https://github.com/tonerdo/coverlet/blob/63dbda8190c505d293c7266445143b11b02f56e9/src/coverlet.msbuild.tasks/MSBuildLogger.cs#L15

MarcoRossignoli: The same for InstrumentationTask

src/coverlet.console/Program.cs
```diff
@@ -6,24 +6,26 @@
...
 
 namespace Coverlet.Console
 {
     class Program
     {
         static int Main(string[] args)
         {
-            var logger = Services.Current.GetService<ILogger>();
-            var fileSystem = Services.Current.GetService<IFileSystem>();
+            var serviceProvider = GetServiceProvider();
```
MarcoRossignoli: Build container here and avoid to pass to `DependencyInjectionExtensions` we'll remove it from core.

src/coverlet.msbuild.tasks/Services.cs
daveMueller: Done

src/coverlet.msbuild.tasks/CoverageResultTask.cs
daveMueller: Done

src/coverlet.console/Program.cs
daveMueller: Done and removed ```DependencyInjectionExtensions```

daveMueller:
I still need to fix tests tomorrow.

src/coverlet.console/Program.cs
```diff
@@ -23,9 +22,17 @@ class Program
...
-            var serviceProvider = GetServiceProvider();
-            var logger = DependencyInjectionExtensions.GetService<ILogger>(serviceProvider);
-            var fileSystem = DependencyInjectionExtensions.GetService<IFileSystem>(serviceProvider);
+            IServiceCollection serviceCollection = new ServiceCollection();
+            serviceCollection.AddTransient<IRetryHelper, RetryHelper>();
+            serviceCollection.AddTransient<IProcessExitHandler, ProcessExitHandler>();
+            serviceCollection.AddTransient<IFileSystem, FileSystem>();
+            serviceCollection.AddTransient<ILogger, ConsoleLogger>();
+            serviceCollection.AddSingleton<IInstrumentationHelper, InstrumentationHelper>();
+            var serviceProvider = serviceCollection.BuildServiceProvider();
```
MarcoRossignoli: Please avoid `var` if right expression type is not "clear"

src/coverlet.msbuild.tasks/InstrumentationTask.cs
```diff
@@ -113,7 +113,7 @@ private void WaitForDebuggerIfEnabled()
 
         public override bool Execute()
         {
-            var serviceProvider = new Services().GetServiceProvider();
+            var serviceProvider = new Services().GetServiceProvider(Log);
```
MarcoRossignoli: ditto no `var`

src/coverlet.msbuild.tasks/CoverageResultTask.cs
```diff
@@ -74,16 +77,18 @@ public string CoverletMultiTargetFrameworksCurrentTFM
 
         public CoverageResultTask()
         {
-            _logger = new MSBuildLogger(Log);
+           _serviceProvider = new Services().GetServiceProvider(Log);
```
MarcoRossignoli: move this to `Execute` so we're aligned with other VSTask

src/coverlet.core/coverlet.core.csproj
```diff
@@ -17,4 +17,8 @@
     <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Folder Include="Logging\" />
```
MarcoRossignoli: why this?

src/coverlet.core/Enums/LogLevel.cs
```diff
@@ -1,4 +1,4 @@
-namespace Coverlet.Console.Logging
+namespace coverlet.core.Enums
```
MarcoRossignoli: Why?

src/coverlet.core/Abstracts/ILogger.cs
```diff
@@ -1,9 +1,11 @@
 using System;
+using coverlet.core.Enums;
 
 namespace Coverlet.Core.Abstracts
 {
     internal interface ILogger
     {
+        LogLevel Level { get; set; }
```
MarcoRossignoli: You don't have to move this piece of code...ILogger interface is stable...the "level" is needed only for console driver so the implementation remains there with level and you build the container inside `app.OnExecute` so you can register proper console logger level.

src/coverlet.console/Program.cs
daveMueller: Done

src/coverlet.msbuild.tasks/InstrumentationTask.cs
daveMueller: Done

src/coverlet.msbuild.tasks/CoverageResultTask.cs
daveMueller: Done

src/coverlet.core/coverlet.core.csproj
daveMueller: Sorry I missed to remove it. Done.

src/coverlet.core/Enums/LogLevel.cs
daveMueller: This was related to the changes on ```ILogger```. See below.

src/coverlet.core/Abstracts/ILogger.cs
daveMueller: Sorry I don't quite understand it. In ```Coverlet.Console``` when I build the container in ```app.OnExecute``` there is no logger when ```app.OnExecute``` throws before ```ILogger``` is resolved. 
That's why I created the container directly in main before ```app.OnExecute``` is invoked. That's also the reason why I added the ```Level``` property to the interface and moved the enum. An alternative would be a direct cast of the resolved ```ILogger``` to ```ConsoleLogger``` but I guess this would break the benefit of using a DI container.

MarcoRossignoli: You're right, for now register a singleton and cast to logger inside `OnExecute`, I need to understand if we can use IOptions<T> pattern in some way to register an option object with default at the start and update inside `OnExecute` and for every new instance the options is used.

daveMueller:
@MarcoRossignoli Could you tell me whats wrong with the integration tests. I currently can't get them running locally on master branch. So I'm a bit confused right now. I don't think my changes make them fail.

src/coverlet.msbuild.tasks/CoverageResultTask.cs
```diff
@@ -22,7 +23,7 @@ public class CoverageResultTask : Task
         private string _thresholdStat;
         private string _coverletMultiTargetFrameworksCurrentTFM;
         private ITaskItem _instrumenterState;
-        private MSBuildLogger _logger;
+        private ILogger _logger;
```
MarcoRossignoli: We can remove this...and use local one in `Execute`

test/coverlet.collector.tests/CoverletCoverageDataCollectorTests.cs
```diff
@@ -42,6 +44,36 @@ public CoverletCoverageDataCollectorTests()
...
+            var serviceCollection = new ServiceCollection();
+            serviceCollection.AddTransient<IRetryHelper, RetryHelper>();
+            serviceCollection.AddTransient<IProcessExitHandler, ProcessExitHandler>();
+            serviceCollection.AddTransient<IFileSystem, FileSystem>();
+            serviceCollection.AddSingleton<IInstrumentationHelper, InstrumentationHelper>();
+            serviceCollection.AddTransient<ILogger, TestCoverletLogger>();
+            _serviceProvider = serviceCollection.BuildServiceProvider();
+        }
+
+        class TestCoverletLogger : ILogger {
```
MarcoRossignoli: Can we create a `Mock<ILogger>().Object` in ctor and use it...less code

test/coverlet.core.tests/Coverage/CoverageTests.cs
```diff
@@ -10,8 +10,8 @@ namespace Coverlet.Core.Tests
 {
     public partial class CoverageTests
     {
-        private readonly InstrumentationHelper _instrumentationHelper = new InstrumentationHelper(new ProcessExitHandler(), new RetryHelper(), new FileSystem());
-        private readonly Mock<ILogger> _mockLogger = new Mock<ILogger>();
+        private static readonly Mock<ILogger> _mockLogger = new Mock<ILogger>();
```
MarcoRossignoli: keep instance semantic...it's not a great problem for test this instances(no static)

test/coverlet.core.tests/Coverage/InstrumenterHelper.cs
```diff
@@ -341,8 +341,19 @@ public static CoverageResult GetCoverageResult(string filePath)
             {
                 Assert.DoesNotContain("not found for module: ", message);
             });
+
+            var serviceCollection = new ServiceCollection();
```
MarcoRossignoli: We have to return same container in `GetCoverageResult` and `Run`.

They run out of process to simulate real behaviour(real instumentation), I propose to create a local static IServiceProvider that will be set inside `Run` and used in `GetCoverageResult` they're always used in sequence and so we simulate real behaviour because we keep `disableRestoreModules` container semantic.

something similar to

```cs
static void SetContainer(bool disableRestoreModules)
{
   var serviceCollection = new ServiceCollection();
            serviceCollection.AddTransient<IRetryHelper, CustomRetryHelper>();
            serviceCollection.AddTransient<IProcessExitHandler, CustomProcessExitHandler>();
            serviceCollection.AddTransient<IFileSystem, FileSystem>();
            if (disableRestoreModules)
            {
                serviceCollection.AddSingleton<IInstrumentationHelper, InstrumentationHelperForDebugging>();
            }
            else
            {
                serviceCollection.AddSingleton<IInstrumentationHelper, InstrumentationHelper>();
            }
         _processWideContainer = serviceCollection.BuildServiceProvider();
}
async public static Task<CoveragePrepareResult> Run<T>(Func<dynamic, Task> callMethod, Func<string, string[]> includeFilter = null, Func<string, string[]> excludeFilter = null, string persistPrepareResultToFile = null, bool disableRestoreModules = false)
{
   ...
  SetContainer(disableRestoreModules);
  ...
  
}
public static CoverageResult GetCoverageResult(string filePath)
{
    ...Coverage coverage = new Coverage(coveragePrepareResultLoaded, _processWideContainer.Get...,
}
```
And add this check ` Assert.DoesNotContain("not found for module: ", message);` at the end of verbose method in `Tests.Logger`.

test/coverlet.core.tests/Instrumentation/InstrumenterTests.cs
```diff
@@ -20,8 +20,8 @@ namespace Coverlet.Core.Instrumentation.Tests
 {
     public class InstrumenterTests
     {
-        private readonly InstrumentationHelper _instrumentationHelper = new InstrumentationHelper(new ProcessExitHandler(), new RetryHelper(), new FileSystem());
-        private readonly Mock<ILogger> _mockLogger = new Mock<ILogger>();
+        private static readonly Mock<ILogger> _mockLogger = new Mock<ILogger>();
```
MarcoRossignoli: ditto keep instance

test/coverlet.core.tests/Helpers/InstrumentationHelperTests.cs
```diff
@@ -3,12 +3,15 @@
...
 using System.Linq;
+using Coverlet.Core.Abstracts;
+using Moq;
 
 namespace Coverlet.Core.Helpers.Tests
 {
     public class InstrumentationHelperTests
     {
-        private readonly InstrumentationHelper _instrumentationHelper = new InstrumentationHelper(new ProcessExitHandler(), new RetryHelper(), new FileSystem());
+        private static readonly Mock<ILogger> _mockLogger = new Mock<ILogger>();
```
MarcoRossignoli: ditto keep instance

MarcoRossignoli:
> @MarcoRossignoli Could you tell me whats wrong with the integration tests. I currently can't get them running locally even when I'm on master branch. So I'm a bit confused right now. I don't think my changes make them fail.

Have you pack before? https://github.com/tonerdo/coverlet/blob/master/CONTRIBUTING.md#building-the-project

test/coverlet.core.tests/Coverage/InstrumenterHelper.cs
daveMueller: Ok I changed it but now some tests fail. Before the change, the static ```DependencyInjection.Current.GetService<T>()``` returned values even when ```DependencyInjection.Set(...)``` wasn't called before.

test/coverlet.collector.tests/CoverletCoverageDataCollectorTests.cs
daveMueller: Done

src/coverlet.msbuild.tasks/CoverageResultTask.cs
daveMueller: Done.

test/coverlet.core.tests/Coverage/CoverageTests.cs
daveMueller: Done

test/coverlet.core.tests/Instrumentation/InstrumenterTests.cs
daveMueller: done

test/coverlet.core.tests/Helpers/InstrumentationHelperTests.cs
daveMueller: Done

daveMueller:
> Have you packed before? https://github.com/tonerdo/coverlet/blob/master/CONTRIBUTING.md#building-the-project
> from root on master
> 
> ```
> git clean -fdx
> dotnet pack
> dotnet test
> ```

Ok thanks you were right. I had to pack again to get the integration tests on master running again. In my branch they still fail so it must be some changes that I made, mhh.

MarcoRossignoli:
> In my branch they still fail so it must be some changes that I made, mhh.

I did a fast check(I'm out for sick at the moment and slowly recovering 🤒 ) and there an issue(my fault) on get test container because Run will run out of process and so GetResult should recreate a singleton new one is not present, something similar to 
```cs
 public static CoverageResult GetCoverageResult(string filePath)
        {
            SetTestContainer();
...
private static void SetTestContainer(bool disableRestoreModules = false)
        {
            LazyInitializer.EnsureInitialized(ref _processWideContainer,
            () =>
            {
                var serviceCollection = new ServiceCollection();
                serviceCollection.AddTransient<IRetryHelper, CustomRetryHelper>();
                serviceCollection.AddTransient<IProcessExitHandler, CustomProcessExitHandler>();
                serviceCollection.AddTransient<IFileSystem, FileSystem>();
                serviceCollection.AddTransient<ILogger, Logger>();
                if (disableRestoreModules)
                {
                    serviceCollection.AddSingleton<IInstrumentationHelper, InstrumentationHelperForDebugging>();
                }
                else
                {
                    serviceCollection.AddSingleton<IInstrumentationHelper, InstrumentationHelper>();
                }
                return serviceCollection.BuildServiceProvider();
            });
        }
```

Second I found an issue in old tests `AsyncAwait_Issue_669_1/AsyncAwait_Issue_669_2` that was miscofigured and ran in-proc, if I set out of proc one of these fails so I'll need to skip for now and I'll check where's the issue

Third also with this fixes on integration tests hang so I think that as you can see there is something wrong somewhere
```
[xUnit.net 00:05:02.30]     Coverlet.Integration.Tests.Msbuild.TestMsbuild_CoverletOutput_Folder_FileNameExtension [FAIL]
  X Coverlet.Integration.Tests.Msbuild.TestMsbuild_CoverletOutput_Folder_FileNameExtension [5m]
  Error Message:
   Command 'dotnet test "C:\git\coverlet\test\coverlet.integration.tests\bin\Debug\netcoreapp3.1\55b88cabcaf54dc6b34df21cdf9e8056" /p:CollectCoverage=true /p:Include="[coverletsamplelib.integration.template]*DeepThought" /p:IncludeTestAssembly=true /p:CoverletOutput="C:\git\coverlet\test\coverlet.integration.tests\bin\Debug\netcoreapp3.1\55b88cabcaf54dc6b34df21cdf9e8056"\file.ext' didn't end after 5 minute
```

So I need some more time to understand what's wrong here.
Thank's for now I'll come back asap. 
Feel free to try to investigate more if you've time!

MarcoRossignoli:
@daveMueller I've a proposal...this PR became too big and "sensible" to do all in one time...what do you think about splitting it in 5?

1) Add service to msbuild
2) Add service to console
3) Add service to collectors
4) Add service to tests
5) Remove DependencyInjection from core.

In this way we can track down PR issue easily, I didn't think that move ILogger to service would lead to a such complexity, I understimate the workload here, my fault, sorry.

You can copy/paste part of this PR that are ok.
Do you agree?

daveMueller:
> I'm out for sick at the moment and slowly recovering 🤒 

Get well soon! Thanks for still responding despite beeing sick.

> what do you think about splitting it in more than one?

I'm OK with that. I already started with 1. yesterday and already ran into the same problems
(hanging MSBuild integration tests).

I continue investigating it and will create the new issues (2-6) today.

MarcoRossignoli:
> I'm OK with that. I already started with 1. yesterday and already ran into the same problems
(hanging MSBuild integration tests).

Weird...seem a trivial update only move ILogger.

> will create the new issues (2-6) today.

You can create an issue with checkbox list:

-Add service to msbuild
-Add service to console
-Add service to collectors
-Add service to tests
-Remove DependencyInjection from core.

Named "Refactor service container".

Issue #584, opened 2019-10-09
Unexpected branches reported for async lambda expression


MarcoRossignoli:
In PR https://github.com/tonerdo/coverlet/pull/583 we fixed the issue with false line coverage for lambda.
Now we report a correct line coverage, but we've unexpected uncovered branches reported.
We should try to avoid to report compiler generated branches.
Test name `Lambda_Issue343`

![Capture](https://user-images.githubusercontent.com/7894084/66461435-50db4300-ea79-11e9-9352-88de7d5075cd.PNG)

FreedCapybara:
For what it's worth, I'm seeing this on a non-lambda:
![coverlet](https://user-images.githubusercontent.com/7003723/73466234-95079180-434f-11ea-812c-8df851fadc48.png)

which works if I de-asyncify the code:
![fixed](https://user-images.githubusercontent.com/7003723/73466739-4d353a00-4350-11ea-9de1-284ef738dda5.png)

(coverlet.msbuild 2.8.0)

MarcoRossignoli:
Should be fixed now on master can you try with nightly build https://github.com/tonerdo/coverlet/blob/master/Documentation/ConsumeNightlyBuild.md?

MarcoRossignoli:
Fixed in https://github.com/tonerdo/coverlet/pull/702

FreedCapybara:
That's okay, I belive you -- I'll just wait 👍

MarcoRossignoli:
cc: @matteoerigozzi

matteoerigozzi:
I agree, should be solved

PR #702, opened 2020-01-24
Improve branch detection for lambda functions and async/await statements

matteoerigozzi:
Propose a fix to #584 

In case of nested compiler generated classes, only the root one presents the ```CompilerGeneratedAttribute```

image

So to detect the MoveNext() method inside an async state machine we need to check recoursively the declaring type for compiler generated attribute.

I also noticed that the operand of the instruction ```ldfld``` inside an async state machine prolog can be ```FieldDefinition``` or ```FieldReference```.
We can see the difference looking at generated IL:

image

image

After this updates we removed some unexpected branches generates by async state machines and the report is this
image

[Here](https://github.com/matteoerigozzi/coverlet/blob/b26eb09d446593a62896776a0e5e244c2a01e19f/src/coverlet.core/Symbols/CecilSymbolHelper.cs#L126) I propose a solution to remove last 2 unexpected branches, that seems to be generated invoking a method with a delegate parameter. I added more tests to check the fairness but I'm not completely sure about it.


src/coverlet.core/Symbols/CecilSymbolHelper.cs
```diff
@@ -19,14 +19,29 @@ internal static class CecilSymbolHelper
     {
         private const int StepOverLineCode = 0xFEEFEE;
 
+        private static bool IsCompilerGenerated(MethodDefinition methodDefinition)
+        {
+            TypeDefinition declaringType = methodDefinition.DeclaringType;
+            while (declaringType != null)
+            {
+                if (declaringType.CustomAttributes.Count(ca => ca.AttributeType.FullName == typeof(CompilerGeneratedAttribute).FullName) > 0)
```
MarcoRossignoli: Can you use Any?Less allocations.

test/coverlet.core.tests/Coverage/CoverageTests.Lambda.cs
```diff
@@ -21,28 +21,27 @@ public void Lambda_Issue343()
                     {
                         instance.InvokeAnonymous_Test();
                         ((Task<bool>)instance.InvokeAnonymousAsync_Test()).ConfigureAwait(false).GetAwaiter().GetResult();
+
                         return Task.CompletedTask;
-                    }, persistPrepareResultToFile: pathSerialize);
+                    }, persistPrepareResultToFile: pathSerialize, disableRestoreModules: true);
```
MarcoRossignoli: revert this, is used only for debugging purpouse

test/coverlet.core.tests/Coverage/CoverageTests.Lambda.cs
```diff
@@ -21,28 +21,27 @@ public void Lambda_Issue343()
...
                     return 0;
-                }, path).Dispose();
+                }, path, invokeInProcess: true).Dispose();
```
MarcoRossignoli: ditto

src/coverlet.core/Symbols/CecilSymbolHelper.cs
```diff
@@ -19,14 +19,29 @@ internal static class CecilSymbolHelper
     {
         private const int StepOverLineCode = 0xFEEFEE;
 
+        private static bool IsCompilerGenerated(MethodDefinition methodDefinition)
```
MarcoRossignoli: Can you add a comment like `// In case of nested state machines CompilerGeneratedAttribute is present only on outermost generated class so we traverse parent/child chain botton-top`

nycdotnet:
Thanks so much for this fix and 2.8.1 - it really cleaned up some of our coverage stats.

MarcoRossignoli:
@nycdotnet you can follow changelog for fix/feat here here https://github.com/tonerdo/coverlet/blob/master/Documentation/Changelog.md#release-date-2020-04-02

Issue #683, opened 2020-01-09
Shutdown msbuild driver innovations?


MarcoRossignoli:
Started from https://github.com/tonerdo/coverlet/issues/678#issuecomment-571585508

@petli comment

A reflection somewhat from the side here is that a lot of these issues around merging, thresholds etc comes from attempts to do coverage collection and analysis being as a single build step. A more unixy way of looking at it is that coverage gets collected during unit tests and written to files, and a separate tool is then run to analyse the coverage in whichever way the user needs. These tools can be combined with a script to get a single command line or build step to run.

(Incidentially, this is how I use coverlet. A script runs all tests with the vstest collector to get coverage files, and then filter and combine in three different ways with reportgenerator to get slightly different views on the results. Partly this is because the early coverlet didn't support anything else, but also because I much prefer the fine-grained control I get this way.)

A major new version av coverlet could look like this:

* The only way to collect coverage is the vstest collector (avoiding all issues with ProcessExit etc)
* The coverlet command line tool holds functionality to merge and analyse coverage (e.g. thresholds). It could also be used to run tests using the vstest collector, and also both run tests and analyse the coverage files in a single step to retain the one-stop-shop command for those who don't need any complex processing.
* The msbuild collector is removed completely.

I think this would make it easier to use coverlet, and simplifying the code a lot. There would be less need for the more complex switches or command line flag combinations, as less common use cases (like the multi-target analysis in this issue or the Linux/Windows combined coverage) could be handled by scripting the coverlet command line tool.

cc: @tonerdo 

EDIT: I've added a documentation draft for new verbs here https://github.com/tonerdo/coverlet/pull/704 please take a look!

MarcoRossignoli:
> The only way to collect coverage is the vstest collector (avoiding all issues with ProcessExit etc)

Eh...this is my "dream"...I'm working on [vstest](https://github.com/microsoft/vstest) side to understand well what missing there, for instance fail on threshold, show data on console etc...that environment lack of some support, but ms team seem available to accomodate our needs.
BTW there are a lot of users that are confortable with msbuild and I'm not sure(need to verify) vstest can run also all netfx(old framework) workload with no issue.

https://nugettrends.com/packages?months=12&ids=coverlet.msbuild&ids=coverlet.collector&ids=coverlet.console

[...]

PR #704, opened 2020-01-25
Add .net tool v2 document design

MarcoRossignoli:
Follow up discussion https://github.com/tonerdo/coverlet/issues/683

cc: @ViktorHofer because coverlet .net tool is used on dotnet runtime repo. Feel free to cc others if needed and this is a breaking change.

I'm not a native speaker/writer so feel free to correct my typos, thank's to all.

### DIFF
--- a/Documentation/Design/.NET Tool v2.md
+++ b/Documentation/Design/.NET Tool v2.md
@@ -1,0 +1,71 @@
+# Functional design .NET tool 2.0 `coverlet`
+
+Tracking issue https://github.com/tonerdo/coverlet/issues/683
+
+The idea is to support some new verbs:
+
+# coverlet run
+
+This will support current instrumentation workflow
+
+```
+coverlet run /path/to/test-assembly.dll --target "dotnet" --targetargs "test /path/to/test-project --no-build"
+```
+
+# coverlet mergereports
+
+Merge `json` default coverlet reports because we should otherwise implement merge for every supported format a huge effort for first version and also with the hope to integrate this feature directly inside collectors.
+
+We'll add an alias to `json` format called `coverlet` to distiguish from possible other future json format and also because this report make sense only inside specific test lifecycle.
+The content of this report is releated to internal structure and was not born to be "portable" or "stable" over time.
+
+```
+coverlet mergereports --coverletreport .\**\coverlet.json --targetdir .\Report\coverlet.xml --format cobertura
+```
+
+`--coverletreport`: will support glob pattern filtering  
+`--format`: specify a supported format  
+`--targetDir`: will create directory folder if not present. If specified a path that ends without a filename(no extension specified) we'll create a filename in that directory called `coverage.reportextension`, i.e.  
+
+`--format cobertura --targetDir .\Report` will create `.\Report\coverage.xml`  
+`--format cobertura --targetDir .\Report\MergedCoverage.xml` will create `.\Report\MergedCoverage.xml`  
+`--format cobertura --targetDir .\Report\MergedCoverage` will create `.\Report\MergedCoverage\coverage.xml`
+
+# coverlet validate 
+
+This command do validation on `json` coverlet format, because the structure is internal and we can support a lot of control without replicate alg for every supported format
+
+Validate could support in future different type of validations, for now we support threshold.
+
+```
+coverlet validate --type threshold --... --Fail or --Warn
+```
+`threshold` will support all curren switch plus a pair of new optional switch `--Fail` and `--Warn`.
+If not specified default will be `--Fail`. This allow a user to specify only `--Warn` and avoid CI failure but only a warning in console. This feat was requested by one user and seem useful.
+
+# coverlet test
+
+Coverlet test should substitute `dotnet test` but applying collector coverage on behalf of user. We'll create runsettings and run '`dotnet test --collect "XPlat Code Coverage" --settings tmp\runsettings`'
+
+```
+coverlet test solution.sln --results-directory .\Results --settings runSetting
+```
+We should run
+```
+dotnet test --collect "XPlat Code Coverage" solution.sln --results-directory .\Results --settings [runSetting+tmp\runsettings]
+```
+The example used is the worste one, if a user has got a custom settings we should merge with our.
+
+With this command we can support merge/check threshold directly in same run when process returns.
+```
+coverlet test solution.sln --format cobertura --threshold 80 --threshold-type line --output .\Report
+```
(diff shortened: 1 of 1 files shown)


Documentation/Design/.NET Tool v2.md
```diff
@@ -0,0 +1,71 @@
...
+
+# coverlet mergereports
+
+Merge `json` default coverlet reports because we should otherwise implement merge for every supported format a huge effort for first version and also with the hope to integrate this feature directly inside collectors.
+
+We'll add an alias to `json` format called `coverlet` to distiguish from possible other future json format and also because this report make sense only inside specific test lifecycle.
+The content of this report is releated to internal structure and was not born to be "portable" or "stable" over time.
+
+```
+coverlet mergereports --coverletreport .\**\coverlet.json --targetdir .\Report\coverlet.xml --format cobertura
```
petli: I think it's better that the filenames are command line arguments, rather than having to specify the `--coverletreport` flag. Also rely on bash filename expansion primarily, but perhaps have support for `**` globbing in the tool.

Documentation/Design/.NET Tool v2.md
```diff
@@ -0,0 +1,71 @@
...
+`--format cobertura --targetDir .\Report\MergedCoverage` will create `.\Report\MergedCoverage\coverage.xml`
+
+# coverlet validate 
+
+This command do validation on `json` coverlet format, because the structure is internal and we can support a lot of control without replicate alg for every supported format
+
+Validate could support in future different type of validations, for now we support threshold.
+
+```
+coverlet validate --type threshold --... --Fail or --Warn
```
petli: This need to specify which coverlet JSON files to process, which could be command line arguments in the same way as for mergereports. If multiple files are listed (or globbed) they should be merged before calculating the validation (but without the merge result being saved) to avoid having to first run a merge-to-coverlet followed by validate.

Documentation/Design/.NET Tool v2.md
```diff
@@ -0,0 +1,71 @@
...
+Tracking issue https://github.com/tonerdo/coverlet/issues/683
+
+The idea is to support some new verbs:
+
+# coverlet run
+
+This will support current instrumentation workflow
+
+```
+coverlet run /path/to/test-assembly.dll --target "dotnet" --targetargs "test /path/to/test-project --no-build"
```
petli: Suggestion: also add `--coverlet-batch ID` and save the ID in the JSON coverage file. Then the same `--coverlet-batch ID` can be provided to `mergereports` and `validate` to select which files to include, if the result directory isn't cleared between runs.

Documentation/Design/.NET Tool v2.md
```diff
@@ -0,0 +1,71 @@
...
+```
+dotnet test --collect "XPlat Code Coverage" solution.sln --results-directory .\Results --settings [runSetting+tmp\runsettings]
+```
+The example used is the worste one, if a user has got a custom settings we should merge with our.
+
+With this command we can support merge/check threshold directly in same run when process returns.
+```
+coverlet test solution.sln --format cobertura --threshold 80 --threshold-type line --output .\Report
+```
+At the moment it's not super clear in my mind how to find and merge report files, especially if user specify `--results-directory`. Also because a user can specify also every other `dotnet test` driver switch.
```
petli: A batch ID can be applied to each individual run (see `--coverlet-batch` suggestion above) and use that to find out which files were produced in a specific run.

Documentation/Design/.NET Tool v2.md
MarcoRossignoli: > I think it's better that the filenames are command line arguments,

Can you explain better, it's not clear to me....can you type a command sample?

Documentation/Design/.NET Tool v2.md
MarcoRossignoli: yep you're right good catch...we could merge in memory and do the check

Documentation/Design/.NET Tool v2.md
MarcoRossignoli: My ideas was not "touch" old feature, to avoid new behaviour for current user...I mean we could add new feature to current .net tool, but it wasn't a goal of this tool.
Again...this run tool today will output a formant different than our json...so where we put that id?

Documentation/Design/.NET Tool v2.md
MarcoRossignoli: Also in this case the resulting file is not a coverlet format...I mean no always depends on `--format` parameter

tonerdo:
Late to the party! Looking this over

MarcoRossignoli:
@tonerdo @petli I'm a bit blocked at the moment on improvements because I need some double check on my ideas to be sure that we've the same goal in mind.
Next version will have only fixes.
I know that this is OSS and we spent on it extra time when we have it.

Documentation/Design/.NET Tool v2.md
MarcoRossignoli: >  use that to find out which files were produced in a specific run.

This is interesting...in general your comment concern about repetitive runs, we could add a pair of more flat to command like `--cleansource` and '--appendtimestamp yyyMMddhhmmss[default]' '--cleantarget' to cleanup source target and add a timestamp to final generated file, for easy re-runs

What do you think?

petli: I'm assuming this would translate into running each individual test project, producing coverlet json files, and then merging them in the end once they all have run. Then having a batch ID in the coverlet files (not the merged file in cobertura or whatever format) would make it easy for this command to pick up the files produced by running the tests in the SLN, without the user having to think about this possibly picking up results from older runs.

Documentation/Design/.NET Tool v2.md
petli: coverlet mergereports --targetdir .\Report\coverlet.xml --format cobertura .\**\coverlet.json previously-merged.json ...

Documentation/Design/.NET Tool v2.md
petli: My comment was assuming the output here is coverlet json. Having this flag, which only would work if also selecting that format, would allow people to build a script that runs a series of specific test projects and then merges/validates the results based on the batch ID. (Though, in such a script, it's not too difficult to clean out the result dir or work in a temporary dir to ensure that only current report files are included)

MarcoRossignoli:
After some discussion with @tonerdo we prefer for now work on current drivers and allow merge and other functionality without external tool to preserve a good usability.

PR #720, opened 2020-02-02
Add ThresholdAct option to MSBuild task

roostaamir:
- This will enable the option to choose if the build should fail when
  the coverage is below the threshold or just give a warning about it

### DIFF
--- a/Documentation/MSBuildIntegration.md
+++ b/Documentation/MSBuildIntegration.md
@@ -105,12 +105,12 @@ The following command will compare the threshold value with the overall total co
 dotnet test /p:CollectCoverage=true /p:Threshold=80 /p:ThresholdType=line /p:ThresholdStat=total
 ```
 
-You can also specify what action to take when the the coverage is below the threshold value using the `/p:ThresholdAct` option. the accepted values are:
+You can also specify what action to take when the the coverage is below the threshold value using the `/p:ThresholdAct` option. The accepted values are:
 
 * `fail`: which fails the build and is the default option
-* `warning`: which will not stop the build and only gives warning
+* `warning`: which will not stop the build and only gives a warning console message
 
-The following will give give a warning (but let's the build to continue) when the threshold value is below 80:
+The following will give a warning (but let's the build to continue) when the threshold value is below 80:
 
 ```bash
 dotnet test /p:CollectCoverage=true /p:Threshold=80 /p:ThresholdAct=warning
--- a/Documentation/MSBuildIntegration.md
+++ b/Documentation/MSBuildIntegration.md
@@ -105,6 +105,17 @@ The following command will compare the threshold value with the overall total co
 dotnet test /p:CollectCoverage=true /p:Threshold=80 /p:ThresholdType=line /p:ThresholdStat=total
 ```
 
+You can also specify what action to take when the the coverage is below the threshold value using the `/p:ThresholdAct` option. the accepted values are:
+
+* `fail`: which fails the build and is the default option
+* `warning`: which will not stop the build and only gives warning
+
+The following will give give a warning (but let's the build to continue) when the threshold value is below 80:
+
+```bash
+dotnet test /p:CollectCoverage=true /p:Threshold=80 /p:ThresholdAct=warning
+```
+
 ## Excluding From Coverage
 
 ### Attributes
--- a/src/coverlet.console/Program.cs
+++ b/src/coverlet.console/Program.cs
@@ -205,10 +205,10 @@ namespace Coverlet.Console
                 {
                     exitCode += (int)CommandExitCodes.TestFailed;
                 }
+
                 thresholdTypeFlags = result.GetThresholdTypesBelowThreshold(summary, dThreshold, thresholdTypeFlags, dThresholdStat);
                 if (thresholdTypeFlags != ThresholdTypeFlags.None)
                 {
-                    exitCode += (int)CommandExitCodes.CoverageBelowThreshold;
                     var exceptionMessageBuilder = new StringBuilder();
                     if ((thresholdTypeFlags & ThresholdTypeFlags.Line) != ThresholdTypeFlags.None)
                     {
@@ -225,12 +225,16 @@ namespace Coverlet.Console
                         exceptionMessageBuilder.AppendLine($"The {dThresholdStat.ToString().ToLower()} method coverage is below the specified {dThreshold}");
                     }
 
-                    if (dThresholdAct == ThresholdAction.Fail)
+                    switch (dThresholdAct)
                     {
-                        throw new Exception(exceptionMessageBuilder.ToString());
+                        case ThresholdAction.Warning:
+                            logger.LogWarning(exceptionMessageBuilder.ToString());
+                            break;
+                        case ThresholdAction.Fail:
+                        default:
+                            exitCode += (int)CommandExitCodes.CoverageBelowThreshold;
+                            throw new Exception(exceptionMessageBuilder.ToString());
                     }
-
-                    logger.LogWarning(exceptionMessageBuilder.ToString());
                 }
 
                 return exitCode;
--- a/src/coverlet.console/Program.cs
+++ b/src/coverlet.console/Program.cs
@@ -37,6 +37,7 @@ namespace Coverlet.Console
             CommandOption threshold = app.Option("--threshold", "Exits with error if the coverage % is below value.", CommandOptionType.SingleValue);
             CommandOption thresholdTypes = app.Option("--threshold-type", "Coverage type to apply the threshold to.", CommandOptionType.MultipleValue);
             CommandOption thresholdStat = app.Option("--threshold-stat", "Coverage statistic used to enforce the threshold value.", CommandOptionType.SingleValue);
+            CommandOption thresholdAct = app.Option("--threshold-act", "The action to take when coverage is below the threshold value. Defaults to fail the build", CommandOptionType.SingleValue);
             CommandOption excludeFilters = app.Option("--exclude", "Filter expressions to exclude specific modules and types.", CommandOptionType.MultipleValue);
             CommandOption includeFilters = app.Option("--include", "Filter expressions to include only specific modules and types.", CommandOptionType.MultipleValue);
             CommandOption excludedSourceFiles = app.Option("--exclude-by-file", "Glob patterns specifying source files to exclude.", CommandOptionType.MultipleValue);
@@ -105,6 +106,7 @@ namespace Coverlet.Console
                 var dThreshold = threshold.HasValue() ? double.Parse(threshold.Value()) : 0;
                 var dThresholdTypes = thresholdTypes.HasValue() ? thresholdTypes.Values : new List<string>(new string[] { "line", "branch", "method" });
                 var dThresholdStat = thresholdStat.HasValue() ? Enum.Parse<ThresholdStatistic>(thresholdStat.Value(), true) : Enum.Parse<ThresholdStatistic>("minimum", true);
+                var dThresholdAct = thresholdAct.HasValue() ? Enum.Parse<ThresholdAction>(thresholdAct.Value(), true) : Enum.Parse<ThresholdAction>("fail", true);
 
                 logger.LogInformation("\nCalculating coverage result...");
 
@@ -223,7 +225,12 @@ namespace Coverlet.Console
                         exceptionMessageBuilder.AppendLine($"The {dThresholdStat.ToString().ToLower()} method coverage is below the specified {dThreshold}");
                     }
 
-                    throw new Exception(exceptionMessageBuilder.ToString());
+                    if (dThresholdAct == ThresholdAction.Fail)
+                    {
+                        throw new Exception(exceptionMessageBuilder.ToString());
+                    }
+
+                    logger.LogWarning(exceptionMessageBuilder.ToString());
                 }
 
                 return exitCode;
--- a/src/coverlet.msbuild.tasks/CoverageResultTask.cs
+++ b/src/coverlet.msbuild.tasks/CoverageResultTask.cs
@@ -20,6 +20,7 @@ namespace Coverlet.MSbuild.Tasks
         private double _threshold;
         private string _thresholdType;
         private string _thresholdStat;
+        private string _thresholdAct;
         private string _coverletMultiTargetFrameworksCurrentTFM;
         private ITaskItem _instrumenterState;
         private MSBuildLogger _logger;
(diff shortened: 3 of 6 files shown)


MarcoRossignoli:
I'll check asap, I'm sick at the moment :face_with_thermometer:

Documentation/MSBuildIntegration.md
```diff
@@ -105,6 +105,17 @@ The following command will compare the threshold value with the overall total co
 dotnet test /p:CollectCoverage=true /p:Threshold=80 /p:ThresholdType=line /p:ThresholdStat=total
 ```
 
+You can also specify what action to take when the the coverage is below the threshold value using the `/p:ThresholdAct` option. the accepted values are:
```
MarcoRossignoli:
```suggestion
You can also specify what action to take when the the coverage is below the threshold value using the `/p:ThresholdAct` option. The accepted values are:
```

Documentation/MSBuildIntegration.md
```diff
@@ -105,6 +105,17 @@ The following command will compare the threshold value with the overall total co
...
+
+* `fail`: which fails the build and is the default option
+* `warning`: which will not stop the build and only gives warning
```
MarcoRossignoli:
```suggestion
* `warning`: which will not stop the build and only gives warning console message
```

Documentation/MSBuildIntegration.md
```diff
@@ -105,6 +105,17 @@ The following command will compare the threshold value with the overall total co
...
+
+The following will give give a warning (but let's the build to continue) when the threshold value is below 80:
```
MarcoRossignoli:
```suggestion
The following will give a warning (but let's the build to continue) when the threshold value is below 80:
```

src/coverlet.console/Program.cs
```diff
@@ -223,7 +225,12 @@ static int Main(string[] args)
...
-                    throw new Exception(exceptionMessageBuilder.ToString());
+                    if (dThresholdAct == ThresholdAction.Fail)
+                    {
+                        throw new Exception(exceptionMessageBuilder.ToString());
+                    }
+
+                    logger.LogWarning(exceptionMessageBuilder.ToString());
                 }
 
                 return exitCode;
```
MarcoRossignoli: Are we sure that build doesn't fail in case of warning...returns a value different than 0, isn't this an error return value for a process?

src/coverlet.console/Program.cs
```diff
@@ -223,7 +225,12 @@ static int Main(string[] args)
                         exceptionMessageBuilder.AppendLine($"The {dThresholdStat.ToString().ToLower()} method coverage is below the specified {dThreshold}");
                     }
 
-                    throw new Exception(exceptionMessageBuilder.ToString());
+                    if (dThresholdAct == ThresholdAction.Fail)
```
MarcoRossignoli: I prefer another `else if (dThresholdAct == ThresholdAction.Warning)` or a switch, today we've 2 condition and should not change...but I want to be sure that semantic won't change if we add other state in future.

src/coverlet.msbuild.tasks/CoverageResultTask.cs
```diff
@@ -230,7 +248,12 @@ public override bool Execute()
                         exceptionMessageBuilder.AppendLine($"The {thresholdStat.ToString().ToLower()} method coverage is below the specified {_threshold}");
                     }
 
-                    throw new Exception(exceptionMessageBuilder.ToString());
+                    if(thresholdAct == ThresholdAction.Fail)
```
MarcoRossignoli: I prefer another else if (dThresholdAct == ThresholdAction.Warning) or a switch, today we've 2 condition and should not change...but I want to be sure that semantic won't change if we add other states in future.

MarcoRossignoli:
General idea is ok to me some question on code and some other notes:

* We should update also .NET tool guide https://github.com/tonerdo/coverlet/blob/master/Documentation/GlobalTool.md

* We should add integration testing for console/msbuild drivers with new switch https://github.com/tonerdo/coverlet/tree/master/test/coverlet.integration.tests  
To simulate coverage fail you can pass a xunit test filter to skip the only present test inside template dll https://github.com/microsoft/vstest-docs/blob/master/docs/filter.md

src/coverlet.console/Program.cs
roostaamir: Nice catch. I made it so that it only changes the exit code when it should fail

Documentation/MSBuildIntegration.md
roostaamir: Fixed

Documentation/MSBuildIntegration.md
roostaamir: Fixed

Documentation/MSBuildIntegration.md
roostaamir: Fixed

src/coverlet.console/Program.cs
roostaamir: changed to using a switch statement

src/coverlet.msbuild.tasks/CoverageResultTask.cs
roostaamir: changed to using a switch statement

roostaamir:
> General idea is ok to me some question on code and some other notes:
> 
> * We should update also .NET tool guide https://github.com/tonerdo/coverlet/blob/master/Documentation/GlobalTool.md
> * We should add integration testing for console/msbuild drivers with new switch https://github.com/tonerdo/coverlet/tree/master/test/coverlet.integration.tests
>   To simulate coverage fail you can pass a xunit test filter to skip the only present test inside template dll https://github.com/microsoft/vstest-docs/blob/master/docs/filter.md

I'm glad you like the idea. I will add the tests then

src/coverlet.console/Program.cs
```diff
@@ -223,7 +225,16 @@ static int Main(string[] args)
...
 
-                    throw new Exception(exceptionMessageBuilder.ToString());
+                    switch (dThresholdAct)
+                    {
+                        case ThresholdAction.Warning:
+                            logger.LogWarning(exceptionMessageBuilder.ToString());
+                            break;
+                        case ThresholdAction.Fail:
+                        default:
+                            exitCode += (int)CommandExitCodes.CoverageBelowThreshold;
```
MarcoRossignoli: I prefer a default with an error like `throw new ArgumentException("Unhandled ThresholdAction")` or something like that, than default `Fail` if we add a new enum and we don't handle it we should catch error immediatly, some test could get false positive(fails but not for right reason).

roostaamir:
There's this really weird behavior that the test framework is giving me.
Take the following as an example:

`[Fact]
        public void Test_CoverageThreshold_Fail_NoThresholdActionDefined()
        {
            using ClonedTemplateProject clonedTemplateProject = PrepareTemplateProject();
            string[] targetFrameworks = new string[] { "netcoreapp2.2", "netcoreapp2.1" };
            UpdateProjectTargetFramework(clonedTemplateProject, targetFrameworks);
            Assert.True(DotnetCli($"test \"{clonedTemplateProject.ProjectRootPath}\" /p:CollectCoverage=true /p:Include=\"[{ClonedTemplateProject.AssemblyName}]*DeepThought\" /p:IncludeTestAssembly=true /p:Threshold=80 /p:CoverletOutput=\"{clonedTemplateProject.ProjectRootPath}\"\\file.ext1.ext2", out string standardOutput, out string standardError, clonedTemplateProject.ProjectRootPath!), standardOutput);
        }`

notice that I added the threshold as a parameter. If you run this code, everything is ok because by default the threshold is 100% which is above 80 and the test passes.
But if you comment out the `TestTemplate` like the following:

`public class TemplateTest
    {
        [Fact]
        public void Answer()
        {
            //DeepThought dt = new DeepThought();
            //Assert.Equal(42, dt.AnswerToTheUltimateQuestionOfLifeTheUniverseAndEverything());
        }
    }
`

Now the build wants to fail because of the threshold failing. In this scenario, the command `commandProcess.WaitForExit` times out and the test will give me this error:

`System.IO.IOException : The process cannot access the file '\\?\C:\Projects\Personal\_Github\coverlet\test\coverlet.integration.tests\bin\Debug\netcoreapp3.1\13359b5bf95c433d9ccf335f85f421fc' because it is being used by another process.`

The same thing happens if I omit the `TestTemplate` file using the `--filter` argument. Do you have any insight on this?

MarcoRossignoli:
Can you post all stack trace or all console output?With the only error is not easy understand.
```
System.IO.IOException : The process cannot access the file '\\?\C:\Projects\Personal\_Github\coverlet\test\coverlet.integration.tests\bin\Debug\netcoreapp3.1\13359b5bf95c433d9ccf335f85f421fc' because it is being used by another process.
```

roostaamir:
This is the best stack trace I could get out of the test:
`
Message: 
    System.IO.IOException : The process cannot access the file '\\?\C:\Projects\Personal\_Github\coverlet\test\coverlet.integration.tests\bin\Debug\netcoreapp3.1\30b7de57646d476dbf276051647f21dc' because it is being used by another process.
  Stack Trace: 
    FileSystem.RemoveDirectoryInternal(String fullPath, Boolean topLevel, Boolean allowDirectoryNotEmpty)
    FileSystem.RemoveDirectoryRecursive(String fullPath, WIN32_FIND_DATA& findData, Boolean topLevel)
    FileSystem.RemoveDirectory(String fullPath, Boolean recursive)
    Directory.Delete(String path, Boolean recursive)
    ClonedTemplateProject.Dispose() line 340
    Msbuild.Test_CoverageThreshold_Fail_NoThresholdActionDefined() line 213
`

It appears that the dispose function is unable to delete the file it created but only when the coverage threshold is not satisfied

MarcoRossignoli:
I think you're not aligned with master...I trapped that benign exception https://github.com/tonerdo/coverlet/blob/master/test/coverlet.integration.tests/BaseTest.cs#L334

roostaamir:
Ok, so there is the problem
The part you said about that benign exception helped but not in the way I was expecting :)

I was aligned with master but the problem was that the exception you mentioned was actually of type `System.IO` not the `UnauthorizedAccessException` that it was supposed to handle. So changing the code to something like this:

`
        public void Dispose()
        {
            if (CleanupOnDispose)
            {
                try
                {
                    Directory.Delete(ProjectRootPath, true);
                }
                catch (UnauthorizedAccessException)
                {
                    // Sometimes on CI AzDo we get Access Denied on delete
                    // swallowed exception to not waste time
                }
                catch (IOException ex)
                {
                    // Sometimes on CI AzDo we get Access Denied on delete
                    // swallowed exception to not waste time
                }
            }
        }
`

Dose not throw the exception related to disposing the class but now it will actually complain about the timeout issue that I told you in the previous comment:


`Command 'dotnet test "C:\Projects\Personal\_Github\coverlet\test\coverlet.integration.tests\bin\Debug\netcoreapp3.1\3b034a125e3c45cba8d11b021320e464" /p:CollectCoverage=true /p:Include="[coverletsamplelib.integration.template]*DeepThought" /p:IncludeTestAssembly=true /p:Threshold=80 /p:CoverletOutput="C:\Projects\Personal\_Github\coverlet\test\coverlet.integration.tests\bin\Debug\netcoreapp3.1\3b034a125e3c45cba8d11b021320e464"\file.ext1.ext2' didn't end after 5 minute`

which will actually happen here:
https://github.com/tonerdo/coverlet/blob/master/test/coverlet.integration.tests/BaseTest.cs#L96

If you remove the `/p:Threshold=80` parameter, the issue will be gone and the timeout will not happen. If I un-comment the `Answer` method (which will cause the threshold to be 100%) and keep the threshold parameter it will also pass without errors. I also tested the exact command in the command line and it works!

So to recap, I think the `RunCommand` method will timeout when the coverage is below the threshold and it only happens through the test methods, not in a real scenario.

Here's the stack trace: (not reveling much though)

`
Stack Trace: 
    BaseTest.RunCommand(String command, String arguments, String& standardOutput, String& standardError, String workingDirectory) line 100

    BaseTest.DotnetCli(String arguments, String& standardOutput, String& standardError, String workingDirectory) line 107

    Msbuild.Test_CoverageThreshold_Fail_NoThresholdActionDefined() line 211

`

roostaamir:
Update:
Based on some research I did, sometimes calling
`standardOutput = commandProcess.StandardOutput.ReadToEnd();`
before
`commandProcess.WaitForExit()` 
causes a deadlock. So changing the `RunCommand` code to the following solved the timeout problem:
`

            Debug.WriteLine($"BaseTest.RunCommand: {command} {arguments}");
            ProcessStartInfo psi = new ProcessStartInfo(command, arguments);
            psi.WorkingDirectory = workingDirectory;
            psi.RedirectStandardError = true;
            psi.RedirectStandardOutput = true;
            Process commandProcess = Process.Start(psi);
            standardOutput = commandProcess.StandardOutput.ReadToEnd();
            standardError = commandProcess.StandardError.ReadToEnd(); // changed the order of commands here
            if (!commandProcess.WaitForExit((int)TimeSpan.FromMinutes(5).TotalMilliseconds))
            {
                throw new XunitException($"Command 'dotnet {arguments}' didn't end after 5 minute");
            }
            return commandProcess.ExitCode == 0;
`

Now a new problem arises :)

Apparently code coverage being less than the threshold won't cause an error code other than 0 (but you can actually see the erros in the 'standardOutput` variable. I suspect that the problem is arising from here:
https://github.com/tonerdo/coverlet/blob/master/src/coverlet.console/Program.cs#L226

it is changing the return code but throwing here will cause it not the return the modified code.
However I am unable to debug this correctly since I cannot set breakpoints in the code because they will not trigger. Any ideas about this and how it can be debugged?

PS: Is it okay if I make another pull request about the timeout issue?

MarcoRossignoli:
> PS: Is it okay if I make another pull request about the timeout issue?

Yes, please add a comment with information links above the updated row, so in future we'll understand why that un-natural ordering.

> it is changing the return code but throwing here will cause it not the return the modified code.
However I am unable to debug this correctly since I cannot set breakpoints in the code because they will not trigger. Any ideas about this and how it can be debugged?

You can setup .net tool project(it's a console app) and pass arguments and it will run like live running, .net tool is a console app, so after child process return(dotnet test) you should be able to go on with debug.

MarcoRossignoli:
@roostaamir any news?

roostaamir-knowit:
I have been really busy this past weeks. I will start working on it again this weekend

MarcoRossignoli:
@roostaamir no worries at your pace, only asked to know if I would close for the moment. No worries at all, this is OSS.

tonerdo:
Since there are just two options, I think a switch will be better. It'll be `false` (will fail the build) by default and can simply be set to `true` (will issue a warning) by the user.

We could call it `ContinueOnFailure` (like how it is in CIs) and we can let it apply to all things in coverlet that could fail a build (only threshold does that for now)

MarcoRossignoli:
@roostaamir I'm going to close this because very old and I didn't get any news about it. 
If you want to finish this feature feel free to re-open and continue.
Thanks

Danny-UKDM:
@roostaamir pls 🙏

Issue #748, opened 2020-02-28
MissingMethodException - SessionStartEventArgs.GetPropertyValue


YulerB:
This new issue is related to https://github.com/tonerdo/coverlet/issues/456, requested by @MarcoRossignoli

The difference is, this issue is to see if a fall-back in the code below would address the issue and allow us to use Coverlet using .net core 2.1.400. 

**Exception:**

    Data collector 'XPlat code coverage' message: System.MissingMethodException: Method not 
    found: '!!0 
    Microsoft.VisualStudio.TestPlatform.ObjectModel.DataCollection.SessionStartEventArgs.GetProperty    Value(System.String)'.
       at     Coverlet.Collector.DataCollection.CoverletCoverageCollector.GetTestModules(SessionStartEventArgs sessionStartEventArgs)
       at Coverlet.Collector.DataCollection.CoverletCoverageCollector.OnSessionStart(Object sender,     SessionStartEventArgs sessionStartEventArgs).

**Fallback:**

        private static IEnumerable<string> GetPropertyValueWrapper(SessionStartEventArgs sessionStartEventArgs)
        {
            try
            {
                return sessionStartEventArgs.GetPropertyValue<IEnumerable<string>>(CoverletConstants.TestSourcesPropertyName);
            }
            catch (MissingMethodException)
            {
                using (var enumProperties = sessionStartEventArgs.GetProperties())
                {
                    while (enumProperties.MoveNext())
                    {
                        if (enumProperties.Current.Key == CoverletConstants.TestSourcesPropertyName &&
                            enumProperties.Current.Value is IEnumerable<string> propertyValue)
                        {
                            return propertyValue;
                        }
                    }
                }
                throw;
            }
        }

YulerB:
From what I can tell, looks like test sources was added only in 2.1.7 in https://github.com/dotnet/cli/pull/11312

MarcoRossignoli:
@vagisha-nidhi does it make sense?

cc: @nohwnd

MarcoRossignoli:
Possible related PR https://github.com/microsoft/vstest/pull/1962
Spec https://github.com/Microsoft/vstest-docs/blob/master/RFCs/0026-Passing-TestPlatform-Properties-to-DataCollector.md
Cannot find in release changelog https://github.com/Microsoft/vstest-docs/blob/master/docs/releases.md

MarcoRossignoli:
Closed for stale conversation.

PR #749, opened 2020-02-28
Updating coverlet collector to use GetPropertyValue and if its missing fallback to GetProperties 

YulerB:
To test https://github.com/tonerdo/coverlet/issues/748

Fallback to get the test source property when the GetPropertyValue<T> method is not found.

If the GetProperties method is also not found, a missingMethodException will also be raised.
If the GetProperties method doesn't return the property, we rethrow the exception.

I've tested both methods return the TestSources property when run using .NET Core SDK 2.2.400.

This is a speculative fix to see if GetProperties method exists for .NET Core SDK 2.1.400.


MarcoRossignoli:
Close for now, current supported .NET Core version > 3.1 will work and we didn't have a lot of issue related to it since collectors integration, but thanks for the investigation!

Issue #736, opened 2020-02-18
Coverlet fails to write coverage with file locked error


abe545:
So this is a bit odd, but it only happens during our CI (and it is intermittent). We're using v1.2 of the vstest collector.

The error we get is this:

```
Data collector 'XPlat code coverage' message: [coverlet]Coverlet.Collector.Utilities.CoverletDataCollectorException: CoverletCoverageDataCollector: Failed to get coverage result
 ---> System.IO.IOException: The process cannot access the file '/tmp/MySourceProject_8b9f64e1-2aa6-4144-aece-e1ec4323d842' because it is being used by another process.
   at System.IO.FileStream.Init(FileMode mode, FileShare share, String originalPath)
   at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options)
   at System.IO.FileStream..ctor(String path, FileMode mode)
   at Coverlet.Core.Helpers.FileSystem.NewFileStream(String path, FileMode mode) in D:\git\coverletToRelease\src\coverlet.core\Helpers\FileSystem.cs:line 43
   at Coverlet.Core.Coverage.CalculateCoverage() in D:\git\coverletToRelease\src\coverlet.core\Coverage.cs:line 352
   at Coverlet.Core.Coverage.GetCoverageResult() in D:\git\coverletToRelease\src\coverlet.core\Coverage.cs:line 137
   at Coverlet.Collector.DataCollection.CoverageWrapper.GetCoverageResult(Coverage coverage) in D:\git\coverletToRelease\src\coverlet.collector\DataCollection\CoverageWrapper.cs:line 44
   at Coverlet.Collector.DataCollection.CoverageManager.GetCoverageResult() in D:\git\coverletToRelease\src\coverlet.collector\DataCollection\CoverageManager.cs:line 91
   --- End of inner exception stack trace ---
   at Coverlet.Collector.DataCollection.CoverageManager.GetCoverageResult() in D:\git\coverletToRelease\src\coverlet.collector\DataCollection\CoverageManager.cs:line 96
   at Coverlet.Collector.DataCollection.CoverletCoverageCollector.OnSessionEnd(Object sender, SessionEndEventArgs e) in D:\git\coverletToRelease\src\coverlet.collector\DataCollection\CoverletCoverageCollector.cs:line 148.
```

[...]

Issue #307, opened 2019-01-17
Combine sequence point hits with adjacent branch hits


sharwell:
Instrumented assemblies sometimes contain a sequence point hit immediately followed by a branch hit (or vice versa). These locations could be coalesced to avoid calling `RecordHit` twice.

PR #754, opened 2020-03-07
Issue 736 fix attempt experiment

MarcoRossignoli:
Experiment fix for issue https://github.com/tonerdo/coverlet/issues/736

The issue is related to possible race condition between testhost process exit flush and get coverage result inside datacollector process.

Today we "reflush" hit file on process exit also in case of collectors usage, if we're unlucky there is a chance that we access to hit files from two different process at same time, test host process write to it datacollector process read from it.

This is a bad design ported from msbuild/.net tool usage, we should provide a fix, we shouldn't flush on process exit if we're on collectors.

cc: @abe545

### DIFF
--- a/src/coverlet.core/Instrumentation/ModuleTrackerTemplate.cs
+++ b/src/coverlet.core/Instrumentation/ModuleTrackerTemplate.cs
@@ -21,6 +21,7 @@ namespace Coverlet.Core.Instrumentation
         public static string HitsFilePath;
         public static int[] HitsArray;
         public static bool SingleHit;
+        public static bool IsCalledByInProcessCollector;
         private static readonly bool _enableLog = int.TryParse(Environment.GetEnvironmentVariable("COVERLET_ENABLETRACKERLOG"), out int result) ? result == 1 : false;
 
         static ModuleTrackerTemplate()
@@ -35,8 +36,8 @@ namespace Coverlet.Core.Instrumentation
         // to UnloadModule will be injected in System.AppContext.OnProcessExit.
         public static void RegisterUnloadEvents()
         {
-            AppDomain.CurrentDomain.ProcessExit += new EventHandler(UnloadModule);
-            AppDomain.CurrentDomain.DomainUnload += new EventHandler(UnloadModule);
+            AppDomain.CurrentDomain.ProcessExit += new EventHandler(ProcessExitEvent);
+            AppDomain.CurrentDomain.DomainUnload += new EventHandler(DomainUnloadEvent);
         }
 
         public static void RecordHitInCoreLibrary(int hitLocationIndex)
@@ -73,11 +74,42 @@ namespace Coverlet.Core.Instrumentation
                 location = 1;
         }
 
-        public static void UnloadModule(object sender, EventArgs e)
+        private static bool IsDotNetCore()
+        {
+            // object for .NET Framework is inside mscorlib.dll
+            return Path.GetFileName(typeof(object).Assembly.Location) == "System.Private.CoreLib.dll";
+        }
+
+        public static void ProcessExitEvent(object sender, EventArgs e)
+        {
+            if (IsCalledByInProcessCollector && IsDotNetCore())
+            {
+                return;
+            }
+
+            Flush(nameof(ProcessExitEvent));
+        }
+
+        public static void DomainUnloadEvent(object sender, EventArgs e)
+        {
+            Flush(nameof(DomainUnloadEvent));
+        }
+
+        public static void AppContextOnProcessExitEvent(object sender, EventArgs e)
+        {
+            Flush(nameof(AppContextOnProcessExitEvent));
+        }
+
+        public static void InProcessCollectorFlush()
+        {
+            Flush(nameof(InProcessCollectorFlush));
+        }
+
+        private static void Flush(string flushType)
         {
             try
             {
-                WriteLog($"Unload called for '{Assembly.GetExecutingAssembly().Location}'");
+                WriteLog($"Unload called for '{Assembly.GetExecutingAssembly().Location}' FlushType: {flushType}");
                 // Claim the current hits array and reset it to prevent double-counting scenarios.
                 int[] hitsArray = Interlocked.Exchange(ref HitsArray, new int[HitsArray.Length]);
 
--- a/src/coverlet.collector/DataCollection/CoverageWrapper.cs
+++ b/src/coverlet.collector/DataCollection/CoverageWrapper.cs
@@ -31,7 +31,8 @@ namespace Coverlet.Collector.DataCollection
                 settings.UseSourceLink,
                 coverletLogger,
                 DependencyInjection.Current.GetService<IInstrumentationHelper>(),
-                DependencyInjection.Current.GetService<IFileSystem>());
+                DependencyInjection.Current.GetService<IFileSystem>(),
+                true);
         }
 
         /// <summary>
--- a/src/coverlet.collector/InProcDataCollection/CoverletInProcDataCollector.cs
+++ b/src/coverlet.collector/InProcDataCollection/CoverletInProcDataCollector.cs
@@ -53,10 +53,10 @@ namespace Coverlet.Collector.DataCollection
 
                 try
                 {
-                    _eqtTrace.Verbose($"Calling ModuleTrackerTemplate.UnloadModule for '{injectedInstrumentationClass.Assembly.FullName}'");
-                    var unloadModule = injectedInstrumentationClass.GetMethod(nameof(ModuleTrackerTemplate.UnloadModule), new[] { typeof(object), typeof(EventArgs) });
-                    unloadModule.Invoke(null, new[] { null, EventArgs.Empty });
-                    _eqtTrace.Verbose($"Called ModuleTrackerTemplate.UnloadModule for '{injectedInstrumentationClass.Assembly.FullName}'");
+                    _eqtTrace.Verbose($"Calling ModuleTrackerTemplate.InProcessCollectorFlush for '{injectedInstrumentationClass.Assembly.FullName}'");
+                    var unloadModule = injectedInstrumentationClass.GetMethod(nameof(ModuleTrackerTemplate.InProcessCollectorFlush));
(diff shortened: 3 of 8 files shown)


src/coverlet.core/Instrumentation/ModuleTrackerTemplate.cs
```diff
@@ -73,11 +74,42 @@ public static void RecordSingleHit(int hitLocationIndex)
...
-        public static void UnloadModule(object sender, EventArgs e)
+        private static bool IsDotNetCore()
+        {
+            // object for .NET Framework is inside mscorlib.dll
+            return Path.GetFileName(typeof(object).Assembly.Location) == "System.Private.CoreLib.dll";
+        }
+
+        public static void ProcessExitEvent(object sender, EventArgs e)
+        {
+            if (IsCalledByInProcessCollector && IsDotNetCore())
```
abe545: At least for my test projects, moving the call to `IsDotNetCore` to https://github.com/tonerdo/coverlet/pull/754/files#diff-164040e68c2b496a18449c2363998152L111 still solves my issue - and should address your concern about people who don't update their test sdk (https://github.com/tonerdo/coverlet/issues/736#issuecomment-592154536).

abe545: Basically, we could remove this whole if block, and replace 
`if (failedToCreateNewHitsFile)` with `if (failedToCreateNewHitsFile && !IsDotNetCore())`
It might be a safer change?

MarcoRossignoli: > At least for my test projects, moving the call to IsDotNetCore to https://github.com/tonerdo/coverlet/pull/754/files#diff-164040e68c2b496a18449c2363998152L111 still solves my issue - and should address your concern about people who don't update their test sdk (#736 (comment)).

Cannot understand where this link pointing...can you rewrite it?

MarcoRossignoli: > if (failedToCreateNewHitsFile) with if (failedToCreateNewHitsFile && !IsDotNetCore())

This could work...I'm only scared about remove redundant "flush" on exit(I want to do it for a long time) because could hidden some latent bug.
I'll take a look asap, busy weeks for me.
Thank's for investigation.

abe545: > > At least for my test projects, moving the call to IsDotNetCore to https://github.com/tonerdo/coverlet/pull/754/files#diff-164040e68c2b496a18449c2363998152L111 still solves my issue - and should address your concern about people who don't update their test sdk (#736 (comment)).
> 
> Cannot understand where this link pointing...can you rewrite it?

Updated it to point at the file outside this pr

MarcoRossignoli: Understood now.

abe545: The other thing I noticed is that the process that updates the hits file is pretty inefficient (seeking back and forth in the file is slow from what I understand). I was able generate coverage successfully after rewriting it as follows (this could probably be further optimized by reading/writing more than one `int` at a time):
```cs
        private static async System.Threading.Tasks.Task UpdateHits(int[] hitsArray)
        {
            var readBuffer = new byte[sizeof(int)];
            var writeBuffer = new byte[sizeof(int)];

            using (var readStream = new FileStream(HitsFilePath, FileMode.Open, FileAccess.ReadWrite, FileShare.ReadWrite))
            using (var writeStream = new FileStream(HitsFilePath, FileMode.Open, FileAccess.ReadWrite, FileShare.ReadWrite))
            {
                int hitsLength = await ReadNext();
                var readTask = ReadNext();
                WriteLog($"Current hits found '{hitsLength}'");

                if (hitsLength != hitsArray.Length)
                {
                    throw new InvalidOperationException(
                        $"{HitsFilePath} has {hitsLength} entries but on memory {nameof(HitsArray)} has {hitsArray.Length}");
                }
                writeStream.Seek(sizeof(int), SeekOrigin.Current);
                for (int i = 0; i < hitsLength; ++i)
                {
                    int hits = await readTask;
                    readTask = ReadNext(); // start reading the next hits before we wait to write

                    if (SingleHit)
                        hits = hitsArray[i] + hits > 0 ? 1 : 0;
                    else
                        hits = hitsArray[i] + hits;
                    
                    await Write(hits);
                }

                async System.Threading.Tasks.Task<int> ReadNext()
                {
                    var bytesRead = await readStream.ReadAsync(readBuffer, 0, sizeof(int));
                    if (bytesRead < sizeof(int))
                    {
                        return -1;
                    }

                    return BitConverter.ToInt32(readBuffer, 0);
                }

                async System.Threading.Tasks.Task Write(int hits)
                {
                    writeBuffer[0] = (byte)hits;
                    writeBuffer[1] = (byte)(hits >> 8);
                    writeBuffer[2] = (byte)(hits >> 16);
                    writeBuffer[3] = (byte)(hits >> 24);
                    await writeStream.WriteAsync(writeBuffer, 0, sizeof(int));
                }
            }
        }
```

MarcoRossignoli: For perf you should open a new issue and we'll tag with proper label, also every perf update must be supported by perf testing in test project https://github.com/tonerdo/coverlet/tree/master/test/coverlet.core.performancetest(we should use BDN)

abe545: I was just trying to offer an alternative to not writing the hits file on process exit 😄 - if you'd like me to submit a separate PR with this change instead, I can do so.

MarcoRossignoli: yep yep I'm ok for help and update...but we need to be very very careful coverlet is used by a lot of people and I try to fix bug and add useful feat before perf and to be honest write file is not a great concern at the moment.
This issue should have more impact but it's very challenging https://github.com/tonerdo/coverlet/issues/307

https://nugettrends.com/packages?months=12&ids=coverlet.msbuild&ids=coverlet.collector&ids=coverlet.console

abe545: My only reason for looking at it is that when the file is not modified, or it is written quickly, then I'm able to successfully generate coverage. When it is slow, my workflow will often fail to generate coverage. I agree that your cautious approach is the correct one, which is also my motivation for thinking of alternatives to the initial approach of not attaching the process exit event handler.
And I do want to make sure you know how much I appreciate your help with this issue! I've tested all the solutions we've discussed with my real project that currently can't generate coverage, and am happy to contribute in a way that makes the most sense for the coverlet repo (even if that is just waiting for you to have a chance to test in the coming weeks).

MarcoRossignoli: It's not clear to me why the "synch with mutex on get reports to protect hit file" doesn't work, that should be the simpler and less intrusive one.

abe545: That is an interesting one - iirc the [repro repo](https://github.com/namely/CoverletErrorRepro) I gave you will exhibit the failure to generate coverage with that approach. Might make it easier to test why?

MarcoRossignoli: will do asap

MarcoRossignoli: I cannot repro on my local...no exception, do you have some advice or variation that could help to repro?

abe545: Huh, that is odd. Just to make sure we're on the same page - this didn't produce an exception when running in `docker-compose`? Running in docker is essentially the only time I see the exception and/or coverage failing to generate.

abe545: And to hopefully add some clarification, I added a readme to that repo.
Let me try to pull your change with the mutex and build it in that repo (in a different branch) so I can demonstrate the issue.

abe545: Right, got it to fail intermittently using the mutex commit: https://github.com/namely/CoverletErrorRepro/tree/use-coverlet-mutex
As stated in the readme, this fails roughly 1/3 times on my laptop. I've noticed our CI environment is much slower than my laptop at running tests, so it is perhaps exacerbating the problem.

MarcoRossignoli: Thank's for sample and for correct exception, that fix doesn't work my fault...synch is working but if thread acquired mutex inside tracker is killed wait on get coverage will fail for abandoned mutex and also if we catch that exception we cannot be sure that file was fully written so we can hit corrupted hit file.

The unique clean solution is the one in full patch or your idea of skip re-write if we're in .net core app because if we already called flush from collector the second one should be process exits(no app domain so no intermediate flushes expected and only one file creation expected for msbuild/.net tool with .net core runtime).

abe545: No problem! Let me know if I can help with this in any way.

MarcoRossignoli: Now we've a complete analysis, I'll think about best and light solution, for now you can fork and use your idea.

MarcoRossignoli:
Close test.
I think that your idea is the best, because I found another issue, there is a issue on old vstest SDK where in-proc collector is not loaded so no flush https://github.com/microsoft/vstest/pull/2221
In this cases btw if works thank's to process exits, I mean it suffer of known issue https://github.com/tonerdo/coverlet/blob/master/Documentation/KnownIssues.md#1-vstest-stops-process-execution-earlydotnet-test but in other cases file will be persisted, better than nothing, so we should keep the "last resort" to avoid empty hits file.
So if we're on .NET Core app and we already flushed it means that someone(collectors) called flush, because here we don't have domains and we don't expect more than one flush.

PR #771, opened 2020-03-21
Bump sdk version, update tests

MarcoRossignoli:
* Bump sdk to `3.1.*`
* Use @tmds remote executor to avoid to keep in synch with arcade one. Thank's Tom for the tool!


MarcoRossignoli:
Uh seem not supported on Mac, it's a shame.

tmds:
> Uh seem not supported on Mac, it's a pity.

Do you want to help make this work? I'll help figure out how to implement it. I don't have a mac myself.

MarcoRossignoli:
>  I don't have a mac myself.

Eh the problem is that I don't have Mac too...I work on win 😞 and WSL.

tmds:
Damn you, Mac!

I've created an issue for it: https://github.com/tmds/Tmds.ExecFunction/issues/6. Maybe we can find someone willing, and able.

MarcoRossignoli:
Ok for now I'll skip that test on Mac, historically I never seen any difference for these tests between three plat. Win/Linux should be enough

PR #786, opened 2020-04-02
Fix Cake Add-In anchor in README

medianick:
There seems to be an extra "." in the anchor here that prevents it from working properly when clicked.


MarcoRossignoli:
Thank's @medianick feel free to publish the PR(at the moment is a draft).

PR #806, opened 2020-04-09
Fix nightly build versioning

MarcoRossignoli:
@clairernovotny I fixed nightly version.

### DIFF
--- a/src/coverlet.collector/version.json
+++ b/src/coverlet.collector/version.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://raw.githubusercontent.com/AArnott/Nerdbank.GitVersioning/master/src/NerdBank.GitVersioning/version.schema.json",
-  "version": "1.2.1",
+  "version": "1.3.0-preview.{height}",
   "publicReleaseRefSpec": [
     "^refs/heads/master$"
   ]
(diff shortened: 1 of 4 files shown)


MarcoRossignoli:
Need to fix integration test to find new version name.

clairernovotny:
@MarcoRossignoli you can see how Rx does integration tests here. We use a multi-stage build where artifacts are published in build and then downloaded on multiple platforms:
https://github.com/dotnet/reactive/blob/master/azure-pipelines.rx.yml

I download the artifacts into a directory and then add/update via the dotnet command:
https://github.com/dotnet/reactive/blob/master/azure-pipelines.rx.yml#L169-L175

In all cases NBGV knows what the version is.

MarcoRossignoli:
Our integration need to check coverage results https://github.com/tonerdo/coverlet/blob/master/test/coverlet.integration.tests/DotnetTool.cs#L30 or for instance https://github.com/tonerdo/coverlet/blob/master/test/coverlet.integration.tests/Collectors.cs not only run coverage.

MarcoRossignoli:
Issue related to long path https://github.com/dotnet/sdk/issues/11244

PR #813, opened 2020-04-13
Skip branches in generated `MoveNext()` for singleton iterators

### LINKED ISSUE #810: Uncovered branch in compiler-generated state machine for `yield return`
The C# compiler generates a state machine for each method that uses the `yield return` syntax. Unfortunately this state machine contains a branch that will never be visited by the execution and thus results in inaccurate branch coverage measurements.

# Repro steps

Create a minimal class library with the following code (`ClassLibrary1.cs`):

```csharp
namespace ClassLibrary1 {
    public static class Bar {
        public static System.Collections.Generic.IEnumerable<int> Foo() { yield return 1; }
    }
}
```

Create a minimal xUnit test project. Add the `coverlet.msbuild` package (2.8.1) and add the following code (`UnitTest1.cs`):

```csharp
using Xunit;

namespace XUnitTestProject1 {
    public class UnitTest1
    {
        [Fact]
        public void Test1()
        {
            Assert.Equal(1, Assert.Single(ClassLibrary1.Bar.Foo()));
        }
    }
}
```

Run coverlet:

```
PS> dotnet test /p:CollectCoverage=true /p:CoverletOutputFormat=opencover
```

## Expected

Since the method `Bar.Foo()` has no branches, the reported branch coverage should be 100%.

## Actual

```
Test run for D:\DEV\XUnitTestProject1\bin\Debug\netcoreapp3.1\XUnitTestProject1.dll(.NETCoreApp,Version=v3.1)
Microsoft (R) Test Execution Command Line Tool Version 16.5.0
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...

A total of 1 test files matched the specified pattern.

Test Run Successful.
Total tests: 1
     Passed: 1
 Total time: 1.4069 Seconds

  Generating report 'D:\DEV\XUnitTestProject1\coverage.opencover.xml'

+---------------+------+--------+--------+
| Module        | Line | Branch | Method |
+---------------+------+--------+--------+
| ClassLibrary1 | 100% | 50%    | 100%   |
+---------------+------+--------+--------+

+---------+------+--------+--------+
|         | Line | Branch | Method |
+---------+------+--------+--------+
| Total   | 100% | 50%    | 100%   |
+---------+------+--------+--------+
| Average | 100% | 50%    | 100%   |
+---------+------+--------+--------+
```

# Suspected cause

The report `coverage.opencover.xml` states that the uncovered branch is in the method `ClassLibrary1.Bar.<Foo>d__0.MoveNext()`:

```xml
<Method cyclomaticComplexity="2" nPathComplexity="0" sequenceCoverage="100" branchCoverage="50"
        isConstructor="False" isGetter="False" isSetter="False" isStatic="True">
  <Summary numSequencePoints="1" visitedSequencePoints="1" numBranchPoints="2"
           visitedBranchPoints="1" sequenceCoverage="100" branchCoverage="50"
           maxCyclomaticComplexity="2" minCyclomaticComplexity="2" visitedClasses="0"
           numClasses="0" visitedMethods="1" numMethods="1" />
  <MetadataToken />
  <Name>System.Boolean ClassLibrary1.Bar/&lt;Foo&gt;d__0::MoveNext()</Name>
```

<details><summary>full report</summary>

```xml
<?xml version="1.0" encoding="utf-8"?>
<CoverageSession>
  <Summary numSequencePoints="1" visitedSequencePoints="1" numBranchPoints="2" visitedBranchPoints="1" sequenceCoverage="100" branchCoverage="50" maxCyclomaticComplexity="2" minCyclomaticComplexity="2" visitedClasses="1" numClasses="1" visitedMethods="1" numMethods="1" />
  <Modules>
    <Module hash="0AA3E2D3-850E-44FF-A43D-06C13D27F9A8">
      <ModulePath>ClassLibrary1.dll</ModulePath>
      <ModuleTime>2020-04-12T12:08:13</ModuleTime>
      <ModuleName>ClassLibrary1</ModuleName>
      <Files>
        <File uid="1" fullPath="D:\DEV\ClassLibrary1\Class1.cs" />
      </Files>
      <Classes>
        <Class>
          <Summary numSequencePoints="1" visitedSequencePoints="1" numBranchPoints="2" visitedBranchPoints="1" sequenceCoverage="100" branchCoverage="50" maxCyclomaticComplexity="2" minCyclomaticComplexity="2" visitedClasses="1" numClasses="1" visitedMethods="1" numMethods="1" />
          <FullName>ClassLibrary1.Bar/&lt;Foo&gt;d__0</FullName>
          <Methods>
            <Method cyclomaticComplexity="2" nPathComplexity="0" sequenceCoverage="100" branchCoverage="50" isConstructor="False" isGetter="False" isSetter="False" isStatic="True">
              <Summary numSequencePoints="1" visitedSequencePoints="1" numBranchPoints="2" visitedBranchPoints="1" sequenceCoverage="100" branchCoverage="50" maxCyclomaticComplexity="2" minCyclomaticComplexity="2" visitedClasses="0" numClasses="0" visitedMethods="1" numMethods="1" />
              <MetadataToken />
              <Name>System.Boolean ClassLibrary1.Bar/&lt;Foo&gt;d__0::MoveNext()</Name>
              <FileRef uid="1" />
              <SequencePoints>
                <SequencePoint vc="3" uspid="5" ordinal="0" sl="5" sc="1" el="5" ec="2" bec="2" bev="1" fileid="1" />
              </SequencePoints>
              <BranchPoints>
                <BranchPoint vc="0" uspid="5" ordinal="0" path="0" offset="14" offsetend="22" sl="5" fileid="1" />
                <BranchPoint vc="1" uspid="5" ordinal="1" path="1" offset="14" offsetend="48" sl="5" fileid="1" />
              </BranchPoints>
              <MethodPoint vc="1" uspid="0" p8:type="SequencePoint" ordinal="0" offset="0" sc="0" sl="5" ec="1" el="5" bec="0" bev="0" fileid="1" xmlns:p8="xsi" />
            </Method>
          </Methods>
        </Class>
      </Classes>
    </Module>
  </Modules>
</CoverageSession>
```

</details>

Inspecting the `<Foo>d__0` class generated by the compiler we can see that its `MoveNext()` method has a `default` case inside the switch on the `<>1__state` variable:

```csharp
private int <>1__state;

// ...

private bool MoveNext()
{
    switch (<>1__state)
    {
    default:
        return false;
    case 0:
        <>1__state = -1;
        <>2__current = 1;
        <>1__state = 1;
        return true;
    case 1:
        <>1__state = -1;
        return false;
    }
}
```

<details><summary>full decompilation</summary>

```csharp
// ClassLibrary1.Bar.<Foo>d__0
using System;
using System.Collections;
using System.Collections.Generic;
using System.Diagnostics;
using System.Runtime.CompilerServices;

[CompilerGenerated]
private sealed class <Foo>d__0 : IEnumerable<int>, IEnumerable, IEnumerator<int>, IEnumerator, IDisposable
{
	private int <>1__state;

	private int <>2__current;

	private int <>l__initialThreadId;

	int IEnumerator<int>.Current
	{
		[DebuggerHidden]
		get
		{
			return <>2__current;
		}
	}

	object IEnumerator.Current
	{
		[DebuggerHidden]
		get
		{
			return <>2__current;
		}
	}

	[DebuggerHidden]
	public <Foo>d__0(int <>1__state)
	{
		this.<>1__state = <>1__state;
		<>l__initialThreadId = Environment.CurrentManagedThreadId;
	}

	[DebuggerHidden]
	void IDisposable.Dispose()
	{
	}

	private bool MoveNext()
	{
		switch (<>1__state)
		{
		default:
			return false;
		case 0:
			<>1__state = -1;
			<>2__current = 1;
			<>1__state = 1;
			return true;
		case 1:
			<>1__state = -1;
			return false;
		}
	}

	bool IEnumerator.MoveNext()
	{
		//ILSpy generated this explicit interface implementation from .override directive in MoveNext
		return this.MoveNext();
	}

	[DebuggerHidden]
	void IEnumerator.Reset()
	{
		throw new NotSupportedException();
	}

	[DebuggerHidden]
	IEnumerator<int> IEnumerable<int>.GetEnumerator()
	{
		if (<>1__state == -2 && <>l__initialThreadId == Environment.CurrentManagedThreadId)
		{
			<>1__state = 0;
			return this;
		}
		return new <Foo>d__0(0);
	}

	[DebuggerHidden]
	IEnumerator IEnumerable.GetEnumerator()
	{
		return ((IEnumerable<int>)this).GetEnumerator();
	}
}
```

</details>

I'm not 100% sure, but I think this `default` case is never being visited.

bert2:
This fixes #810 by skipping the two branches inside `MoveNext()` that will be generated when the iterator only `yield`s once.

For instance, the iterator...

```csharp
public class SingletonIterator
{
    public IEnumerable<string> Fetch()
    {
        yield return "one";
    }
}
```

...will result in a generated `MoveNext()` method that has no `switch` instruction, but a `brfalse.s` and a `beq.s` instruction instead:

```msil
<Fetch>d__0.MoveNext:
IL_0000:  ldarg.0     
IL_0001:  ldfld       UserQuery+SingletonIterator+<Fetch>d__0.<>1__state
IL_0006:  stloc.0     
IL_0007:  ldloc.0     
IL_0008:  brfalse.s   IL_0012
IL_000A:  br.s        IL_000C
IL_000C:  ldloc.0     
IL_000D:  ldc.i4.1    
IL_000E:  beq.s       IL_0014
IL_0010:  br.s        IL_0016
IL_0012:  br.s        IL_0018
IL_0014:  br.s        IL_0034
IL_0016:  ldc.i4.0    
IL_0017:  ret         
IL_0018:  ldarg.0     
IL_0019:  ldc.i4.m1   
IL_001A:  stfld       UserQuery+SingletonIterator+<Fetch>d__0.<>1__state
IL_001F:  nop         
IL_0020:  ldarg.0     
IL_0021:  ldstr       "one"
IL_0026:  stfld       UserQuery+SingletonIterator+<Fetch>d__0.<>2__current
IL_002B:  ldarg.0     
IL_002C:  ldc.i4.1    
IL_002D:  stfld       UserQuery+SingletonIterator+<Fetch>d__0.<>1__state
IL_0032:  ldc.i4.1    
IL_0033:  ret         
IL_0034:  ldarg.0     
IL_0035:  ldc.i4.m1   
IL_0036:  stfld       UserQuery+SingletonIterator+<Fetch>d__0.<>1__state
IL_003B:  ldc.i4.0    
IL_003C:  ret 
```

### DIFF
--- a/src/coverlet.core/Symbols/CecilSymbolHelper.cs
+++ b/src/coverlet.core/Symbols/CecilSymbolHelper.cs
@@ -412,6 +412,11 @@ namespace Coverlet.Core.Symbols
 
             bool isAsyncStateMachineMoveNext = IsMoveNextInsideAsyncStateMachine(methodDefinition);
             bool isMoveNextInsideAsyncStateMachineProlog = isAsyncStateMachineMoveNext && IsMoveNextInsideAsyncStateMachineProlog(methodDefinition);
+
+            // State machine for enumerator uses `brfalse.s`/`beq` or `switch` opcode depending on how many `yield` we have in the method body.
+            // For more than one `yield` a `switch` is emitted so we should only skip the first branch. In case of a single `yield` we need to
+            // skip the first two branches to avoid reporting a phantom branch. The first branch (`brfalse.s`) jumps to the `yield`ed value,
+            // the second one (`beq`) exits the enumeration.
             bool skipFirstBranch = IsMoveNextInsideEnumerator(methodDefinition);
             bool skipSecondBranch = false;
 
--- a/src/coverlet.core/Symbols/CecilSymbolHelper.cs
+++ b/src/coverlet.core/Symbols/CecilSymbolHelper.cs
@@ -413,7 +413,7 @@ namespace Coverlet.Core.Symbols
             bool isAsyncStateMachineMoveNext = IsMoveNextInsideAsyncStateMachine(methodDefinition);
             bool isMoveNextInsideAsyncStateMachineProlog = isAsyncStateMachineMoveNext && IsMoveNextInsideAsyncStateMachineProlog(methodDefinition);
             bool skipFirstBranch = IsMoveNextInsideEnumerator(methodDefinition);
-            bool skipSecondBranch = skipFirstBranch && instructions.All(i => i.OpCode.Code != Code.Switch);
+            bool skipSecondBranch = false;
 
             foreach (Instruction instruction in instructions.Where(instruction => instruction.OpCode.FlowControl == FlowControl.Cond_Branch))
             {
@@ -422,6 +422,7 @@ namespace Coverlet.Core.Symbols
                     if (skipFirstBranch)
                     {
                         skipFirstBranch = false;
+                        skipSecondBranch = instruction.OpCode.Code != Code.Switch;
                         continue;
                     }
 
--- a/src/coverlet.core/Symbols/CecilSymbolHelper.cs
+++ b/src/coverlet.core/Symbols/CecilSymbolHelper.cs
@@ -413,6 +413,7 @@ namespace Coverlet.Core.Symbols
             bool isAsyncStateMachineMoveNext = IsMoveNextInsideAsyncStateMachine(methodDefinition);
             bool isMoveNextInsideAsyncStateMachineProlog = isAsyncStateMachineMoveNext && IsMoveNextInsideAsyncStateMachineProlog(methodDefinition);
             bool skipFirstBranch = IsMoveNextInsideEnumerator(methodDefinition);
+            bool skipSecondBranch = skipFirstBranch && !instructions.Any(i => i.OpCode.Code == Code.Switch);
 
             foreach (Instruction instruction in instructions.Where(instruction => instruction.OpCode.FlowControl == FlowControl.Cond_Branch))
             {
@@ -421,6 +422,12 @@ namespace Coverlet.Core.Symbols
                     if (skipFirstBranch)
                     {
                         skipFirstBranch = false;
+                        continue;
+                    }
+
+                    if (skipSecondBranch)
+                    {
+                        skipSecondBranch = false;
                         continue;
                     }
 
--- a/src/coverlet.core/Symbols/CecilSymbolHelper.cs
+++ b/src/coverlet.core/Symbols/CecilSymbolHelper.cs
@@ -413,7 +413,7 @@ namespace Coverlet.Core.Symbols
             bool isAsyncStateMachineMoveNext = IsMoveNextInsideAsyncStateMachine(methodDefinition);
             bool isMoveNextInsideAsyncStateMachineProlog = isAsyncStateMachineMoveNext && IsMoveNextInsideAsyncStateMachineProlog(methodDefinition);
             bool skipFirstBranch = IsMoveNextInsideEnumerator(methodDefinition);
-            bool skipSecondBranch = skipFirstBranch && !instructions.Any(i => i.OpCode.Code == Code.Switch);
+            bool skipSecondBranch = skipFirstBranch && instructions.All(i => i.OpCode.Code != Code.Switch);
 
             foreach (Instruction instruction in instructions.Where(instruction => instruction.OpCode.FlowControl == FlowControl.Cond_Branch))
             {
--- a/test/coverlet.core.tests/Coverage/CoverageTest.Yield.cs
+++ b/test/coverlet.core.tests/Coverage/CoverageTest.Yield.cs
@@ -12,7 +12,7 @@ namespace Coverlet.Core.Tests
     {
         [ConditionalFact]
         [SkipOnOS(OS.MacOS)]
-        public void Yield_Singleton()
+        public void Yield_Single()
         {
             // We need to pass file name to remote process where it save instrumentation result
             // Similar to msbuild input/output
@@ -26,7 +26,7 @@ namespace Coverlet.Core.Tests
                     CoveragePrepareResult coveragePrepareResult = await TestInstrumentationHelper.Run<Yield>(instance =>
                     {
                         // We call method to trigger coverage hits
-                        _ = instance.One().Single();
+                        foreach(var _ in instance.One());
 
                         // For now we have only async Run helper
                         return Task.CompletedTask;
@@ -46,7 +46,8 @@ namespace Coverlet.Core.Tests
                 // Asserts on doc/lines/branches
                 result.Document("Instrumentation.Yield.cs")
                       // (line, hits)
-                      .AssertLinesCovered((9, 1));
+                      .AssertLinesCovered((9, 1))
+                      .ExpectedTotalNumberOfBranches(0);
             }
             finally
             {
@@ -66,7 +67,7 @@ namespace Coverlet.Core.Tests
                 {
                     CoveragePrepareResult coveragePrepareResult = await TestInstrumentationHelper.Run<Yield>(instance =>
                     {
-                        _ = instance.Two().ToArray();
+                        foreach (var _ in instance.Two());
                         return Task.CompletedTask;
                     }, persistPrepareResultToFile: pathSerialize[0]);
                     return 0;
@@ -74,8 +75,9 @@ namespace Coverlet.Core.Tests
 
                 CoverageResult result = TestInstrumentationHelper.GetCoverageResult(path);
 
-                result.Document("Instrumentation.SelectionStatements.cs")
-                      .AssertLinesCovered((14, 1), (15, 0));
+                result.Document("Instrumentation.Yield.cs")
+                      .AssertLinesCovered((14, 1), (15, 1))
+                      .ExpectedTotalNumberOfBranches(0);
             }
             finally
             {
--- a/test/coverlet.core.tests/Coverage/CoverageTest.Yield.cs
+++ b/test/coverlet.core.tests/Coverage/CoverageTest.Yield.cs
@@ -19,7 +19,7 @@ namespace Coverlet.Core.Tests
                 {
                     CoveragePrepareResult coveragePrepareResult = await TestInstrumentationHelper.Run<Yield>(instance =>
                     {
-                        foreach(var _ in instance.One()) ;
+                        foreach(var _ in instance.One());
 
                         return Task.CompletedTask;
                     }, persistPrepareResultToFile: pathSerialize[0]);
@@ -50,7 +50,7 @@ namespace Coverlet.Core.Tests
                 {
                     CoveragePrepareResult coveragePrepareResult = await TestInstrumentationHelper.Run<Yield>(instance =>
                     {
-                        foreach (var _ in instance.Two()) ;
+                        foreach (var _ in instance.Two());
 
                         return Task.CompletedTask;
                     }, persistPrepareResultToFile: pathSerialize[0]);
@@ -80,7 +80,7 @@ namespace Coverlet.Core.Tests
                 {
                     CoveragePrepareResult coveragePrepareResult = await TestInstrumentationHelper.Run<Yield>(instance =>
                     {
-                        foreach (var _ in instance.OneWithSwitch()) ;
+                        foreach (var _ in instance.OneWithSwitch(2)) ;
 
                         return Task.CompletedTask;
                     }, persistPrepareResultToFile: pathSerialize[0]);
@@ -92,6 +92,7 @@ namespace Coverlet.Core.Tests
 
                 result.Document("Instrumentation.Yield.cs")
                       .Method("System.Boolean Coverlet.Core.Samples.Tests.Yield/<OneWithSwitch>d__2::MoveNext()")
+                      .AssertLinesCovered((30, 1), (31, 1), (37, 1))
                       .ExpectedTotalNumberOfBranches(1);
             }
             finally
--- a/test/coverlet.core.tests/Coverage/CoverageTest.Yield.cs
+++ b/test/coverlet.core.tests/Coverage/CoverageTest.Yield.cs
@@ -152,7 +152,7 @@ namespace Coverlet.Core.Tests
 
                 result.Document("Instrumentation.Yield.cs")
                       .Method("System.Boolean Coverlet.Core.Samples.Tests.Yield/<Enumerable>d__4::MoveNext()")
-                      .AssertLinesCovered((51, 4))
+                      .AssertLinesCovered((48, 1), (49, 11), (50, 4), (51, 4), (52, 4), (53, 1))
                       .ExpectedTotalNumberOfBranches(1);
             }
             finally
--- a/test/coverlet.core.tests/Coverage/CoverageTest.Yield.cs
+++ b/test/coverlet.core.tests/Coverage/CoverageTest.Yield.cs
@@ -92,7 +92,7 @@ namespace Coverlet.Core.Tests
 
                 result.Document("Instrumentation.Yield.cs")
                       .Method("System.Boolean Coverlet.Core.Samples.Tests.Yield/<OneWithSwitch>d__2::MoveNext()")
-                      .AssertLinesCovered((30, 1), (31, 1), (37, 1))
+                      .AssertLinesCovered((21, 1), (30, 1), (31, 1), (37, 1))
                       .ExpectedTotalNumberOfBranches(1);
             }
             finally
--- a/test/coverlet.core.tests/Coverage/CoverageTest.Yield.cs
+++ b/test/coverlet.core.tests/Coverage/CoverageTest.Yield.cs
@@ -2,7 +2,6 @@
 using System.Threading.Tasks;
 
 using Coverlet.Core.Samples.Tests;
-using Coverlet.Tests.Xunit.Extensions;
 using Tmds.Utils;
 using Xunit;
 
@@ -10,54 +9,37 @@ namespace Coverlet.Core.Tests
 {
     public partial class CoverageTests : ExternalProcessExecutionTest
     {
-        [ConditionalFact]
-        [SkipOnOS(OS.MacOS)]
+        [Fact]
         public void Yield_Single()
         {
-            // We need to pass file name to remote process where it save instrumentation result
-            // Similar to msbuild input/output
             string path = Path.GetTempFileName();
             try
             {
-                // Lambda will run in a custom process to avoid issue with statics and file locking
                 FunctionExecutor.Run(async (string[] pathSerialize) =>
                 {
-                    // Run load and call a delegate passing class as dynamic to simplify method call
                     CoveragePrepareResult coveragePrepareResult = await TestInstrumentationHelper.Run<Yield>(instance =>
                     {
-                        // We call method to trigger coverage hits
                         foreach(var _ in instance.One());
 
-                        // For now we have only async Run helper
                         return Task.CompletedTask;
                     }, persistPrepareResultToFile: pathSerialize[0]);
 
-                    // we return 0 if we return something different assert fail
                     return 0;
                 }, new string[] { path });
 
-                // We retrive and load CoveragePrepareResult and run coverage calculation
-                // Similar to msbuild coverage result task
                 CoverageResult result = TestInstrumentationHelper.GetCoverageResult(path);
 
-                // Generate html report to check
-                // TestInstrumentationHelper.GenerateHtmlReport(result);
-
-                // Asserts on doc/lines/branches
                 result.Document("Instrumentation.Yield.cs")
-                      // (line, hits)
                       .AssertLinesCovered((9, 1))
                       .ExpectedTotalNumberOfBranches(0);
             }
             finally
             {
-                // Cleanup tmp file
                 File.Delete(path);
             }
         }
 
-        [ConditionalFact]
-        [SkipOnOS(OS.MacOS)]
+        [Fact]
         public void Yield_Multiple()
         {
             string path = Path.GetTempFileName();
@@ -68,6 +50,7 @@ namespace Coverlet.Core.Tests
                     CoveragePrepareResult coveragePrepareResult = await TestInstrumentationHelper.Run<Yield>(instance =>
                     {
                         foreach (var _ in instance.Two());
+
                         return Task.CompletedTask;
                     }, persistPrepareResultToFile: pathSerialize[0]);
                     return 0;
--- a/test/coverlet.core.tests/Coverage/CoverageTest.Yield.cs
+++ b/test/coverlet.core.tests/Coverage/CoverageTest.Yield.cs
@@ -19,7 +19,7 @@ namespace Coverlet.Core.Tests
                 {
                     CoveragePrepareResult coveragePrepareResult = await TestInstrumentationHelper.Run<Yield>(instance =>
                     {
-                        foreach(var _ in instance.One());
+                        foreach (var _ in instance.One()) ;
 
                         return Task.CompletedTask;
                     }, persistPrepareResultToFile: pathSerialize[0]);
@@ -50,7 +50,7 @@ namespace Coverlet.Core.Tests
                 {
                     CoveragePrepareResult coveragePrepareResult = await TestInstrumentationHelper.Run<Yield>(instance =>
                     {
-                        foreach (var _ in instance.Two());
+                        foreach (var _ in instance.Two()) ;
 
                         return Task.CompletedTask;
                     }, persistPrepareResultToFile: pathSerialize[0]);
@@ -142,7 +142,6 @@ namespace Coverlet.Core.Tests
                     CoveragePrepareResult coveragePrepareResult = await TestInstrumentationHelper.Run<Yield>(instance =>
                     {
                         foreach (var _ in instance.Enumerable(new[] { "one", "two", "three", "four" })) ;
-
                         return Task.CompletedTask;
                     }, persistPrepareResultToFile: pathSerialize[0]);
                     return 0;
@@ -152,7 +151,7 @@ namespace Coverlet.Core.Tests
 
                 result.Document("Instrumentation.Yield.cs")
                       .Method("System.Boolean Coverlet.Core.Samples.Tests.Yield/<Enumerable>d__4::MoveNext()")
-                      .AssertLinesCovered(BuildConfiguration.Debug, (48, 1), (49, 11), (50, 4), (51, 4), (52, 4), (53, 1))
+                      .AssertLinesCovered(BuildConfiguration.Debug, (48, 1), (49, 1), (50, 4), (51, 5), (52, 1), (54, 4), (55, 4), (56, 4), (57, 1))
                       .ExpectedTotalNumberOfBranches(1);
             }
             finally
--- a/test/coverlet.core.tests/Coverage/CoverageTest.Yield.cs
+++ b/test/coverlet.core.tests/Coverage/CoverageTest.Yield.cs
@@ -41,7 +41,7 @@ namespace Coverlet.Core.Tests
         }
 
         [Fact]
-        public void Yield_Multiple()
+        public void Yield_Two()
         {
             string path = Path.GetTempFileName();
             try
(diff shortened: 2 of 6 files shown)


MarcoRossignoli:
You should add also a complete real test inside folder https://github.com/tonerdo/coverlet/tree/master/test/coverlet.core.tests/Coverage:
1) Add a file named `CoverageTest.Yield.cs`
2) Add a file named `Instrumentation.Yield.cs` and add a simple class like your failing sample like https://github.com/tonerdo/coverlet/blob/master/test/coverlet.core.tests/Samples/Instrumentation.SelectionStatements.cs#L5
3) Inside `CoverageTest.Yield.cs` prepare and run real instrumentation test like https://github.com/tonerdo/coverlet/blob/master/test/coverlet.core.tests/Coverage/CoverageTests.SelectionStatements.cs#L14 read the comment as guide, copy same structure.

If you need to debug your test you have to simply change code like (`CoverageTests.SelectionStatements.cs` file sample)
```cs
  FunctionExecutor.RunInProcess(async (string[] pathSerialize) =>
                {
                    // Run load and call a delegate passing class as dynamic to simplify method call
                    CoveragePrepareResult coveragePrepareResult = await TestInstrumentationHelper.Run<SelectionStatements>(instance =>
                    {
                        // We call method to trigger coverage hits
                        instance.If(true);

                        // For now we have only async Run helper
                        return Task.CompletedTask;
                    }, persistPrepareResultToFile: pathSerialize[0], disableRestoreModules: true);

                    // we return 0 if we return something different assert fail
                    return 0;
                }, new string[] { path });
```
These real tests run in separate process so you cannot debug as is, you need to change in 2 place `RunInProcess` instead of `Run` and add `disableRestoreModules: true)` to avoid module restore at the end of instrumentation, this avoid error related to locked file.

Let me know if you're in throuble.

MarcoRossignoli:
cc: @matteoerigozzi

bert2:
Ok, I'll look into it.

I noticed one problem though: I'm unable to run the tests as described in [CONTRIBUTING.md](https://github.com/tonerdo/coverlet/blob/master/CONTRIBUTING.md#building-the-project):

```
PS> dotnet --version
3.1.201
PS> dotnet restore; dotnet build --no-restore; dotnet pack
[...]
PS> dotnet test --no-build /p:CollectCoverage=true /p:Include=\"[coverlet.collector]*,[coverlet.core]*,[coverlet.msbuild.tasks]*\" /p:Exclude=\"[coverlet.core.tests.samples.netstandard]*,[coverlet.tests.xunit.extensions]*\"
MSBUILD : error MSB1006: Property is not valid.
Switch: [coverlet.core]*

For switch syntax, type "MSBuild -help"
```

The escaping of the list doesn't seem to work on my end. Is this guideline still accurate?

MarcoRossignoli:
You're inside powershell you need to escale differently https://github.com/tonerdo/coverlet/blob/master/Documentation/MSBuildIntegration.md#note-for-powershell--vsts-users

bert2:
Well... that was easier than expected. Do you want me to squash those commits?

test/coverlet.core.tests/Coverage/CoverageTest.Yield.cs
```diff
@@ -0,0 +1,88 @@
...
+using Coverlet.Tests.Xunit.Extensions;
+using Tmds.Utils;
+using Xunit;
+
+namespace Coverlet.Core.Tests
+{
+    public partial class CoverageTests : ExternalProcessExecutionTest
+    {
+        [ConditionalFact]
+        [SkipOnOS(OS.MacOS)]
```
MarcoRossignoli, line 14: remove skip and use simple `Fact`, now we support mac.

test/coverlet.core.tests/Coverage/CoverageTest.Yield.cs
```diff
@@ -0,0 +1,88 @@
...
+        public void Yield_Single()
+        {
+            // We need to pass file name to remote process where it save instrumentation result
```
MarcoRossignoli, line 17: remove comments pls, we keep those only on "guide line test".

test/coverlet.core.tests/Coverage/CoverageTest.Yield.cs
```diff
@@ -0,0 +1,88 @@
...
+            }
+            finally
+            {
+                // Cleanup tmp file
+                File.Delete(path);
+            }
+        }
+
+        [ConditionalFact]
+        [SkipOnOS(OS.MacOS)]
```
MarcoRossignoli, line 60: ditto

MarcoRossignoli:
I'm bit busy these days...update seems ok but I need to go a bit deeper to confirm, I'll come back asap!Thank's!

src/coverlet.core/Symbols/CecilSymbolHelper.cs
```diff
@@ -413,6 +413,7 @@ public static List<BranchPoint> GetBranchPoints(MethodDefinition methodDefinitio
             bool isAsyncStateMachineMoveNext = IsMoveNextInsideAsyncStateMachine(methodDefinition);
             bool isMoveNextInsideAsyncStateMachineProlog = isAsyncStateMachineMoveNext && IsMoveNextInsideAsyncStateMachineProlog(methodDefinition);
             bool skipFirstBranch = IsMoveNextInsideEnumerator(methodDefinition);
+            bool skipSecondBranch = skipFirstBranch && instructions.All(i => i.OpCode.Code != Code.Switch);
```
MarcoRossignoli, line 416: This rule seem too generic, if inside method we have also switch instruction we won't skip that branch
```cs
        public System.Collections.Generic.IEnumerable<int> One()
        {
            int result = 0;
            switch (new Random().Next())
            {
                case 1:
                    {
                        result = 23;
                        break;
                    }
                case 2:
                    {
                        result = -5;
                        break;
                    }
                case 4:
                    {
                        result = -3;
                        break;
                    }
                default:
                    break;
            }

            if (result < 0)
            {
                throw new Exception();
            }

            yield return 1;
        }
```
Branch on `10` is wrong  `12` is `switch` and `33` the `if`
image

bert2: Darn, that's true. I added a failing test case for this scenario. Currently, I have no idea how to solve this. Any help would be much appreciated.

bert2: Sheesh, for some reason that test only fails in debug, but not in release configuration. When I use LINQPad to compare the generated MSIL in both cases it has the same number of branches though...

MarcoRossignoli: Need some time to understand how to fix this edge case, not simple...seem that we need to understand that we are inside enumerator and follow jump of first condition and skip that similar to:
```
if/switch (trim)

   another if/switch

case/else (trim)

   another if/switch

case/else (trim)

```
Not simple...this is the hardest thing with coverage 😄

bert2: I don't think following the execution flow is necessary, because it will always be either the first or the first and the second branch we need to trim/skip.

I just pushed a commit the fixes this issue. However, I'd feel uncomfortable using it until I understand why the test case didn't fail in release builds.

MarcoRossignoli: Have you compared with ILSpy?

bert2: Yes, of course. Did I miss something?

bert2: Oh you mean this in regard to the different integration test result of debug vs release. I couldn't find any structural differences.

<details><summary>debug MSIL</summary>

.method private final hidebysig newslot virtual 
	instance bool MoveNext () cil managed 
{
	.override method instance bool [System.Runtime]System.Collections.IEnumerator::MoveNext()
	// Method begins at RVA 0x1012c
	// Code size 142 (0x8e)
	.maxstack 2
	.locals init (
		[0] int32,
		[1] int32,
		[2] int32
	)

	IL_0000: ldarg.0
	IL_0001: ldfld int32 Coverlet.Core.Samples.Tests.Yield/'<OneWithSwitch>d__2'::'<>1__state'
	IL_0006: stloc.0
	IL_0007: ldloc.0
	IL_0008: brfalse.s IL_0012

	IL_000a: br.s IL_000c

	IL_000c: ldloc.0
	IL_000d: ldc.i4.1
	IL_000e: beq.s IL_0014

	IL_0010: br.s IL_0016

	IL_0012: br.s IL_0018

	IL_0014: br.s IL_0085

	IL_0016: ldc.i4.0
	IL_0017: ret

	IL_0018: ldarg.0
	IL_0019: ldc.i4.m1
	IL_001a: stfld int32 Coverlet.Core.Samples.Tests.Yield/'<OneWithSwitch>d__2'::'<>1__state'
	IL_001f: nop
	IL_0020: ldarg.0
	IL_0021: ldarg.0
	IL_0022: ldfld int32 Coverlet.Core.Samples.Tests.Yield/'<OneWithSwitch>d__2'::n
	IL_0027: stloc.1
	IL_0028: ldloc.1
	IL_0029: stfld int32 Coverlet.Core.Samples.Tests.Yield/'<OneWithSwitch>d__2'::'<>s__2'
	IL_002e: ldarg.0
	IL_002f: ldfld int32 Coverlet.Core.Samples.Tests.Yield/'<OneWithSwitch>d__2'::'<>s__2'
	IL_0034: stloc.2
	IL_0035: ldloc.2
	IL_0036: switch (IL_0049, IL_0053, IL_005d)

	IL_0047: br.s IL_0067

	IL_0049: ldarg.0
	IL_004a: ldc.i4.s 10
	IL_004c: stfld int32 Coverlet.Core.Samples.Tests.Yield/'<OneWithSwitch>d__2'::'<result>5__1'
	IL_0051: br.s IL_0070

	IL_0053: ldarg.0
	IL_0054: ldc.i4.s 11
	IL_0056: stfld int32 Coverlet.Core.Samples.Tests.Yield/'<OneWithSwitch>d__2'::'<result>5__1'
	IL_005b: br.s IL_0070

	IL_005d: ldarg.0
	IL_005e: ldc.i4.s 12
	IL_0060: stfld int32 Coverlet.Core.Samples.Tests.Yield/'<OneWithSwitch>d__2'::'<result>5__1'
	IL_0065: br.s IL_0070

	IL_0067: ldarg.0
	IL_0068: ldc.i4.m1
	IL_0069: stfld int32 Coverlet.Core.Samples.Tests.Yield/'<OneWithSwitch>d__2'::'<result>5__1'
	IL_006e: br.s IL_0070

	IL_0070: ldarg.0
	IL_0071: ldarg.0
	IL_0072: ldfld int32 Coverlet.Core.Samples.Tests.Yield/'<OneWithSwitch>d__2'::'<result>5__1'
	IL_0077: stfld int32 Coverlet.Core.Samples.Tests.Yield/'<OneWithSwitch>d__2'::'<>2__current'
	IL_007c: ldarg.0
	IL_007d: ldc.i4.1
	IL_007e: stfld int32 Coverlet.Core.Samples.Tests.Yield/'<OneWithSwitch>d__2'::'<>1__state'
	IL_0083: ldc.i4.1
	IL_0084: ret

	IL_0085: ldarg.0
	IL_0086: ldc.i4.m1
	IL_0087: stfld int32 Coverlet.Core.Samples.Tests.Yield/'<OneWithSwitch>d__2'::'<>1__state'
	IL_008c: ldc.i4.0
	IL_008d: ret
} // end of method '<OneWithSwitch>d__2'::MoveNext

</details>

<details><summary>release MSIL</summary>

.method private final hidebysig newslot virtual 
	instance bool MoveNext () cil managed 
{
	.override method instance bool [System.Runtime]System.Collections.IEnumerator::MoveNext()
	// Method begins at RVA 0xeca4
	// Code size 92 (0x5c)
	.maxstack 2
	.locals init (
		[0] int32,
		[1] int32 result,
		[2] int32
	)

	IL_0000: ldarg.0
	IL_0001: ldfld int32 Coverlet.Core.Samples.Tests.Yield/'<OneWithSwitch>d__2'::'<>1__state'
	IL_0006: stloc.0
	IL_0007: ldloc.0
	IL_0008: brfalse.s IL_0010

	IL_000a: ldloc.0
	IL_000b: ldc.i4.1
	IL_000c: beq.s IL_0053

	IL_000e: ldc.i4.0
	IL_000f: ret

	IL_0010: ldarg.0
	IL_0011: ldc.i4.m1
	IL_0012: stfld int32 Coverlet.Core.Samples.Tests.Yield/'<OneWithSwitch>d__2'::'<>1__state'
	IL_0017: ldarg.0
	IL_0018: ldfld int32 Coverlet.Core.Samples.Tests.Yield/'<OneWithSwitch>d__2'::n
	IL_001d: stloc.2
	IL_001e: ldloc.2
	IL_001f: switch (IL_0032, IL_0037, IL_003c)

	IL_0030: br.s IL_0041

	IL_0032: ldc.i4.s 10
	IL_0034: stloc.1
	IL_0035: br.s IL_0043

	IL_0037: ldc.i4.s 11
	IL_0039: stloc.1
	IL_003a: br.s IL_0043

	IL_003c: ldc.i4.s 12
	IL_003e: stloc.1
	IL_003f: br.s IL_0043

	IL_0041: ldc.i4.m1
	IL_0042: stloc.1

	IL_0043: ldarg.0
	IL_0044: ldloc.1
	IL_0045: stfld int32 Coverlet.Core.Samples.Tests.Yield/'<OneWithSwitch>d__2'::'<>2__current'
	IL_004a: ldarg.0
	IL_004b: ldc.i4.1
	IL_004c: stfld int32 Coverlet.Core.Samples.Tests.Yield/'<OneWithSwitch>d__2'::'<>1__state'
	IL_0051: ldc.i4.1
	IL_0052: ret

	IL_0053: ldarg.0
	IL_0054: ldc.i4.m1
	IL_0055: stfld int32 Coverlet.Core.Samples.Tests.Yield/'<OneWithSwitch>d__2'::'<>1__state'
	IL_005a: ldc.i4.0
	IL_005b: ret
} // end of method '<OneWithSwitch>d__2'::MoveNext

</details>

test/coverlet.core.tests/Coverage/InstrumenterHelper.Assertions.cs
```diff
@@ -79,11 +79,46 @@ public static Document Document(this CoverageResult coverageResult, string docNa
             throw new XunitException($"Document not found '{docName}'");
         }
 
+        public static Document Method(this Document document, string methodName)
```
MarcoRossignoli, line 82: Why you need this?I mean the assertion on lines seem enough

bert2: The problem is that `ExpectedTotalNumberOfBranches()` counts _all_ branches in the whole file. Without the `Method()` filter, I'd have to write `.ExpectedTotalNumberOfBranches(1)` in all three test cases, even when the method I'm testing doesn't have any branches which looks kind of weird. I.e. I'd be asserting things that have nothing to do with my test case.

MarcoRossignoli: Ok usually I test all three method in same test because are related, btw could be useful for future more complex tests assertions.

test/coverlet.core.tests/Samples/Instrumentation.Yield.cs
```diff
@@ -0,0 +1,40 @@
...
+namespace Coverlet.Core.Samples.Tests
+{
+    public class Yield
+    {
+        public System.Collections.Generic.IEnumerable<int> One()
+        {
+            yield return 1;
+        }
+
+        public System.Collections.Generic.IEnumerable<int> Two()
```
MarcoRossignoli, line 12: Can you also add also samples like this?
```cs
        public System.Collections.Generic.IEnumerable<int> Three()
        {
            yield return 1;
            yield return 2;
            yield return 3;
        }

        public System.Collections.Generic.IEnumerable<string> Enumerable(IList<string> ls)
        {
            foreach (var item in ls)
            {
                yield return item;
            }
        }
```

src/coverlet.core/Symbols/CecilSymbolHelper.cs
```diff
@@ -413,6 +413,7 @@ public static List<BranchPoint> GetBranchPoints(MethodDefinition methodDefinitio
             bool isAsyncStateMachineMoveNext = IsMoveNextInsideAsyncStateMachine(methodDefinition);
             bool isMoveNextInsideAsyncStateMachineProlog = isAsyncStateMachineMoveNext && IsMoveNextInsideAsyncStateMachineProlog(methodDefinition);
             bool skipFirstBranch = IsMoveNextInsideEnumerator(methodDefinition);
+            bool skipSecondBranch = false;
```
MarcoRossignoli, line 416: Can you add a comment above `skipFirstBranch` like
```
// State machine for enumerator is handled with brfalse/beq or switch opcode and it depends on how many 'yield' we have in method body.
// For more than one 'yield' a switch is emitted so we can skip only first branch.
// In case of one 'yield' we need to skip first two branches to avoid to report phantom branch, first branch to jump to 'yield' value, second one to exit from enumeration
```

bert2: > Some more tests, seems ok.

Yup, that's right. Will also add some more unit tests.

bert2:
> Some more tests, seems ok.

Yup, that's right. Will also add some more unit tests.

test/coverlet.core.tests/Coverage/CoverageTest.Yield.cs
```diff
@@ -100,5 +100,65 @@ public void Yield_SingleWithSwitch()
...
+                    }, persistPrepareResultToFile: pathSerialize[0]);
+                    return 0;
+                }, new string[] { path });
+
+                CoverageResult result = TestInstrumentationHelper.GetCoverageResult(path);
+
+                result.Document("Instrumentation.Yield.cs")
+                      .Method("System.Boolean Coverlet.Core.Samples.Tests.Yield/<Enumerable>d__4::MoveNext()")
+                      .AssertLinesCovered((51, 4))
+                      .ExpectedTotalNumberOfBranches(1);
```
MarcoRossignoli, line 156: Where is this branch?

test/coverlet.core.tests/Coverage/CoverageTest.Yield.cs
```diff
@@ -100,5 +100,65 @@ public void Yield_SingleWithSwitch()
...
+                        return Task.CompletedTask;
+                    }, persistPrepareResultToFile: pathSerialize[0]);
+                    return 0;
+                }, new string[] { path });
+
+                CoverageResult result = TestInstrumentationHelper.GetCoverageResult(path);
+
+                result.Document("Instrumentation.Yield.cs")
+                      .Method("System.Boolean Coverlet.Core.Samples.Tests.Yield/<Enumerable>d__4::MoveNext()")
+                      .AssertLinesCovered((51, 4))
```
MarcoRossignoli, line 155: can you be more specific with covered lines?i.e. 48,49,50,51,52,53

test/coverlet.core.tests/Coverage/CoverageTest.Yield.cs
```diff
@@ -0,0 +1,164 @@
...
+                    }, persistPrepareResultToFile: pathSerialize[0]);
+
+                    return 0;
+                }, new string[] { path });
+
+                CoverageResult result = TestInstrumentationHelper.GetCoverageResult(path);
+
+                result.Document("Instrumentation.Yield.cs")
+                      .Method("System.Boolean Coverlet.Core.Samples.Tests.Yield/<OneWithSwitch>d__2::MoveNext()")
+                      .AssertLinesCovered((30, 1), (31, 1), (37, 1))
```
MarcoRossignoli, line 95: Be more specific 20,36 seems missing

test/coverlet.core.tests/Coverage/CoverageTest.Yield.cs
bert2: The body of `Yield.Enumerable()` is a `foreach` loop. I assumed that they involve a branch. But I'm actually not sure.

MarcoRossignoli: Mmm maybe c# lies also here we should take a look at IL to understand if there is some branch, maybe an hidden null check

bert2: Yup you are right. Here's the generated `MoveNext()`:

```
<Enumerable>d__0.MoveNext:
IL_0000:  ldarg.0     
IL_0001:  ldfld       UserQuery+Yield+<Enumerable>d__0.<>1__state
IL_0006:  stloc.1     
IL_0007:  ldloc.1     
IL_0008:  brfalse.s   IL_0012
IL_000A:  br.s        IL_000C
IL_000C:  ldloc.1     
IL_000D:  ldc.i4.1    
IL_000E:  beq.s       IL_0014
IL_0010:  br.s        IL_0016
IL_0012:  br.s        IL_001D
IL_0014:  br.s        IL_006A
IL_0016:  ldc.i4.0    
IL_0017:  stloc.0     
IL_0018:  leave       IL_00A1
IL_001D:  ldarg.0     
IL_001E:  ldc.i4.m1   
IL_001F:  stfld       UserQuery+Yield+<Enumerable>d__0.<>1__state
IL_0024:  nop         
IL_0025:  nop         
IL_0026:  ldarg.0     
IL_0027:  ldarg.0     
IL_0028:  ldfld       UserQuery+Yield+<Enumerable>d__0.ls
IL_002D:  callvirt    System.Collections.Generic.IEnumerable<System.String>.GetEnumerator
IL_0032:  stfld       UserQuery+Yield+<Enumerable>d__0.<>s__1
IL_0037:  ldarg.0     
IL_0038:  ldc.i4.s    FD 
IL_003A:  stfld       UserQuery+Yield+<Enumerable>d__0.<>1__state
IL_003F:  br.s        IL_007A
IL_0041:  ldarg.0     
IL_0042:  ldarg.0     
IL_0043:  ldfld       UserQuery+Yield+<Enumerable>d__0.<>s__1
IL_0048:  callvirt    System.Collections.Generic.IEnumerator<System.String>.get_Current
IL_004D:  stfld       UserQuery+Yield+<Enumerable>d__0.<item>5__2
IL_0052:  nop         
IL_0053:  ldarg.0     
IL_0054:  ldarg.0     
IL_0055:  ldfld       UserQuery+Yield+<Enumerable>d__0.<item>5__2
IL_005A:  stfld       UserQuery+Yield+<Enumerable>d__0.<>2__current
IL_005F:  ldarg.0     
IL_0060:  ldc.i4.1    
IL_0061:  stfld       UserQuery+Yield+<Enumerable>d__0.<>1__state
IL_0066:  ldc.i4.1    
IL_0067:  stloc.0     
IL_0068:  leave.s     IL_00A1
IL_006A:  ldarg.0     
IL_006B:  ldc.i4.s    FD 
IL_006D:  stfld       UserQuery+Yield+<Enumerable>d__0.<>1__state
IL_0072:  nop         
IL_0073:  ldarg.0     
IL_0074:  ldnull      
IL_0075:  stfld       UserQuery+Yield+<Enumerable>d__0.<item>5__2
IL_007A:  ldarg.0     
IL_007B:  ldfld       UserQuery+Yield+<Enumerable>d__0.<>s__1
IL_0080:  callvirt    System.Collections.IEnumerator.MoveNext
IL_0085:  brtrue.s    IL_0041
IL_0087:  ldarg.0     
IL_0088:  call        UserQuery+Yield+<Enumerable>d__0.<>m__Finally1
IL_008D:  nop         
IL_008E:  ldarg.0     
IL_008F:  ldnull      
IL_0090:  stfld       UserQuery+Yield+<Enumerable>d__0.<>s__1
IL_0095:  ldc.i4.0    
IL_0096:  stloc.0     
IL_0097:  leave.s     IL_00A1
IL_0099:  ldarg.0     
IL_009A:  call        UserQuery+Yield+<Enumerable>d__0.System.IDisposable.Dispose
IL_009F:  nop         
IL_00A0:  endfinally  
IL_00A1:  ldloc.0     
IL_00A2:  ret
```

`IL_0085` is a `brtrue.s` back up to `IL_0041` which seems to implement a loop like:

```csharp
do {
    // ...
} while (MoveNext())
```

bert2: After having a closer look at this case it turned out to be a true nightmare. I don't see a way how to skip that branch from the `foreach` without the possibility of accidentally skipping a branch that actually was part of the user's code.

It might be that I'll be able to find some kind of pattern in the generated IL when I compare lots of different cases, but I'm not sure when I'll find the time to do that.

test/coverlet.core.tests/Coverage/CoverageTest.Yield.cs
bert2: Line 20 (`int result;`) is not an actual instruction. Line 36 is blank. I added line 21 (`switch (n)`).

test/coverlet.core.tests/Coverage/CoverageTest.Yield.cs
bert2: I updated the assert, but line 49 (`foreach (var item in ls)`) got 11 hits which seems too high to me. Is this correct or another bug?

test/coverlet.core.tests/Coverage/CoverageTest.Yield.cs
bert2: Looks like line 21 is only hit in debug builds, but not in release builds.

test/coverlet.core.tests/Coverage/CoverageTest.Yield.cs
bert2: Also added extra assertions for release builds.

test/coverlet.core.tests/Coverage/CoverageTest.Yield.cs
MarcoRossignoli: For now it's ok...I had the same thought when I see that branch...too risky an btw that branch will be always covered if there is at least one element.

test/coverlet.core.tests/Coverage/CoverageTest.Yield.cs
MarcoRossignoli: ok usually we test only debug build, feel free to avoid release one.

test/coverlet.core.tests/Coverage/CoverageTest.Yield.cs
MarcoRossignoli: > I updated the assert, but line 49 (foreach (var item in ls)) got 11 hits which seems too high to me. Is this correct or another bug?

Interesting, we should understand why, try to debug test and take a look this part to understand if that row has got 11 hits or is a sum https://github.com/tonerdo/coverlet/blob/master/src/coverlet.core/Coverage.cs#L357

bert2: As you suspected it's a sum. `foreach (var item in ls)` can have four individual breakpoints:

```csharp
foreach (       // hit 1x on entry
    var item    // hit 4x
    in          // hit 5x
    ls)         // hit 1x on entry
```

test/coverlet.core.tests/Coverage/CoverageTest.Yield.cs
bert2: Ok, I removed the release build assertions again to stay consistent with the other tests.

test/coverlet.core.tests/Coverage/CoverageTest.Yield.cs
MarcoRossignoli: ok so it's expected!

MarcoRossignoli: Refactored test a bit to be more clear!

MarcoRossignoli:
@bert2 feel free to take any other issue, if you like coverage one this is the query https://github.com/coverlet-coverage/coverlet/issues?q=is%3Aissue+is%3Aopen+label%3Atenet-coverage maybe start from untriaged.
There aren't a lot of devs that likes lose himself in the rabbit hole 😅

MarcoRossignoli:
CI ran but there were some issue between github and AzDo likely because we're in the middle of move btw this is the link of CI https://dev.azure.com/tonerdo/coverlet/_build/results?buildId=1023&view=results

bert2:
@MarcoRossignoli it definitely was a very interesting journey :) I will probably look into some of the coverage bugs, however I'm not sure when I'll be able to spend that much time for it again.

MarcoRossignoli:
> @MarcoRossignoli it definitely was a very interesting journey :) I will probably look into some of the coverage bugs, however I'm not sure when I'll be able to spend that much time for it again.

No props at all it's not mandatory!Thank's again for the PR and help!

bert2:
Hey @MarcoRossignoli, I wrote a CodeLens extension for Visual Studio that shows you the instructions of a method: https://github.com/bert2/microscope

Shameless self-promotion, I know, but it I think it could be quite useful when working on software like coverlet 😊

MarcoRossignoli:
Thank you I'll give it a try for sure!

PR #820, opened 2020-04-18
Add deterministic build tests sample

MarcoRossignoli:
Add deterministic build tests sample

cc: @clairernovotny


clairernovotny:
Does this show the issue without the workarounds/updates?

MarcoRossignoli:
> Does this show the issue without the workarounds/updates?

No no it's only a sample on how to workaround. Tomorrow I'll try to test with more tfms, this is only part of documentation.
I need to add some explanation and link to readme.

Issue #825, opened 2020-04-25
Allow paremeters to support set the paths in reports (run test inside docker)


JorTurFer:
When I am trying to generate cobertura report inside a docker container, the source element in the xml takes the root path inside the container

```xml
<sources>
    <source>/</source>
</sources>
```

Then the filename attribute takes the full path from the container.

```xml
<class name="Device.Simulator.Worker" filename="/source/src/Device.Simulator/Worker.cs" line-rate="1" branch-rate="1" complexity="1">
```

This paths produce error when I am trying to generate a report for example.
![image](https://user-images.githubusercontent.com/36899226/80280810-d1f0e380-8706-11ea-8893-fe935de60bb3.png)

The only workaround that I found is making a replace in the paths inside the file but will be fantastic if coverlet supports the option of set the source element's value and the origin of the relative path for the filename.

I can make a pull request if you think that this option is usefull.

I have tha same problem runing in local and in an Azure Pipeline

MarcoRossignoli:
> This paths produce error when I am trying to generate a report for example.

Only to be sure to understand well, you move report out of container and after generate the report?Or run report generation inside a docker?

JorTurFer:
I run the tests inside container and with a binding mount I mové out of container to make the other steps.
Give me 1 hour and I will post an example repo with a pipeline where you can see the problem.

JorTurFer:
Hi @MarcoRossignoli ,
I created an example repository with the cases that I exposed

https://github.com/JorTurFer/coverlet-issue-example

The repository has 2 CI, one with replacements and other without them in order to be able to check the problem in the next steps. You can download the html report from each pipeline.

rickywiens:
Hi @MarcoRossignoli and @JorTurFer! My team has run in to a [similar problem](https://gitlab.com/gitlab-org/gitlab/-/issues/213444) with the cobertura report parsing from the coverlet library. If we could pass in a parameter that would allow us to prime the `<source>` node with the project path, and calculate the `filename` attributes based on that, it would allow us to avoid modifying the cobertura xml file after it is generated.

You can see an example project in the issue link I pasted above. We are also building the report inside of a docker container and analyzing it outside of a docker container in a CI environment in order to enable [test coverage visualization](https://docs.gitlab.com/ee/user/project/merge_requests/test_coverage_visualization.html). Please let me know if there's anything I can do to assist.

JorTurFer:
[...]

Issue #413, opened 2019-05-17
Unable to change paths in cobertura report


EricStG:
Getting an issue where our application is compiled within docker and we're trying to export the cobertura xml file from it. 

Since the `filename` attribute is the full path to the file, we need to do a search a replace. 

Adding the root of the project to `sources/source` and using relative paths for `filename`, it would be a lot easier to manager.

sasivishnu:
I agree, this will be nice enhancement. Instead of doing only for cobertura format, It will be good to do for all other format's except Team City. We can add one more options, `--format --relative-code-path`

MarcoRossignoli:
could be similar to https://github.com/tonerdo/coverlet/issues/408?
and https://github.com/tonerdo/coverlet/issues/408#issuecomment-491502234 could fix?

sasivishnu:
Seems like Coverlet was using relative path earlier but @tonerdo changed it.

[#295 (Comment)](https://github.com/tonerdo/coverlet/issues/295#issuecomment-469011441)

> This happens because I switched to using absolute paths. #356 re-includes the source element

I think, change was done as part of  #260
https://github.com/tonerdo/coverlet/commit/de558c5f62b46d30fb2ba15231e50fc7f41877f7?diff=unified

There is an open issue for similar request in Icov Format, #263

ViceIce:
jenkins don't like absolute path, I would like to see a switch to change to relative path.

MarcoRossignoli:
FYI https://github.com/tonerdo/coverlet/issues/408#issuecomment-497674737

daveMueller:
In #614 we changed the cobertura report back to using relative paths so that the jenkins plugin is working again and the source link option is still supported. 
The difference is that currently the root dir of the OS is used in ```sources/source``` and not the root of the project. Most likely this could be changed but I'm not really sure how to determine the root of the project? To be precise, I'm not sure about the definition of project root. Is it the .csproj files, the .sln file or even the repository?
@MarcoRossignoli What do you think?

MarcoRossignoli:
@daveMueller I think that we shouldn't search the "root" of a project, but the algs should simply try to find longest shared absolute(per drive on win) and add to sources, the relative parts will go inside file attribute.

Report is "unrelated" to project structure and we don't know how a project will be setup, could be created runtime, could o couldn't have a sln an so on...report wants only "locate" source files.
Something similar to trie structure where the absolute stops at first fork and letter are folders.

[...]

PR #828, opened 2020-05-01
[WIP] PoC about override paths for generate Cobertura report inside docker containers

JorTurFer:
This PR is a PoC about solve the problem in #825

### DIFF
--- a/src/coverlet.collector/DataCollection/CoverageWrapper.cs
+++ b/src/coverlet.collector/DataCollection/CoverageWrapper.cs
@@ -32,7 +32,8 @@ namespace Coverlet.Collector.DataCollection
                 coverletLogger,
                 instrumentationHelper,
                 fileSystem,
-                sourceRootTranslator);
+                sourceRootTranslator,
+                null);
         }
 
         /// <summary>
--- a/src/coverlet.console/Program.cs
+++ b/src/coverlet.console/Program.cs
@@ -61,6 +61,7 @@ namespace Coverlet.Console
             CommandOption singleHit = app.Option("--single-hit", "Specifies whether to limit code coverage hit reporting to a single hit for each location", CommandOptionType.NoValue);
             CommandOption mergeWith = app.Option("--merge-with", "Path to existing coverage result to merge.", CommandOptionType.SingleValue);
             CommandOption useSourceLink = app.Option("--use-source-link", "Specifies whether to use SourceLink URIs in place of file system paths.", CommandOptionType.NoValue);
+            CommandOption pathOverrides = app.Option("--path-overrides", "Specifies the source path and the base path where filename is overrided from", CommandOptionType.SingleValue);
 
             app.OnExecute(() =>
             {
@@ -69,6 +70,9 @@ namespace Coverlet.Console
 
                 if (!target.HasValue())
                     throw new CommandParsingException(app, "Target must be specified.");
+
+                if(pathOverrides.HasValue() && pathOverrides.Value().Split(",").Length != 2)
+                    throw new CommandParsingException(app, "The source path and base path overrides are required");
 
                 if (verbosity.HasValue())
                 {
@@ -88,7 +92,8 @@ namespace Coverlet.Console
                     logger,
                     DependencyInjection.Current.GetService<IInstrumentationHelper>(),
                     fileSystem,
-                    sourceTranslator);
+                    sourceTranslator,
+                    pathOverrides.Value().Split(","));
                 coverage.PrepareModules();
 
                 Process process = new Process();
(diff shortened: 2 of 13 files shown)


MarcoRossignoli:
The idea is clear...I'm not sure if have specific code/flag for environment+specific report type is a good thing, or if these kind of problem should be fixed by external script, I mean the use cases are very different a one solution could be not enough or correct or working.
BTW keep this PoC open to discuss on it cc: @tonerdo @petli @jakubch1

rickywiens:
> The idea is clear...I'm not sure if have specific code/flag for environment+specific report type is a good thing or if these kind of problem should be fixed by external script, I mean the use cases are very different a one solution could be not enough or correct or working.
> BTW keep this PoC open to discuss on it cc: @tonerdo @petli @jakubch1
> Thanks @JorTurFer for the effort!

FYI I've documented a possible external workaround here: https://gitlab.com/gitlab-org/gitlab/-/issues/213444#note_334104771

As you can see it is not very robust, and involves brittle regex which can be troublesome for people to modify to suit their own purposes. I think that having the flexibility to use a relative path in the `filename attribute` on the `class` node by setting the `source` node contents upfront is a more elegant solution than modifying the file after the fact.

MarcoRossignoli:
Sorry but maybe I'm missing something, at the moment cobertura report try to put a base shared path inside source element with relative path on class element(https://github.com/coverlet-coverage/coverlet/issues/413).
Which version are you using?

JorTurFer:
You are right @MarcoRossignoli. The issue that I opened and this PR are related with the diferent paths inside/outside the containers. 
The case that I want solve with this issue-PR is when you have a path inside container that doesn't exist outside container. For example in an Azure Pipelines, you should use the Azure Pipelines agent's folder hierarchy inside the conteiner in order to be able to use the cobertura reports outside the container.

rickywiens:
> Sorry but maybe I'm missing something, at the moment cobertura report try to put a base shared path inside source element with relative path on class element(#413).
> The issue it's different if I understood well, the path inside container does not exists outside container so there is no way to find sources by report generators, right?

For our purposes being able to designate a source directory would be enough to allow for our software to parse the file correctly. The issue being the algorithm (#413) used to select a base shared path isn't sufficient to avoid sometimes (not always) putting an absolute path that doesn't exist in the CI file system (only exists in the docker filesystem) inside of the `filename` attribute of the `class` nodes.

In the [example](https://gitlab.com/mclayton7/dotnet-example/) I have previously linked to in the corresponding GitLab issue there is only one Library directory in the project, with only one class inside that library. Because of that the algorithm will put the entirety of the absolute path into the `filename` attribute, and the only thing that we can do is to remove it manually after the fact in order to allow it to work with our code coverage feature (Because the absolute path is referring to a containerized file-system). Caveat: I'm not very experienced with C# so I'm unsure of how often this specific situation would occur. 

Essentially if I was able to configure it so that in the above example project the `filename` attribute of the class node would contain: `ExampleLib/Calculator.cs` no matter how many files were in the ExampleLib directory, or how many other directories there were (Not sure if that is common in C# either) that would be ideal.

MarcoRossignoli:
> In the example I have previously linked to in the corresponding GitLab issue there is only one Library directory in the project, with only one class inside that library. Because of that the algorithm will put the entirety of the absolute path into the filename attribute,

This is a bug I think(need to investigate) we would have a base source always also for a single source, this is a corner case so it wasn't happened yet to some user. Thank's for reporting!

> Because the absolute path is referring to a containerized file-system

Yep this is related to this issue ok, container path are not valid out of container. I don't have a strong opinion yet, waiting other guys for agreement/thought on fix @tonerdo @petli @jakubch1, I mean container app are very frequent today so we need to chose if fix here or let to user fix after generation, not so comfortable I'm seeing.

JorTurFer:
Any news about this?

MarcoRossignoli:
Not strong opinion, I don't like too much the idea to add specific command for specific report don't seem a feat of a coverage tool imo, btw wait other @tonerdo @petli and ms guy @jakubch1 if wants give an opinion.

tonerdo:
Shouldn't this problem happen with other reports as well? We pretty much always use absolute paths in all reports

MarcoRossignoli:
@tonerdo are you suggesting to add a parameter like `-basePathSourceReport` and update all reports where file path is reported?
If so do we need to pass a mapping?We could have different project in same report(merge) and so mayb we need something like `-basePathSourceReportMapping "/app;c:\app,/app2:c:\app2"` and so on.

JorTurFer:
> Shouldn't this problem happen with other reports as well? We pretty much always use absolute paths in all reports

I am not sure that this happens with other reports, but provably yes if the other reports are used with another report generator tools. I discovered this problem with the tool that I use but is possible that happens with others

PR #849, opened 2020-05-16
Fix skip [ExcludefromCoverage] for generated async state machine

MarcoRossignoli:
Contributes to https://github.com/coverlet-coverage/coverlet/issues/809
We need also to fix issue with lamda.

NB. Cannot add tests because in all other case the `ordinal` value calculation is correct so I cannot find an example where old alg(synthetized names) fails.

cc: @matteoerigozzi @VenkateshSrini


MarcoRossignoli:
Local manual test with repro

image

Still present for lambda

image

src/coverlet.core/Instrumentation/Instrumenter.cs
```diff
@@ -393,6 +424,11 @@ private void InstrumentType(TypeDefinition type)
                         customAttributes = customAttributes.Union(prop.CustomAttributes);
                 }
 
+                if (IsMethodOfCompilerGeneratedClassOfAsyncStateMachineToBeExcluded(method))
+                {
+                    continue;
+                }
+
                 ordinal++;
```
matteoerigozzi, line 432: I think this instruction must be moved before the if

MarcoRossignoli: Done can you take a look?

MarcoRossignoli:
@matteoerigozzi PTAL

MarcoRossignoli:
Test case with lambda bug
image

PR #884, opened 2020-06-23
Fix attribute exclusion

### LINKED ISSUE #883: Attribute exclusion does not work if attribute name does not end with "Attribute"
I came across the same issue as #604 and added `TestSDKAutoGeneratedCode` to my `ExcludeByAttribute` usage (I'm using the MSBuild integration).

However, it looks like coverlet assumes all attributes' type names will end with `Attribute`:
https://github.com/coverlet-coverage/coverlet/blob/f7c3fd4e3a52de4eda4d3aa26918487e23f2aaa1/src/coverlet.core/Instrumentation/Instrumenter.cs#L696-L697
That isn't the case for `TestSDKAutoGeneratedCode`:
https://github.com/microsoft/vstest/blob/50b5b33261f7593eb92ae221dae1b9c74234207e/src/Microsoft.TestPlatform.ObjectModel/CodeCoverage/TestSDKAutoGeneratedCode.cs#L10

I believe the fix would be to do something like this, to keep the original functionality:
```c#
return excludeAttributeNames.Any(a =>
    a.EndsWith("Attribute")
        ? customAttribute.AttributeType.Name.Equals(a)
        : customAttribute.AttributeType.Name.Equals($"{a}Attribute"));
```

bddckr:
Closes #883

The following changes have been added additionally:
* Reduce allocations to be done once only.
* Update the documentation to get rid of a mistake: Fully qualified attribute type names are actually not supported. I checked all the other documentation documents - they are already OK.


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/coverlet-coverage/coverlet?pullRequest=884) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/coverlet-coverage/coverlet?pullRequest=884) before we can accept your contribution.<br/><br/>:x: bddckr [sign now](https://cla.dotnetfoundation.org/coverlet-coverage/coverlet?pullRequest=884)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/coverlet-coverage/coverlet?pullRequest=884) it.</sub>

bddckr:
I took the liberty to also
* Reduce allocations to be done once only.
* Update the documentation to get rid of a mistake: Fully qualified attribute type names are actually not supported. I checked all the other documentation documents - they are already OK.

MarcoRossignoli:
Thanks, can you add a test with attribute doesn't end with `Attribute`?
You can add a test near https://github.com/coverlet-coverage/coverlet/blob/master/test/coverlet.core.tests/Instrumentation/InstrumenterTests.cs#L125 and a sample near https://github.com/coverlet-coverage/coverlet/blob/master/test/coverlet.core.tests/Samples/Samples.cs#L205

bddckr:
@MarcoRossignoli I totally forgot about those 😅 Added now. Let me know if you think the test changes are OK. I figured there's no need to create a custom attribute type as the one that originally gave us issues can be referenced directly in the tests. I also didn't add any specific test for this and instead updated the various exclusion tests so they all test this case, too.

bddckr:
I had to update the `CecilSymbolHelperTests` to make the tests pass. I think that is needed due to me adding a test class into `DeclaredConstructorClass`, but let me know if that is wrong.

MarcoRossignoli:
> I had to update the CecilSymbolHelperTests to make the tests pass. I think that is needed due to me adding a test class into DeclaredConstructorClass, but let me know if that is wrong.

It's ok, it's related to new `using` on top of sample file.

PR #887, opened 2020-06-25
Add configuration option InstrumentationRetries

### LINKED ISSUE #886: Collector fails due to blocked files, propose simple fix
Hi Everyone, I realize that a similar issue has been discussed recently, but I have some substantiated evidence about what exactly happens. In my case McAfee Adaptive Thread Protection (yikes) is blocking the instrumented modules, so that `InstrumentationHelper.RestoreOriginalModule()` fails to restore the original file even after the default 12 retries with increasing sleep intervals of 6, 12, 24, ... milliseconds.

For me, it is really sufficient to increase that interval by modifing the default `initialSleepSeconds` (which is actually misnamed, as the number is interpreted as milliseconds) of `InstrumentationHelper.CreateRetryStrategy()` from 6 to a somewhat larger value of 50. Alternatively increasing the number of retries from 12 to 15 or something along these lines also fixes it for me.

It would be great if either of those values would be configurable, either by a config file, command line argument or MSBuild parameter.

I would do the neccessary changes myself if someone more familiar with coverlet could give me a hint about what would be the appropriate method for injecting this parameter/configuration into the `InstrumentationHelper`. 

Edit: I realized I talked about having evidence but did not include it here. I used Sysinternals Process Monitor to see when dotnet.exe, testhost.exe and mfeatp.exe (McAfee) open and close the files, and this is exactly what I saw: dotnet.exe failing to open the file for writing, because mfeatp.exe has this file still open and allows only read only sharing. As soon as mfeatp.exe closes the file, dotnet.exe succeeds in opening the file for writing.

Ablinne:
This adds a configuration option for the number of retries that the various RetryHelpers in InstrumentationHelper do.

Fixes #886 .

Please let me know if anything is not in the right place.

### DIFF
--- a/src/coverlet.collector/DataCollection/CoverageWrapper.cs
+++ b/src/coverlet.collector/DataCollection/CoverageWrapper.cs
@@ -27,7 +27,8 @@ namespace Coverlet.Collector.DataCollection
                 IncludeTestAssembly = settings.IncludeTestAssembly,
                 SingleHit = settings.SingleHit,
                 MergeWith = settings.MergeWith,
-                UseSourceLink = settings.UseSourceLink
+                UseSourceLink = settings.UseSourceLink,
+                InstrumentationRetries = settings.InstrumentationRetries
             };
 
             return new Coverage(
--- a/src/coverlet.collector/DataCollection/CoverletSettings.cs
+++ b/src/coverlet.collector/DataCollection/CoverletSettings.cs
@@ -63,6 +63,11 @@ namespace Coverlet.Collector.DataCollection
         /// </summary>
         public bool IncludeTestAssembly { get; set; }
 
+        /// <summary>
+        /// Instrumentation retries
+        /// </summary>
+        public int InstrumentationRetries { get; set; }
+
         public override string ToString()
         {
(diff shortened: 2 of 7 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/coverlet-coverage/coverlet?pullRequest=887) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/coverlet-coverage/coverlet?pullRequest=887) before we can accept your contribution.<br/><br/>:x: Ablinne [sign now](https://cla.dotnetfoundation.org/coverlet-coverage/coverlet?pullRequest=887)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/coverlet-coverage/coverlet?pullRequest=887) it.</sub>

tonerdo:
@Ablinne Thanks for the contribution! The number of times coverlet attempts to instrument is an implementation detail and I'll prefer it stays that way

Issue #898, opened 2020-07-12
Exclude lines & branches that come after calls to methods with [DoesNotReturn]


kevin-montrose:
It'd be nice to be able to exclude lines, like `return;`, that follow methods that contractually never return from coverage.  The BCL already ships with the [`DoesNotReturnAttribute`](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.doesnotreturnattribute) that could be used to indicate this.

I stumble across this a lot when using the "throw helper"-pattern.  An example from [one my GitHub repos](https://github.com/kevin-montrose/Cesil).

[`PoisonableBase<T>`](https://github.com/kevin-montrose/Cesil/blob/0dcaeccf6234f1b28037fdf7e54f38f9d0154203/Cesil/Common/PoisonableBase.cs#L36) has a method like so:
```csharp
internal void AssertNotPoisoned<T>(IBoundConfiguration<T> self)
{
    if (Poison != null)
    {
        switch (Poison.Value)
        {
            case PoisonType.Cancelled: 
                Throw.InvalidOperationException<object>("Object is in an invalid state, a previous operation was canceled"); 
                return;
            case PoisonType.Exception: 
                Throw.InvalidOperationException<object>("Object is in an invalid state, a previous operation raised an exception"); 
                return;
            default:
                Throw.ImpossibleException<object, T>($"Unexpected {nameof(PoisonType)}: {Poison}", self);
                return;
        }
    }
}
```

Each method on [`Throw`](https://github.com/kevin-montrose/Cesil/blob/0dcaeccf6234f1b28037fdf7e54f38f9d0154203/Cesil/Common/Throw.cs) never returns.

Even though I do have tests for each `PoisonType`, my reports will always include each `return` statement as uncovered:

 - ![image](https://user-images.githubusercontent.com/1219467/87252404-b7462500-c440-11ea-9215-2fafed259c07.png)

For non-void methods I can hack around this restriction by pretending these methods return an object, but for void returning methods you're out of luck.

---

I suspect this could be hacked into [`Instrumentor.InstrumentIL`](https://github.com/coverlet-coverage/coverlet/blob/master/src/coverlet.core/Instrumentation/Instrumenter.cs#L494), but maybe there's a pass elsewhere that would be a better place.

I _think_ the logic to support this is something like:

 1. For every method body...
     a. Break the IL into runs based on branch targets
[...]

PR #904, opened 2020-07-20
Exclude code that follows [DoesNotReturn] from code coverage (per #898)

kevin-montrose:
Per #898 

Implementing reachability analysis, which will allow instructions that follow calls to methods that never return to be excluded from coverage metrics.

Determines what methods will return with [DoesNotReturn], though that is configurable.
Adds a test that explores a few "odd" patterns.

Though this seems to work, there are a few open questions:

 - Preferred way to punch configuration of this attribute in?
   * The bits are there for setting in on Coverage, but the last mile (how to set it from the command line, build tasks, etc.) needs to be done
 - Test suite is really fragile, especially around multiple tests locking each other out of the same files - is there a better way to run them?
   * I've been using [the contributing instructions](https://github.com/coverlet-coverage/coverlet/blob/master/CONTRIBUTING.md) and killing `dotnet build-server shutdown` (with copious direct killing of dotnet and MSBuild process that VS spins up) and there are still some tests (like `Coverlet.Core.Instrumentation.Tests.InstrumenterTests.TestInstrumentCoreLib`) that just will not run
 - Reachability does not distinguish between "unreachable because of method call" and "unreachable because the IL is actually unreachable"
   * The compiler is... disinclined to produce unreachable IL, but this is a real (potentially) significant change
 - Related, reachability is always calculated and used - should it be ignored if there are no calls to [DoesNotReturn] methods?
 - "Odder" unconditional branches are not handled (leave, leave_s, jmp, etc.) yet
 - Naming and code layout things
   * There are several new classes (mostly nested), where should these go?
   * There's yet-another-block-class, what should it be named?  Are there existing classes that can be used instead?

Finally, there are performance concerns.  Current approach doesn't go wild with allocations or an insane number of passes or anything, but could probably be improved.


kevin-montrose:
So these failures are much easier to understand than what I was seeing locally.  Appears that this [`Resolve`](https://github.com/coverlet-coverage/coverlet/blob/e915869de3ad0f76d0dca9e8fbca65054dc7b42f/src/coverlet.core/Instrumentation/Instrumenter.cs#L1337) is the where "file in use" is killing things.

@MarcoRossignoli maybe I should be treating corelib specially?  Or is there something else I should consider.

MarcoRossignoli:
Thanks Kevin I'll take a deep look asap(bit busy moment) cc: @tonerdo @petli for double check

kevin-montrose:
Poking around a little more, the root issue seems to be that I'm trying to resolve a type that is defined in a module that is being actively instrumented.  I think if I instead build a list of `[DoesNotReturn]` methods _in the current module_ and check against that list instead of `Resolve()`'ing  when the referenced method is defined in the current module it'll go away.

Will see if I can get that hacked up in the next day or two.

kevin-montrose:
Ultimately ended up doing a read-only pass over a module before instrumentation starts to avoid any need to call `.Resolve()` during actual instrumentation.  This removes concerns around trying to open an assembly file that's being modified, but does introduce another pass over the module.

There's still a chance that resolving fails (if, for example, the assembly referenced cannot be found).  This seems to happen in some integrations tests, which might just be an error in their configuration ([seen in these logs](https://dev.azure.com/tonerdo/coverlet/_build/results?buildId=1139&view=logs&jobId=a27823c2-9b39-589b-c166-e9f67508f4ea)).  I chose to assume those methods always return - whether that is correct is debatable.  [Relevant logic is here](https://github.com/coverlet-coverage/coverlet/blob/575edcbadfd80aa73cb3d843d5b590dc307ccabe/src/coverlet.core/Instrumentation/Instrumenter.cs#L1168) if it should change.

I expanded [the new test case](https://github.com/coverlet-coverage/coverlet/blob/575edcbadfd80aa73cb3d843d5b590dc307ccabe/test/coverlet.core.tests/Samples/Instrumentation.DoesNotReturn.cs#L116) to include generic methods and methods on generic types, which did not work in my first pass at this.

I've also run this against the personal project that prompted #898, and confirmed this PR addresses my original issue:
image

MarcoRossignoli:
Thanks again Kevin I'm on vacation this weeks I'll take a look asap :bow:

MarcoRossignoli:
> Preferred way to punch configuration of this attribute in?
The bits are there for setting in on Coverage, but the last mile (how to set it from the command line, build tasks, etc.) needs to be done

You mean a way to opt-out in case of bugs?At the moment we don't have settings for coverage engine...ops sorry we have new one added last PR only to skip autoprops(but it's related to needs of user to avoid a % increase decrease for autoprops, make sense). Btw could be better avoid too much flags, are you concerned about some ambiguity on some custom user `DoesNotReturn` attribute?

> Test suite is really fragile, especially around multiple tests locking each other out of the same files - is there a better way to run them?

To test in local usually works well
```
git clean -fdx
dotnet restore
dotnet pack
dotnet test
```
Btw for testing you should add some tests like in file https://github.com/coverlet-coverage/coverlet/pull/904/files#diff-b0c2dc6aa3b37bd4720bfeea133be676 (I pushed a file sample to speedup)
In this way you can debug step-by-step inside core code, test/debug instrumentation is a bit complex due to invasive nature(rewrite/rollback at the end).
If you run test inside vs at the end you'll see live report generation for fast check.

> Reachability does not distinguish between "unreachable because of method call" and "unreachable because the IL is actually unreachable"
The compiler is... disinclined to produce unreachable IL, but this is a real (potentially) significant change
Related, reachability is always calculated and used - should it be ignored if there are no calls to [DoesNotReturn] methods?

I think we should apply only in case of call to `DoesNotReturn` if doable with low possible noise.

> "Odder" unconditional branches are not handled (leave, leave_s, jmp, etc.) yet

Ok

> Naming and code layout things
There are several new classes (mostly nested), where should these go?
There's yet-another-block-class, what should it be named? Are there existing classes that can be used instead?

Create a new file under `Instrumentation` folder called for instance `ReachabilityHelper.cs` and add all classes you need under `Coverlet.Core.Instrumentation.Reachability` namespace. It's a complex feat feel free to add needed object for it.

MarcoRossignoli (approved):
Thanks and sorry for the delay 🙇 busy summer

PR #912, opened 2020-07-26
Add skipautoprops feature

MarcoRossignoli:
closes https://github.com/coverlet-coverage/coverlet/issues/328
High pri feat as planned https://github.com/coverlet-coverage/coverlet/blob/master/Documentation/Roadmap.md#high-priority

I did also manual tests on my local with all three drivers

cc: @tonerdo @petli


src/coverlet.msbuild.tasks/CoverageResultTask.cs
```diff
@@ -14,62 +14,27 @@ namespace Coverlet.MSbuild.Tasks
...
-        private ITaskItem _instrumenterState;
         private MSBuildLogger _logger;
 
         [Required]
-        public string Output
-        {
-            get { return _output; }
-            set { _output = value; }
-        }
+        public string Output { get; set; }
```
tonerdo, line 20: One reason I use explicit backing fields is to prevent the overhead of frequent method calls. It's probably bad for performance

src/coverlet.msbuild.tasks/InstrumentationTask.cs
```diff
@@ -15,86 +15,33 @@ namespace Coverlet.MSbuild.Tasks
...
-        private ITaskItem _instrumenterState;
         private readonly MSBuildLogger _logger;
 
         [Required]
-        public string Path
-        {
-            get { return _path; }
-            set { _path = value; }
-        }
+        public string Path { get; set; }
```
tonerdo, line 21: Same on using explicit backing fields

src/coverlet.msbuild.tasks/CoverageResultTask.cs
MarcoRossignoli: I think that autoprops is inlined by jitter, take a look here https://sharplab.io/#v2:C4LghgzgtgPgAgJgIwFgBQcDMACR2DC2A3utmbjvgBQCUp5Ja5z2ACgE4D2ADhNgLzYkABgDc9MgF8JFbAEsAdsDZdexAOYBTYKIjbR0tIfTosuBNgCy6Rsxlm4AFmwAVTRGC0ZtlmUIBjAWwFTQB3AlpxJl9cJABOKn8AOg4eCBoo5kk7I3QgA=

MarcoRossignoli: @jkotas, sorry to bother you, can you confirm that autoprops are inlined by jitter?

tonerdo: I was referring more to the general case because we're targeting .NET standard. Inlining of auto props is an implementation detail of the JIT, at least I don't recall it being in the CIL spec. Since we're only supporting Coverlet running on .NET Core, this is probably fine

MarcoRossignoli: > Do we have any plans to support other runtimes beyond .NET Core?

We support .NET Framework at the moment(not yet on collectors but I'm waiting some answer from vstest team). In this case(inlining apart) I would prefer clean syntax over perf...I mean this code is called only one time for asm, we spend a lot of time on IL instrumentation, btw I won't oppose if you prefer custom back field, let me know!It was instinctive from my side 😄

jkotas: Any runtime where performance matters will inline auto-properties just fine. It is certainly the case for both .NET Framework and .NET Core.

If you wanted to optimize for hypothetical runtimes that are missing basic optimizations, your code would look very unnatural. You would want to e.g. prefer public fields over properties, minimize number of methods and making methods as big as possible, etc. It does not sounds like a good idea.

PR #932, opened 2020-08-25
Add `CoverletReport` MSBuild item in the CoverageResultTask MSBuild task

0xced:
So that the coverlet reports can be easily used by other MSBuild tasks. For example, using [ReportGenerator](https://github.com/danielpalme/ReportGenerator#usage--command-line-parameters) to generate an html coverage report.

```xml
<Target Name="GenerateHtmlCoverageReport" AfterTargets="GenerateCoverageResultAfterTest">
  <ReportGenerator ReportFiles="@(CoverletReport)" TargetDirectory="../html-coverage-report" />
</Target>
```

### DIFF
--- a/src/coverlet.msbuild.tasks/CoverageResultTask.cs
+++ b/src/coverlet.msbuild.tasks/CoverageResultTask.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Text;
@@ -8,6 +9,7 @@ using Coverlet.Core.Abstractions;
 using Coverlet.Core.Enums;
 using Coverlet.Core.Reporters;
 using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
 using Microsoft.Extensions.DependencyInjection;
 
 namespace Coverlet.MSbuild.Tasks
@@ -35,6 +37,9 @@ namespace Coverlet.MSbuild.Tasks
         public ITaskItem InstrumenterState { get; set; }
 
         public string CoverletMultiTargetFrameworksCurrentTFM { get; set; }
+
+        [Output]
+        public ITaskItem[] ReportItems { get; set; }
 
         public CoverageResultTask()
         {
@@ -87,6 +92,7 @@ namespace Coverlet.MSbuild.Tasks
                 }
 
                 var formats = OutputFormat.Split(',');
+                var coverageReportPaths = new List<ITaskItem>(formats.Length);
                 foreach (var format in formats)
                 {
                     var reporter = new ReporterFactory(format).CreateReporter();
@@ -110,9 +116,13 @@ namespace Coverlet.MSbuild.Tasks
                                                                 fileSystem,
                                                                 ServiceProvider.GetService<IConsole>(),
                                                                 result);
-                        writer.WriteReport();
+                        var path = writer.WriteReport();
+                        var metadata = new Dictionary<string, string> { ["Format"] = format };
+                        coverageReportPaths.Add(new TaskItem(path, metadata));
                     }
                 }
+
+                ReportItems = coverageReportPaths.ToArray();
 
                 var thresholdTypeFlags = ThresholdTypeFlags.None;
                 var thresholdStat = ThresholdStatistic.Minimum;
--- a/src/coverlet.msbuild.tasks/coverlet.msbuild.targets
+++ b/src/coverlet.msbuild.tasks/coverlet.msbuild.targets
@@ -66,7 +66,9 @@
       ThresholdType="$(ThresholdType)"
       ThresholdStat="$(ThresholdStat)"
       InstrumenterState="$(InstrumenterState)"
-      CoverletMultiTargetFrameworksCurrentTFM="$(_coverletMultiTargetFrameworksCurrentTFM)" />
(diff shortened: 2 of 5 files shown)


MarcoRossignoli:
Thanks I'll take a look asap!

0xced:
I just rebased on master, I hope you get the chance to have a look at it.

MarcoRossignoli (approved):
LGTM @tonerdo seems a good addition for msbuild driver, can you double check?

src/coverlet.msbuild.tasks/coverlet.msbuild.targets
```diff
@@ -66,7 +66,9 @@
       ThresholdType="$(ThresholdType)"
       ThresholdStat="$(ThresholdStat)"
       InstrumenterState="$(InstrumenterState)"
-      CoverletMultiTargetFrameworksCurrentTFM="$(_coverletMultiTargetFrameworksCurrentTFM)" />
+      CoverletMultiTargetFrameworksCurrentTFM="$(_coverletMultiTargetFrameworksCurrentTFM)">
+      <Output TaskParameter="ReportItems" ItemName="CoverletReport" />
```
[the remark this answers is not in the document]
0xced: I thought about it too but decided to go with the singular spelling because basically all MSBuild [item elements](https://docs.microsoft.com/en-us/visualstudio/msbuild/item-element-msbuild) have a singular name: `ProjectReference`, `PackageReference`, `None`, `Compile`, `EmbeddedResource`, `Service` etc.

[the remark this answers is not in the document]
MarcoRossignoli: ok I agree

test/coverlet.core.tests/Reporters/Reporters.cs
```diff
@@ -30,13 +30,6 @@ public class Reporters
         public void Msbuild_ReportWriter(string coverletMultiTargetFrameworksCurrentTFM, string coverletOutput, string reportFormat, string expectedFileName)
```
tonerdo, line 30: Maybe add a new test instead of changing this one

src/coverlet.msbuild.tasks/CoverageResultTask.cs
```diff
@@ -110,10 +116,14 @@ public override bool Execute()
                                                                 fileSystem,
                                                                 ServiceProvider.GetService<IConsole>(),
                                                                 result);
-                        writer.WriteReport();
+                        var path = writer.WriteReport();
+                        var metadata = new Dictionary<string, string> { ["Format"] = format };
+                        coverageReportPaths.Add(new TaskItem(path, metadata));
```
tonerdo, line 121: Please update msbuild documentation with the format of the output and an example of how it can be used

test/coverlet.core.tests/Reporters/Reporters.cs
0xced: I'm not sure what new test should be written. The purpose of the `Msbuild_ReportWriter` test is to ensure a proper report path. The change I introduced, i.e. returning the path of the report in the `WriteReport` method instead of discarding it just allowed to simplify that test by not intercepting the `WriteAllText` call and use the returned value.

src/coverlet.msbuild.tasks/CoverageResultTask.cs
0xced: I already added a new paragraph in MSBuildIntegration.md and provided an example of how it can be used with ReportGenerator. Is there something missing or that I should clarify?

test/coverlet.core.tests/Reporters/Reporters.cs
MarcoRossignoli: @tonerdo I think it's ok the tests is for `ReportWriter` and return path check is in the context of test.

src/coverlet.msbuild.tasks/CoverageResultTask.cs
MarcoRossignoli: @tonerdo you mean this https://github.com/coverlet-coverage/coverlet/pull/932/files#diff-318e71585840883ebf9a0b9a112c5ea3?

tonerdo: My bad, missed that bit

tonerdo:
Thanks @0xced

0xced:
Excellent, thanks for merging! Do you have an estimate when a new `coverlet.msbuild` NuGet package will ship with those changes? Now that the documentation is merged, it could lead to confusion since the current package (2.9.0) does not support the `CoverletReport` MSBuild item yet.

MarcoRossignoli:
Not yet...but we're close I think, we need some more PR to align versions(we're changing schema version) and after we can deploy new version, we've one not easy coverage issue with new switch pattern https://github.com/coverlet-coverage/coverlet/issues/936  and we need to understand if deploy a version and fix after or not(I don't know atm how much time to solve, bit busy period for me)

mikes-gh:
Just got hit by this as new user.  I presume this is not released.  I struggled with the docs for ages

MarcoRossignoli:
Yep is not released yet, we missed to add this to the changelog as unreleased, we apologize.

0xced:
Released since version 3.0.0. 🎉

Issue #907, opened 2020-07-21
Incorrect branch coverage with await ValueTask


a-jackson:
Awaiting a ValueTask shows an uncovered branch. I made a test project to show the difference between a ValueTask and a Task.
![image](https://user-images.githubusercontent.com/2199309/88040280-936d9800-cb40-11ea-8fc9-9a0bf236bff8.png)
Task is not showing any branches while ValueTask is showing 2 branches and one of the uncovered.

The source of the test project is [here](https://github.com/a-jackson/CoverletValueTaskCoverageIssue), I then used the command `dotnet test --collect:"XPlat Code Coverage" Tests` and used report generator to get the above image.

I think the issue is that `"System.Runtime.CompilerServices.ValueTaskAwaiter"` needs adding to `SkipIsCompleteAwaiters` along with `"System.Runtime.CompilerServices.TaskAwaiter"` in `CecilSymbolHelper` but I don't really understand half that code to be confident enough to make a PR.

MarcoRossignoli:
Thanks for reporting this!

alexthornton1:
I took a crack at fixing this, based on the suggestion by @a-jackson, and that seems to do the trick.

MarcoRossignoli:
closed in https://github.com/coverlet-coverage/coverlet/pull/949

Issue #914, opened 2020-07-29
Incorrect branch coverage with await using


itsurkan:
If use 'using' statement only branch coverate is 100%
![image](https://user-images.githubusercontent.com/6732323/88797872-af121780-d1ac-11ea-9c1e-4aef9332187f.png)

If use 'await using' branch coverage is 50%
![image](https://user-images.githubusercontent.com/6732323/88797975-d10b9a00-d1ac-11ea-9467-d81724a099c4.png)
But they should be the same

MarcoRossignoli:
Which version are you using?

alexthornton1:
Out of curiosity, I checked whether my attempted fix of #907 would also address this.  It appears the answer to that is "no".  I started with this example:

```
public class AwaitUsing
{
    async public ValueTask AsyncAwait()
    {
        await using (var ms = new MemoryStream(System.Text.Encoding.ASCII.GetBytes("abc")))
        {
        }
    }
}
```

I tried this test (similar to what I added in #907, according to patterns of what was already there):

```
[Fact]
public void GetBranchPoints_IgnoresBranchesIn_AwaitUsing()
{
    // arrange
    var nestedName = typeof(AwaitUsing).GetNestedTypes(BindingFlags.NonPublic).First().Name;
    var type = _module.Types.FirstOrDefault(x => x.FullName == typeof(AwaitUsing).FullName);
    var nestedType = type.NestedTypes.FirstOrDefault(x => x.FullName.EndsWith(nestedName));
    var method = nestedType.Methods.First(x => x.FullName.EndsWith("::MoveNext()"));

    // act
    var points = _cecilSymbolHelper.GetBranchPoints(method);

    // assert
    Assert.Empty(points);
}
```

Without my #907 fix, there were eight branch points found.  With my #907 fix, there were six branch points found.

So, at the very least, we know these are separate issues.  Whether I'll be able to figure out what the issue _is_, well, that's a different story, but I'm just finding my way around this code base for the first time.

alexthornton1:
I took a crack at this one today, too.  The #914 fix depends on the #907 fix, but I'll create a separate pull request for the #914 fix.  (I can close the pull request for #907 and combine them into one, if you prefer.)

alexthornton1:
Working in Debug mode, but not Release mode.  Live and learn!  I'll investigate.

alexthornton1:
Fixed (I hope!) in 61e0d9f.

Pankraty:
[...]

PR #949, opened 2020-09-12
Coverage for awaiting ValueTasks

alexthornton1:
For issue #907, first, I added a failing unit test, in two parts:

* A new AsyncAwaitValueTaskStateMachine class in Samples.cs
  in coverlet.core.tests/Samples, in line with the existing
  AsyncAwaitStateMachine class, but returning a completed
  ValueTask instead of a completed Task.

* A new test in CecilSymbolHelperTests.cs in
  coverlet.core.tests/Symbols that expects the "await" in
  that sample class not to have a branch in it.

After it failed, I updated CecilSymbolHelper to include
get_IsCompleted from System.Runtime.CompilerServices.ValueTaskAwaiter,
as suggested by @a-jackson in issue #907.  The test passed.

One open question is whether it might be worth adding a
ValueTask-based analogue for the Instrumentation.AsyncAwait.cs
samples.

### DIFF
--- a/src/coverlet.core/Symbols/CecilSymbolHelper.cs
+++ b/src/coverlet.core/Symbols/CecilSymbolHelper.cs
@@ -172,6 +172,7 @@ namespace Coverlet.Core.Symbols
                     operand.Name == "get_IsCompleted" &&
                     (
                         operand.DeclaringType.FullName.StartsWith("System.Runtime.CompilerServices.TaskAwaiter") ||
+                        operand.DeclaringType.FullName.StartsWith("System.Runtime.CompilerServices.ValueTaskAwaiter") ||
                         operand.DeclaringType.FullName.StartsWith("System.Runtime.CompilerServices.ConfiguredTaskAwaitable") ||
                         operand.DeclaringType.FullName.StartsWith("System.Runtime.CompilerServices.ConfiguredValueTaskAwaitable")
                     )
--- a/test/coverlet.core.tests/Coverage/CoverageTests.AsyncAwaitValueTask.cs
+++ b/test/coverlet.core.tests/Coverage/CoverageTests.AsyncAwaitValueTask.cs
@@ -1,0 +1,70 @@
+using System.IO;
+using System.Threading.Tasks;
+
+using Coverlet.Core.Samples.Tests;
+using Coverlet.Tests.Xunit.Extensions;
+using Xunit;
+
+namespace Coverlet.Core.Tests
+{
+    public partial class CoverageTests
+    {
+        [Fact]
+        public void AsyncAwaitWithValueTask()
+        {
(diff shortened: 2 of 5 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/coverlet-coverage/coverlet?pullRequest=949) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/coverlet-coverage/coverlet?pullRequest=949) before we can accept your contribution.<br/><br/>:x: alexthornton1 [sign now](https://cla.dotnetfoundation.org/coverlet-coverage/coverlet?pullRequest=949)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/coverlet-coverage/coverlet?pullRequest=949) it.</sub>

alexthornton1:
Also included in this branch is an attempted fix for issue #914, which was partly dependent on the fix for issue #907, anyway.

alexthornton1:
So, it turns out the fix for #914 works nicely in the Debug configuration, but not the Release configuration.  The compiler-generated IL for `await using` is different in Release mode, so I'll need to adjust for that.

MarcoRossignoli:
Thanks for looking into the rabbit hole 😃  for tests please add also full coverage assertion like https://github.com/coverlet-coverage/coverlet/blob/master/test/coverlet.core.tests/Coverage/CoverageTests.SelectionStatements.cs#L18

You can use `FunctionExecutor.RunInProcess` to run/attach live during debugging and `result.GenerateReport(show:true)` https://github.com/coverlet-coverage/coverlet/blob/5ec636bb16642cba5ca8116cea4df51de78a4e35/test/coverlet.core.tests/Coverage/InstrumenterHelper.Assertions.cs#L23 to get live report generator check

alexthornton1:
No worries.  I had thought the work might be done, but I also thought there might be more to the story.  I'll work on adding some higher-level tests and see where that leads.  For now, I've changed the name of the PR to begin with [WIP].  I'll change it back when I think I'm done and we'll circle back again.  Thanks for the response.

alexthornton1:
Let's pull these apart.  Once I started adding coverage tests, the #907 fix worked fine, but #914 is a gnarlier beast than I had hoped.  I'll still aim to attack it, but it seems clear it'll take longer, so no need to hold up the #907 fix -- which is what brought me here in the first place -- in the meantime.  I moved #914 to another branch and will create a PR when I get a fuller solution together.

So this PR is now just a #907 fix with coverage testing.  Happy to add whatever else is necessary; this has been fun so far.

MarcoRossignoli (approved):
LGTM Thanks a lot!

PR #952, opened 2020-09-18
A couple of code quality fixes

UselessToucan:
1. `endFilter.OpCode` was accessed before verifying that `endFilter` is not null.
2. As far as I see, `LoadSourceRootMapping` always returns a value. I.e. `_sourceRootMapping` can't be null.


MarcoRossignoli (approved):
Thanks!

PR #991, opened 2020-11-14
Allow standalone coverlet usage for integration/end-to-end tests using .NET tool driver

MarcoRossignoli:
contributes https://github.com/coverlet-coverage/coverlet/issues/781

This is an attempt to support end-to-end coverage scenario.
In this case at the end of tests all generated reports should be moved to a folder and merged with report tools like report generator.

@jakubch1 I know that the workflow you described is a bit different(I call it "cold coverage") but at the moment this is the best we can do.

FYI @AbhitejJohn @pavelhorak

### DIFF
--- a/Documentation/GlobalTool.md
+++ b/Documentation/GlobalTool.md
@@ -9,12 +9,12 @@ coverlet --help
 The current options are (output of `coverlet --help`):
 
 ```bash
-Cross platform .NET Core code coverage tool 1.0.0.0
+Cross platform .NET Core code coverage tool 3.0.0.0
 
-Usage: coverlet [arguments] [options]
+Usage: coverlet [options] <<ASSEMBLY> or <APP DIR>>
 
 Arguments:
-  <ASSEMBLY>  Path to the test assembly.
+  <ASSEMBLY> or <APP DIR>      Path to the test assembly or application directory.
 
 Options:
   -h|--help                    Show help information
@@ -23,21 +23,21 @@ Options:
   -a|--targetargs              Arguments to be passed to the test runner.
   -o|--output                  Output of the generated coverage report
   -v|--verbosity               Sets the verbosity level of the command. Allowed values are quiet, minimal, normal, detailed.
-  -f|--format                  Format of the generated coverage report[multiple value].
+  -f|--format                  Format of the generated coverage report.
   --threshold                  Exits with error if the coverage % is below value.
-  --threshold-type             Coverage type to apply the threshold to[multiple value].
+  --threshold-type             Coverage type to apply the threshold to.
   --threshold-stat             Coverage statistic used to enforce the threshold value.
-  --exclude                    Filter expressions to exclude specific modules and types[multiple value].
-  --include                    Filter expressions to include specific modules and types[multiple value].
-  --include-directory          Include directories containing additional assemblies to be instrumented[multiple value].
-  --exclude-by-file            Glob patterns specifying source files to exclude[multiple value].
-  --exclude-by-attribute       Attributes to exclude from code coverage[multiple value].
+  --exclude                    Filter expressions to exclude specific modules and types.
+  --include                    Filter expressions to include only specific modules and types.
+  --exclude-by-file            Glob patterns specifying source files to exclude.
+  --include-directory          Include directories containing additional assemblies to be instrumented.
+  --exclude-by-attribute       Attributes to exclude from code coverage.
   --include-test-assembly      Specifies whether to report code coverage of the test assembly.
-  --single-hit                 Specifies whether to limit code coverage hit reporting to a single hit for each location.
+  --single-hit                 Specifies whether to limit code coverage hit reporting to a single hit for each location
+  --skipautoprops              Neither track nor record auto-implemented properties.
   --merge-with                 Path to existing coverage result to merge.
   --use-source-link            Specifies whether to use SourceLink URIs in place of file system paths.
-  --skipautoprops              Neither track nor record auto-implemented properties.
-  --does-not-return-attribute  Attributes that mark methods that do not return[multiple value].
+  --does-not-return-attribute  Attributes that mark methods that do not return.
 ```
 
 NB. For a [multiple value] options you have to specify values multiple times i.e.
@@ -59,6 +59,22 @@ coverlet /path/to/test-assembly.dll --target "dotnet" --targetargs "test /path/t
 After the above command is run, a `coverage.json` file containing the results will be generated in the directory the `coverlet` command was run. A summary of the results will also be displayed in the terminal.
 
 _Note: The `--no-build` flag is specified so that the `/path/to/test-assembly.dll` isn't rebuilt_
+
+## Code Coverage for integration tests and end-to-end tests.
+
+Sometimes there are tests that doesn't use classic tests framework like XUnit/NUnit and so on.  
+For instance suppose to have some integration end-to-end tests where you do some calls to a backend web site with Selenium and at the end do assertions on database data. In this use case there is no test engine involved but simple script to startup tests and do evaluation at the end. 
+To gather the coverage for this kind of tests you can use Coverlet .NET tool and startup tests through our driver.  
+For instance suppose to have a folder `/application` where you've built your project and where there is the app entry point `myapp` file.
+You can use our tool to startup the app and gather live coverage:
+bash
+```
+coverlet "/application" --target "/application/myapp"
+```
+Coverlet will instrument all assemblies inside de `application` folder and after it will start the `myapp` application. Finally at shutdown of your app it will generate the coverage report. You can use all parameters available to customize the report generation.  
+
+**CAVEAT: today  Coverlet relies on `AppDomain.CurrentDomain.ProcessExit` and `AppDomain.CurrentDomain.DomainUnload` to register coverage hits on file system for the accounting, due to this you need to ensure a gracefully process shutdown. If you'll kill the process you won't get correct coverage report.**
+
 
 ## Coverage Output
 
--- a/Documentation/GlobalTool.md
+++ b/Documentation/GlobalTool.md
@@ -11,10 +11,10 @@ The current options are (output of `coverlet --help`):
 ```bash
 Cross platform .NET Core code coverage tool 3.0.0.0
 
-Usage: coverlet [options] <<ASSEMBLY> or <APP DIR>>
+Usage: coverlet [arguments] [options]
 
 Arguments:
-  <ASSEMBLY> or <APP DIR>      Path to the test assembly or application directory.
+  <ASSEMBLY|DIRECTORY>         Path to the test assembly or application directory.
 
 Options:
   -h|--help                    Show help information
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -29,7 +29,7 @@ Coverlet can be used through three different *drivers*
 
 * VSTest engine integration
 * MSBuild task integration
-* As a .NET Global tool  
+* As a .NET Global tool(used also for integration test, end-to-end tests)
```
tonerdo, line 32:
```suggestion
* As a .NET Global tool (supports standalone integration tests)
```

test/coverlet.core.tests/Helpers/InstrumentationHelperTests.cs
```diff
@@ -220,11 +224,21 @@ public void TestIncludeDirectories()
             Assert.Single(currentDirModules);
             Assert.Equal("coverlet.msbuild.tasks.dll", Path.GetFileName(currentDirModules[0]));
 
-            var moreThanOneDirectory = _instrumentationHelper.GetCoverableModules(module, new string[] { newDir2.FullName }, false);
+            var moreThanOneDirectory = _instrumentationHelper.GetCoverableModules(module, new string[] { newDir2.FullName }, false)
+                                       .OrderBy(f => f).ToArray();
```
tonerdo, line 228: Fix indentation:
```csharp
var moreThanOneDirectory = _instrumentationHelper
                               .GetCoverableModules(module, new string[] { newDir2.FullName }, false)
                               .OrderBy(f => f).ToArray();
```

test/coverlet.core.tests/Helpers/InstrumentationHelperTests.cs
```diff
@@ -220,11 +224,21 @@ public void TestIncludeDirectories()
...
+
             Assert.Equal(2, moreThanOneDirectory.Length);
             Assert.Equal("coverlet.msbuild.tasks.dll", Path.GetFileName(moreThanOneDirectory[0]));
             Assert.Equal("coverlet.core.dll", Path.GetFileName(moreThanOneDirectory[1]));
 
+            var moreThanOneDirectoryPlusTestAssembly = _instrumentationHelper.GetCoverableModules(module, new string[] { newDir2.FullName }, true)
+                                                       .OrderBy(f => f).ToArray();
+
```
tonerdo, line 236: Ditto on indentation

tonerdo, line 236: Ditto on indentation

test/coverlet.integration.tests/DotnetTool.cs
```diff
@@ -29,5 +29,20 @@ public void DotnetTool()
             Assert.Contains("Test Run Successful.", standardOutput);
             AssertCoverage(clonedTemplateProject, standardOutput: standardOutput);
         }
+
+        [ConditionalFact]
+        [SkipOnOS(OS.Linux)]
+        [SkipOnOS(OS.MacOS)]
```
tonerdo, line 35: Curious. Why do we only support Windows?

Documentation/GlobalTool.md
```diff
@@ -9,12 +9,12 @@ coverlet --help
 The current options are (output of `coverlet --help`):
 
 ```bash
-Cross platform .NET Core code coverage tool 1.0.0.0
+Cross platform .NET Core code coverage tool 3.0.0.0
 
-Usage: coverlet [arguments] [options]
+Usage: coverlet [options] <<ASSEMBLY> or <APP DIR>>
```
tonerdo, line 14:
```suggestion
Usage: coverlet [arguments] [options]
```

Documentation/GlobalTool.md
```diff
@@ -60,6 +60,22 @@ After the above command is run, a `coverage.json` file containing the results wi
 
 _Note: The `--no-build` flag is specified so that the `/path/to/test-assembly.dll` isn't rebuilt_
 
+## Code Coverage for integration tests and end-to-end tests.
```
tonerdo, line 63:
```suggestion
## Code Coverage for standalone integration tests.
```

Documentation/GlobalTool.md
```diff
@@ -9,12 +9,12 @@ coverlet --help
...
 ```bash
-Cross platform .NET Core code coverage tool 1.0.0.0
+Cross platform .NET Core code coverage tool 3.0.0.0
 
-Usage: coverlet [arguments] [options]
+Usage: coverlet [options] <<ASSEMBLY> or <APP DIR>>
 
 Arguments:
-  <ASSEMBLY>  Path to the test assembly.
+  <ASSEMBLY> or <APP DIR>      Path to the test assembly or application directory.
```
tonerdo, line 17:
```suggestion
  <ASSEMBLY|DIRECTORY>         Path to the test assembly or application directory.
```

src/coverlet.core/Helpers/InstrumentationHelper.cs
```diff
@@ -31,11 +31,14 @@ public InstrumentationHelper(IProcessExitHandler processExitHandler, IRetryHelpe
...
 
-        public string[] GetCoverableModules(string module, string[] directories, bool includeTestAssembly)
+        public string[] GetCoverableModules(string moduleOrAppDirectory, string[] directories, bool includeTestAssembly)
         {
             Debug.Assert(directories != null);
+            Debug.Assert(moduleOrAppDirectory != null);
 
-            string moduleDirectory = Path.GetDirectoryName(module);
+            bool isAppDirectory = !File.Exists(moduleOrAppDirectory) && Directory.Exists(moduleOrAppDirectory);
+
```
tonerdo, line 40: Nit: Remove blank line

Documentation/GlobalTool.md
```diff
@@ -60,6 +60,22 @@ After the above command is run, a `coverage.json` file containing the results wi
...
+You can use our tool to startup the app and gather live coverage:
+bash
+```
+coverlet "/application" --target "/application/myapp"
+```
+Coverlet will instrument all assemblies inside de `application` folder and after it will start the `myapp` application. Finally at shutdown of your app it will generate the coverage report. You can use all parameters available to customize the report generation.  
+
+**CAVEAT: today  Coverlet relies on `AppDomain.CurrentDomain.ProcessExit` and `AppDomain.CurrentDomain.DomainUnload` to register coverage hits on file system for the accounting, due to this you need to ensure a gracefully process shutdown. If you'll kill the process you won't get correct coverage report.**
+
+
```
tonerdo, line 78: Sometimes, there are tests that doesn't use regular unit test frameworks like xunit. You may find yourself in a situation where your tests are driven by a custom executable/script, which when run, could do anything from making API calls to driving Selenium.

As an example, suppose you have a folder `/integrationtest` which contains said executable (lets call it `runner.exe`) and everything it needs to successfully execute. You can use our tool to startup the executable and gather live coverage:

```bash
coverlet "/integrationtest" --target "/application/runner.exe"
```

Coverlet will first instrument all .NET assemblies within the `integrationtests` folder, after which it will execute `runner.exe`. Finally, at shutdown of your `runner.exe`, it will generate the coverage report. You can use all parameters available to customize the report generation. Coverage results will be generated once `runner.exe` exits. You can use all parameters available to customize the report generation.

_Note: Today, Coverlet relies on `AppDomain.CurrentDomain.ProcessExit` and `AppDomain.CurrentDomain.DomainUnload` to record hits to the filesystem, as a result, you need to ensure a graceful process shutdown. Forcefully, killing the process will result in an incomplete coverage report._

src/coverlet.core/Helpers/InstrumentationHelper.cs
```diff
@@ -31,11 +31,14 @@ public InstrumentationHelper(IProcessExitHandler processExitHandler, IRetryHelpe
...
+            string moduleDirectory = isAppDirectory ? moduleOrAppDirectory : Path.GetDirectoryName(moduleOrAppDirectory);
```
tonerdo, line 41: Nit: New line

tonerdo (changes requested):

test/coverlet.integration.tests/DotnetTool.cs
MarcoRossignoli: we've an open issue for that, never had time to investigate https://github.com/coverlet-coverage/coverlet/issues/660

petli (commented):
Looks good! 

A thought about that the test subject process must be shut down somehow: could this be tricky to achieve in CI environments without some help from the coverlet test process? I'm thinking about CI scripts like this, in `sh` syntax with a hypothetical new flag:

```sh
coverlet "/integrationtest" --target "/application/runner.exe" --terminate-test-on-kill &
coverlet_pid=$!
./my_integration_test_suite.sh
kill -TERM $coverlet_pid
wait $coverlet_pid
```

Based on what I see in e.g. https://github.com/dotnet/coreclr/pull/4309 this would require adding a `ProcessExit()` handler in `coverlet` that calls `process.Kill()` on the test subject. There would be timing issues in this though, since now there's two `ProcessExit()` events that must happend quickly, which may make it untenable. 

Perhaps the recommendation will have to be that the integration test suite have a way to tell the test subject to shut down.

MarcoRossignoli:
> Perhaps the recommendation will have to be that the integration test suite have a way to tell the test subject to shut down.

Added this in guide https://github.com/coverlet-coverage/coverlet/pull/991/files#diff-0ef674889bda556582dddf36bb4f1f71254f7dcfb27a5aeff2f4e5d379582cb1R75 do you think is not enough?

Another idea could be provide a code snipped to use to deterministically flush coverage with some reflection like , `CoverletCoverage.FlushCoverage()` and looking for tracker.unload call. I mean if a gracefully shutdown is not possible.

petli:
> > Perhaps the recommendation will have to be that the integration test suite have a way to tell the test subject to shut down.
> 
> Added this in guide https://github.com/coverlet-coverage/coverlet/pull/991/files#diff-0ef674889bda556582dddf36bb4f1f71254f7dcfb27a5aeff2f4e5d379582cb1R75 do you think is not enough?

True, it's difficult to say anything more specific than that given how diverse the test subjects can be.

> Another idea could be provide a code snipped to use to deterministically flush coverage with some reflection like , `CoverletCoverage.FlushCoverage()` and looking for tracker.unload call. I mean if a gracefully shutdown is not possible.

That's a really good idea! If an integration test anyway have to rely on adding a special endpoint or method to shut down, it would help to ensure that all the hit files are written out too given the limitations on `ProcessExit()`.

> > A thought about that the test subject process must be shut down somehow: could this be tricky to achieve in CI environments without some help from the coverlet test process?
> 
> I think that a test suite has got a workflow like start server -> do tests -> asserts -> stop server -> stop test...do you think that's possible keep a process in hang at the end?

I was thinking it may be a bit fiddly to find the server process to stop, when it is run from coverlet rather than the CI script. This may be very Unix-flavoured of me though, and it would be solved otherwise in a Windows CI environment.

MarcoRossignoli:
> I was thinking it may be a bit fiddly to find the server process to stop, when it is run from coverlet rather than the CI script. 

Understood what you mean...if I should do that I would use a mutex to sync test/server...but it's true...depends on how the tests was thought.

We end always at same point, we should use shared memory in a client/server architecture to avoid all problems 😄

petli:
> We end always at same point, we should use shared memory in a client/server architecture to avoid all problems 😄

Indeed :)

petli (approved):

MarcoRossignoli:
@tonerdo PTAL

src/coverlet.console/Program.cs
```diff
@@ -44,7 +44,7 @@ static int Main(string[] args)
             app.VersionOption("-v|--version", GetAssemblyVersion());
             int exitCode = (int)CommandExitCodes.Success;
 
-            CommandArgument module = app.Argument("<ASSEMBLY>", "Path to the test assembly.");
+            CommandArgument moduleOrAppDirectory = app.Argument("<ASSEMBLY> or <APP DIR>", "Path to the test assembly or application directory.");
```
tonerdo, line 47:
```suggestion
            CommandArgument moduleOrAppDirectory = app.Argument("<ASSEMBLY|DIRECTORY>", "Path to the test assembly or application directory.");
```

tonerdo (approved):
LGTM! Modulo one comment

jeremyLJ:
The command below on Linux also has the same issue described in https://github.com/coverlet-coverage/coverlet/issues/974

still failed after try try with `collectors`

```
coverlet TestColverlet/TestColverlet/bin/Debug/netcoreapp3.1/TestColverlet.dll  --target "dotnet" --targetargs "TestColverlet/TestColverlet/bin/Debug/netcoreapp3.1/TestColverlet.dll" --format cobertura --include-test-assembly --verbosity:detailed
```

MarcoRossignoli:
@jeremyLJ can you open a new issue? The reason for this kind of fails are a lot so we need to go deeper on your use case.

jeremyLJ:
@MarcoRossignoli thanks for the quick feedback.
what I met was the exact same issue as https://github.com/coverlet-coverage/coverlet/blob/master/Documentation/KnownIssues.md#6-code-coverage-returns-nan

sorry for the randomization...

this enhancement is pretty cool~

PR #1000, opened 2020-11-24
Escape curly braces in file names when logging

### LINKED ISSUE #998: Unable to instrument module because : Input string was not in a correct format
Using VSTest integration, we generate coverage reports for our primarily netstandard2.0-targeting assemblies. However, for a subset of assemblies (seemingly the first in a directory listing per service directory, e.g. in https://github.com/azure/azure-sdk-for-net: Azure.Core under sdk/core and Azure.Security.KeyVault.Administration in sdk/keyvault) diagnostic lots on the CI (Azure Pipelines) show something like:

```
TpTrace Warning: 0 : 1076, 1, 2020/11/21, 02:02:34.280, 5338411430, datacollector.dll, [coverlet]Unable to instrument module: 'D:\a\1\s\artifacts\bin\Azure.Security.KeyVault.Administration.Tests\Debug\netcoreapp2.1\Azure.Security.KeyVault.Administration.dll' because : Input string was not in a correct format.
```

This is from build https://dev.azure.com/azure-sdk/public/_build/results?buildId=629023&view=logs&j=1e00682f-6081-5348-8da2-7d601664b2e3&t=4f3e1a81-cacf-5ad2-6e75-5049507bfdad (should be retained for ~30 days).

Running the same commands (just different source path) locally yields the expected results, e.g. Azure.Core is covered, as is Azure.Security.KeyVault.Administration.

Looking at other issues here, it seems the `FormatException` that's being thrown (based on the error message) is most often from targeting < netstandard2.0, but our baseline already is netstandard2.0. Our common root dependency, Azure.Core, targets net461 and netstandard2.0 (for those few rare classes in net461 that aren't actually in netstandard2.0) but, either way, should work according to other issue conversations herein.

We are using the latest GA for Microsoft.NET.Test.Sdk, 16.8.0.

Update: setting `/p:ContinuousIntegrationBuild=true` on the local repro (same command line as CI otherwise; that property is set via targets in CI based on `$TF_BUILD`) repro'd locally. I'll attempt to debug into the problem.

heaths:
Fixes #998

### DIFF
--- a/src/coverlet.core/Coverage.cs
+++ b/src/coverlet.core/Coverage.cs
@@ -107,7 +107,7 @@ namespace Coverlet.Core
 
             Array.ForEach(_excludeFilters ?? Array.Empty<string>(), filter => _logger.LogVerbose($"Excluded module filter '{filter}'"));
             Array.ForEach(_includeFilters ?? Array.Empty<string>(), filter => _logger.LogVerbose($"Included module filter '{filter}'"));
-            Array.ForEach(_excludedSourceFiles ?? Array.Empty<string>(), filter => _logger.LogVerbose($"Excluded source files filter '{filter}'"));
+            Array.ForEach(_excludedSourceFiles ?? Array.Empty<string>(), filter => _logger.LogVerbose($"Excluded source files filter '{FileSystem.EscapeFileName(filter)}'"));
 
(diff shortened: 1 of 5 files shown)


MarcoRossignoli (approved):
LGTM thanks!

PR #1004, opened 2020-11-27
Relative paths lcov

phaniva:
Please review this design to see if it addresses the requirement. I changed it so that we can enable relative directory option for all report formats but only changed it for lcov for now.


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/coverlet-coverage/coverlet?pullRequest=1004) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/coverlet-coverage/coverlet?pullRequest=1004) before we can accept your contribution.<br/><br/>:x: phaniva [sign now](https://cla.dotnetfoundation.org/coverlet-coverage/coverlet?pullRequest=1004)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/coverlet-coverage/coverlet?pullRequest=1004) it.</sub>

MarcoRossignoli:
@phaniva please sign the CLA.

I'll take a look asap

phaniva:
@MarcoRossignoli I have signed the CLA. Not sure if you received a notification automatically

MarcoRossignoli:
@phaniva yep I see, bit busy this period, I'll take a look asap.

src/coverlet.core/Helpers/FilePathHelper.cs
```diff
@@ -0,0 +1,93 @@
+﻿using Coverlet.Core.Abstractions;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+
+
+namespace Coverlet.Core.Helpers
+{
+    internal class FilePathHelper : IFilePathHelper
```
MarcoRossignoli, line 10: I think that extract a service is too much for a path helper...I mean we won't inject different implementation there are no different way to get that path from file system.
Should be enough add a new base class `abstract ReporterBase` and add this two method as `protected` and inherit the reporter that needs this logic.

MarcoRossignoli (changes requested):
I like the idea, we can simplify and simply add a base class with this two helpers.

src/coverlet.core/Helpers/FilePathHelper.cs
phaniva: So only the reporter classes that need this behavior will inherit from ReporterBase and rest of them will continue to implement IReporter?
Why not inherit every reporter class from ReporterBase to keep things consistent?

MarcoRossignoli: Ok I won't oppose, the important things is that we have less cose as possible to maintain, let's go for base class inherited by reporters

phaniva: Cool. I am going to close this pr and open a new one with base class design.

PR #1006, opened 2020-11-28
Fix switch pattern

MarcoRossignoli:
fixes https://github.com/coverlet-coverage/coverlet/issues/936

cc: @daveMueller I did some test and skipping branches and breakpoint expression seems to work

image

image

### DIFF
--- a/src/coverlet.core/Coverage.cs
+++ b/src/coverlet.core/Coverage.cs
@@ -396,24 +396,14 @@ namespace Coverlet.Core
                 // Nested ranges win on outermost one
                 foreach (HitCandidate hitCandidate in result.HitCandidates)
                 {
-                    if (hitCandidate.isBranch)
-                    {
-                        continue;
-                    }
-
-                    if (hitCandidate.end == hitCandidate.start)
+                    if (hitCandidate.isBranch || hitCandidate.end == hitCandidate.start)
                     {
                         continue;
                     }
 
                     foreach (HitCandidate hitCandidateToCompare in result.HitCandidates)
                     {
-                        if (hitCandidateToCompare.isBranch)
-                        {
-                            continue;
-                        }
-
-                        if (hitCandidate != hitCandidateToCompare)
+                        if (hitCandidate != hitCandidateToCompare && !hitCandidateToCompare.isBranch)
                         {
                             if (hitCandidateToCompare.start >= hitCandidate.start &&
                                hitCandidateToCompare.end <= hitCandidate.end)
--- a/src/coverlet.core/Coverage.cs
+++ b/src/coverlet.core/Coverage.cs
@@ -422,7 +422,7 @@ namespace Coverlet.Core
                                      i <= (hitCandidateToCompare.end == 0 ? hitCandidateToCompare.start : hitCandidateToCompare.end);
                                      i++)
                                 {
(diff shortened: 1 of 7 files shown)


daveMueller:
Yes that's a good solution. Thanks again to @jakubch1 for the hint. The only thing that doesn't really fit is the hits count, line 36 is actually only hit once.

image

But it's certainly more lightweight than what I did. I mixed this skip and the other algorithm and therefore had to do some changes in the instrumenter ..., definitely more heavy.

MarcoRossignoli:
@daveMueller with the idea of "skip nested lines" works better
image

MarcoRossignoli:
@daveMueller please feel free to review this.

daveMueller:
@MarcoRossignoli Sure I will just need approx. 2 days to find some spare time. I'm really interested in the key differences to my implementation and why I couldn't make I run without touching the ```Intrumenter```. Also I think this might solve another issue I have in mind.

src/coverlet.core/Symbols/CecilSymbolHelper.cs
```diff
@@ -714,6 +729,43 @@ static Instruction GetPreviousNoNopInstruction(Instruction i)
...
+               IL_0001: brtrue.s IL_0004
+               // if (value is int)
+               // sequence point: (line 34, col 9) to (line 40, col 10) in C:\git\coverletfork\test\coverlet.core.tests\Samples\Instrumentation.SelectionStatements.cs
+               IL_0003: nop
+                // sequence point: hidden
+                ...
+             */
+            if (
+                    instruction.OpCode == OpCodes.Ldc_I4 && instruction.Operand is int operandValue && operandValue == 1 &&
+                    instruction.Next?.OpCode == OpCodes.Brtrue &&
```
jakubch1, line 752: To check exactly above pattern you should also check that target of `brtrue.s` is instruction after `nop`

MarcoRossignoli: Uh you're right thanks.

daveMueller: According to the pattern I also checked ```methodDefinition.DebugInformation.GetSequencePoint(instruction.Next?.Next?.Next).IsHidden```. That's also what I observed in the IL that after the ```nop``` instruction a hidden sequence point follows. But I'm not a specialist here and maybe this isn't important.

src/coverlet.core/Coverage.cs
```diff
@@ -392,7 +392,33 @@ private void CalculateCoverage()
...
+                    {
+                        continue;
+                    }
+
+                    foreach (HitCandidate hitCandidateToCompare in result.HitCandidates)
+                    {
+                        if (hitCandidate != hitCandidateToCompare && !hitCandidateToCompare.isBranch)
+                        {
+                            if (hitCandidateToCompare.start >= hitCandidate.start &&
+                               hitCandidateToCompare.end <= hitCandidate.end)
```
daveMueller, line 409: Here is where I made the mistake in my branch. I just had an issue with the range :man_facepalming:. I really like your idea of saving the excluded lines in a ```HashSet<int>``` this removes the overhead of finding the top most :thumbsup:.

PR #1031, opened 2021-01-03
This is not implemented

mikes-gh:
- Save users time my removing from the doc until a new version is released
- See https://github.com/coverlet-coverage/coverlet/pull/932


MarcoRossignoli:
Sorry we missed to add the feature on unreleased section on changelog https://github.com/coverlet-coverage/coverlet/blob/master/Documentation/Changelog.md


Please can you change your PR and update changelog in the Add section?

PR #1060, opened 2021-01-17
Fix multi-line lambda coverage regression

MarcoRossignoli:
closes https://github.com/coverlet-coverage/coverlet/issues/1056

Before

before

After

image

cc: @ulrichb @daveMueller

### DIFF
--- a/src/coverlet.core/Coverage.cs
+++ b/src/coverlet.core/Coverage.cs
@@ -405,8 +405,8 @@ namespace Coverlet.Core
                     {
                         if (hitCandidate != hitCandidateToCompare && !hitCandidateToCompare.isBranch)
                         {
-                            if (hitCandidateToCompare.start >= hitCandidate.start &&
-                               hitCandidateToCompare.end <= hitCandidate.end)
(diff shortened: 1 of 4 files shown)


MarcoRossignoli:
The overlapped lambda accounting is not correct yet...I have some working that fix it but I need to work more on it because there is an ambiguity with C# switch pattern  and I trying to understand how to solve.
BTW is not a regression it's always been incorrect.

SebastienPortier:
Hi !

Thanks for release, the situation is improving on multi lines lambda 👍 . Coverage is ok when having one lambda on multi line, but when chaining multiple lambdas (for example in Fluent synthax way), coverage has still some holes :

One line is OK
Capture d’écran 2021-01-25 122008

One lambda chained multiple line is OK
Capture d’écran 2021-01-25 122149

Multiple lambdas chained multiple lines is KO
Capture d’écran 2021-01-25 122132

MarcoRossignoli:
@SebastienPortier can you open another issue pls?
Is this a regression from 2.9.0?

Issue #1104, opened 2021-03-02
Partially covered `await foreach` statement


meggima:
Hello

In the following code the `await foreach` as well as the `for` and the `yield` statements are marked as partially covered due to the hidden async/await logic of the compiler. Can this be reported as covered by coverlet?

Reproduction class:

```csharp
public class AwaitForeachReproduction
{
    public async Task<int> Execute()
    {
        int sum = 0;

        await foreach (int result in AsyncEnumerable())
        {
            sum += result;
        }

        return sum;
    }

    private async IAsyncEnumerable<int> AsyncEnumerable()
    {
        for(int i = 0; i < 100; i++)
        {
            await Task.Delay(10);
            yield return i;
        }
    }
}
```

Test Case for it:
```csharp
public class AwaitForeachReproductionFixture
{
    [Fact]
    public async Task Execute_ShouldWork()
    {
        // Arrange
        var sut = new AwaitForeachReproduction();

        // Act
        int result = await sut.Execute();

        // Assert
        Assert.Equal(4950, result);
    }
}
```

Reported coverage:
![image](https://user-images.githubusercontent.com/22348611/109637331-f3073980-7b4c-11eb-9e22-d79aa948543f.png)

Coverlet version: 3.0.3

meggima:
You can find the full reproduction code (and also for some other cases) in this repository: https://github.com/meggima/coverlet-reproductions

MarcoRossignoli:
assigned to @alexthornton1

MarcoRossignoli:
Closed in https://github.com/coverlet-coverage/coverlet/pull/1107

ipetrushevskiy:
If you change the _AsyncEnumerable_ method in the original example by changing method like that (code below), it will be partially covered again.

private async IAsyncEnumerable<int> AsyncEnumerable([EnumeratorCancellation] CancellationToken cancellationToken)
{
    for(int i = 0; i < 100; i++)
    {
        await Task.Delay(10, cancellationToken);
        yield return i;
    }
}

PR #1107, opened 2021-03-04
Coverage for "await foreach" loops and compiler-generated async iterators (issue #1104)

alexthornton1:
This is an attempt to fix code coverage for the "await foreach" loop introduced in C# 8, as reported in issue #1104.

The presence of an "await foreach" loop causes for new kinds of branch patterns to be generated in the async state machine.  Three of these are to be eliminated, while the fourth is actually the "should I stay in the loop or not?" branch and is best left in a code coverage report.

CecilSymbolHelper now eliminates the three branch patterns that need to be eliminated.  There are tests both in CecilSymbolHelperTests as well as a new set of coverage tests in CoverageTests.AsyncForeach.cs.


alexthornton1:
On further review, it appears that issue #1104 is about more than just "await foreach".  In particular, this PR fixes "await foreach", but #1104 is also about the code generated for a yield-based method that returns IAsyncEnumerable, which my current PR does not solve.  I'll work on adding that to the mix; these go hand-in-hand.

alexthornton1:
I'm unclear on why there were failures in Windows Debug and Windows Release.  Looking at the raw logs, here's the failure that was reported, which boiled down to an integration test taking longer than five minutes to run:

```
2021-03-06T03:40:56.6123173Z [xUnit.net 00:05:54.17]     Coverlet.Integration.Tests.Msbuild.Test_MultipleTargetFrameworkReport_CoverletOutput_Folder_FileNameWithExtension [FAIL]
2021-03-06T03:40:56.6125752Z   Failed Coverlet.Integration.Tests.Msbuild.Test_MultipleTargetFrameworkReport_CoverletOutput_Folder_FileNameWithExtension [5 m]
2021-03-06T03:40:56.6126854Z   Error Message:
2021-03-06T03:40:56.6128307Z    Command 'dotnet test -c Debug "D:\a\1\s\test\coverlet.integration.tests\bin\Debug\net5.0\0c144e8" /p:CollectCoverage=true /p:Include="[coverletsamplelib.integration.template]*DeepThought" /p:IncludeTestAssembly=true /p:CoverletOutput="D:\a\1\s\test\coverlet.integration.tests\bin\Debug\net5.0\0c144e8"\file.ext' didn't end after 5 minute
2021-03-06T03:40:56.6130099Z   Stack Trace:
2021-03-06T03:40:56.6131191Z      at Coverlet.Integration.Tests.BaseTest.RunCommand(String command, String arguments, String& standardOutput, String& standardError, String workingDirectory) in /_/test/coverlet.integration.tests/BaseTest.cs:line 91
2021-03-06T03:40:56.6132655Z    at Coverlet.Integration.Tests.BaseTest.DotnetCli(String arguments, String& standardOutput, String& standardError, String workingDirectory) in /_/test/coverlet.integration.tests/BaseTest.cs:line 100
2021-03-06T03:40:56.6134080Z    at Coverlet.Integration.Tests.Msbuild.Test_MultipleTargetFrameworkReport_CoverletOutput_Folder_FileNameWithExtension() in /_/test/coverlet.integration.tests/Msbuild.cs:line 182
```

Is this an indication that there's something wrong with my changes, or is this an indication of something else?

I tried running the integration tests locally, but they all fail immediately for a different reason (**Coverlet.Integration.Tests.BaseTest.GetPackageVersion** finding more than one result on this line, so that the call to **Single()** fails:

```
            using Stream pkg = File.OpenRead(Directory.GetFiles($"../../../../../bin/{GetAssemblyBuildConfiguration()}/Packages", filter).Single());
```

I've yet to figure out the right workaround for that to make the integration tests run locally.  Happy to help, but not sure yet where to go from here.

MarcoRossignoli:
Thx @alexanderkozlenko!

cc: @daveMueller for help on review

src/coverlet.core/Symbols/CecilSymbolHelper.cs
```diff
@@ -396,6 +412,256 @@ private bool SkipGeneratedBranchesForExceptionHandlers(MethodDefinition methodDe
...
+            // IL_0131: brfalse.s IL_0196
+
+            if (instruction.OpCode != OpCodes.Brfalse &&
+                instruction.OpCode != OpCodes.Brfalse_S)
+            {
+                return false;
+            }
+
+            if (currentIndex >= 1 &&
+                instructions[currentIndex - 1].OpCode == OpCodes.Ldfld &&
```
MarcoRossignoli, line 468: Can we add the `ldarg` before `ldfld` and `brfalse.s` to be more sure?

image

src/coverlet.core/Symbols/CecilSymbolHelper.cs
```diff
@@ -396,6 +412,256 @@ private bool SkipGeneratedBranchesForExceptionHandlers(MethodDefinition methodDe
...
+            // by "ldfld" of the compiler-generated "<>w__disposeMode" field.
+
+            if (instruction.OpCode != OpCodes.Brfalse &&
+                instruction.OpCode != OpCodes.Brfalse_S)
+            {
+                return false;
+            }
+
+            if (currentIndex >= 1 &&
+                instructions[currentIndex - 1].OpCode == OpCodes.Ldfld &&
```
MarcoRossignoli, line 655: also here enforce with the load of `__disposeMode`

image

MarcoRossignoli (approved):
Some nits but looks georgeous!

MarcoRossignoli:
@alexthornton1 can you attach an image of the resulting report?
You can run test passing true to show for extension method `GenerateReport` https://github.com/coverlet-coverage/coverlet/blob/master/test/coverlet.core.tests/Coverage/InstrumenterHelper.Assertions.cs#L23

MarcoRossignoli:
@alexthornton1  feel free to take any other `tenet-coverage` issue if you like it! 
Are the most important ones https://github.com/coverlet-coverage/coverlet/issues?q=is%3Aissue+is%3Aopen+label%3Atenet-coverage

src/coverlet.core/Symbols/CecilSymbolHelper.cs
```diff
@@ -396,6 +412,256 @@ private bool SkipGeneratedBranchesForExceptionHandlers(MethodDefinition methodDe
             return _compilerGeneratedBranchesToExclude[methodDefinition.FullName].Contains(instruction.Offset);
         }
 
+        private bool SkipGeneratedBranchesForAwaitForeach(List<Instruction> instructions, Instruction instruction)
```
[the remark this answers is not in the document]
alexthornton1: Yep.  There are a couple of other existing **Skip...** helper methods that aren't static, so I'll go ahead and fix those, too.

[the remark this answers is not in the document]
MarcoRossignoli: Thx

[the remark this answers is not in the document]
alexthornton1: Note: One of them (`SkipGeneratedBranchesForExceptionHandlers`) can't be made `static`, since it references the non-static field `_compilerGeneratedBranchesToExclude`.

src/coverlet.core/Symbols/CecilSymbolHelper.cs
alexthornton1: Yep.  Will do.

src/coverlet.core/Symbols/CecilSymbolHelper.cs
alexthornton1: Indeed.  Will do.

alexthornton1:
`Instrumentation.AsyncForeach.cs`
image
Note that we expect the uncovered portions of the above to be uncovered.  `SumWithATwist` is called with only positive values in its parameter.  `SumEmpty` never enters the loop, since there are no values being iterated.

`Instrumentation.AsyncIterator.cs`
image

MarcoRossignoli:
@alexthornton1 sorry I'm on phone and struggling with correct row for comment, a question, is it possible move there helper as a local function of main one where we elide three new branches?Or are they used also inside async enumerator?I mean if it's possible keep that helper in the context to have more compact code, this class is becoming huge and hard to read.

MarcoRossignoli:
Anyway it's cosmetics, you did great we can go as-is thanks a lot!

alexthornton1:
Great.  I'm taking a crack at `await using` (#914), so I'll see about rearranging that code as part of that work.

PR #1111, opened 2021-03-07
Correct coverage for `await using` (issue #914)

alexthornton1:
This commit corrects coverage for the `await using` statement added
in C# 8, including the variant without an explicit scope (i.e.,
without curly braces and a body), by eliminating two kinds of
patterns that arise in the async state machine that weren't
already being detected and eliminated.


alexthornton1:
There are a couple of interesting things going on in this commit that I thought I should mention:

1. I found myself wanting to ask a question like "Does this instruction load from argument 0?" or "Does this instruction store to the same local variable that the previous instruction loaded from?", but ran into some trouble with Cecil -- which I'm not very familiar with yet -- sometimes deciding the instruction would be something like `ldc.i4.1` and sometimes deciding it would be `ldc.i4 1` instead, inconsistent with what I'd see in disassembly.  So I added a new class in `coverlet.core` called `InstructionExtensions` that smooths out those differences.  If that's working around my misunderstanding of Cecil, I'd be glad to be set straight on that.
2. I found that some of the same patterns I was looking for in `async foreach` occurred again in `async using`, so there's a refactoring opportunity available there, but I wonder if the refactoring opportunity is actually a larger one, because I see similar patterns being searched for elsewhere.  For example, I needed to look for a pattern in `SkipGeneratedBranchesForAwaitUsing.CheckForCleanup` that seemed similar to `SkipGeneratedBranchesForExceptionHandlers`.  But the existing code wouldn't match those patterns -- and that might partly be caused by the issue I found in my own code that seemed to require `InstructionExtensions`.

alexthornton1:
Before:
Before

After:
After

alexthornton1:
Okay, that's my last tweak until it's reviewed.  :)

src/coverlet.core/Extensions/InstructionExtensions.cs
```diff
@@ -0,0 +1,247 @@
+using Mono.Cecil;
+using Mono.Cecil.Cil;
+
+
+namespace Coverlet.Core.Extensions
+{
+    // Extension methods that can smooth out small differences between similar CIL
+    // instructions, such as "ldarg.0" as opposed to "ldarg 0".
+    public static class InstructionExtensions
```
MarcoRossignoli, line 9: If it's not strictly necessary know the index of arg I prefer avoid helper like this for now and keep pattern recognition simple, also because arg index could change in future version of compiler, but a load is a load forever, I hope 😄 

So the question is, does this index id make difference for pattern recognition or can collide with other pattern?
Are there collision sample at the moment?

src/coverlet.core/Symbols/CecilSymbolHelper.cs
```diff
@@ -434,145 +434,321 @@ private static bool SkipGeneratedBranchesForAwaitForeach(List<Instruction> instr
...
-            // IL_012c: ldfld class [System.Private.CoreLib]System.Collections.Generic.IAsyncEnumerator`1<int32> AwaitForeachStateMachine/'<AsyncAwait>d__0'::'<>7__wrap1'
-            // IL_0131: brfalse.s IL_0196
+            // In Debug mode, there are additional things that happen in between
+            // the "call" and branch, but it's the same idea either way: branch
+            // if GetResult() returned true.
 
-            if (instruction.OpCode != OpCodes.Brfalse &&
-                instruction.OpCode != OpCodes.Brfalse_S)
+
+            bool CheckForAsyncEnumerator(List<Instruction> instructions, Instruction instruction, int currentIndex)
```
MarcoRossignoli, line 454: Can you use `static` local function everywhere? https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-8.0/static-local-functions

MarcoRossignoli (changes requested):
Let me know if we can avoid new helpers.
A part from that LGTM

src/coverlet.core/Symbols/CecilSymbolHelper.cs
alexthornton1: Yep.  Will do.

src/coverlet.core/Extensions/InstructionExtensions.cs
alexthornton1: The reason I was checking the index wasn't so much because I needed the index to be something particular like 0, but because I wanted to check that the index used in two different instructions was the same.  So my main goal was to be able to do this:

```
                else if (instructions[currentIndex - 1].IsLdloc(out int localVariableIndex) &&
                         instructions[currentIndex + 1].IsLdloc(out int reloadedLocalVariableIndex) &&
                         localVariableIndex == reloadedLocalVariableIndex &&
                         instructions[currentIndex + 2].OpCode == OpCodes.Callvirt &&
                         instructions[currentIndex + 2].Operand is MethodReference method &&
                         method.DeclaringType.FullName == "System.IAsyncDisposable" &&
                         method.Name == "DisposeAsync")
```

Of course, the pattern would be simpler if I didn't do that, but it also seemed like it was becoming simple enough that it might intersect with some other call to `DisposeAsync` that appeared in user-written code.

Still, I don't think there was an example of where this would lead us astray, so I'll try removing all of that, see if the tests pass, and, if so, will take out the helpers and move on with the simpler approach.

alexthornton1: My other goal was to be able to ask "Is this a local variable load?" without caring what the index was, but the only way I could figure out how to do that reliably was this:

```
                else if ((instructions[currentIndex - 1].OpCode == OpCodes.Ldloc ||
                          instructions[currentIndex - 1].OpCode == OpCodes.Ldloc_S ||
                          instructions[currentIndex - 1].OpCode == OpCodes.Ldloc_0 ||
                          instructions[currentIndex - 1].OpCode == OpCodes.Ldloc_1 ||
                          instructions[currentIndex - 1].OpCode == OpCodes.Ldloc_2 ||
                          instructions[currentIndex - 1].OpCode == OpCodes.Ldloc_3) &&
```

Because unless we know what the index is, then a load from local variable index 1 could theoretically be a `ldloc 1`, a `ldloc.s 1`, or a `ldloc.1`.  But if I don't know which index I care about, then I have to allow for the fact that it might be any of these.

alexthornton1:
After commit 56d124d:
After_56d124d

src/coverlet.core/Extensions/InstructionExtensions.cs
MarcoRossignoli: Got it, let's see....never had a collision like that in past, but you could be right.

MarcoRossignoli (approved):
LGTM

PR #1115, opened 2021-03-08
Threshold not fail when no module

### LINKED ISSUE #1083: Code coverage threshold does not fail if no coverage generated
If coverlet.msbuild is configured with a `Threshold` value for a minimum code coverage required by tests, but then no coverage is output for coverlet for some reason, then rather than fail the build, it instead passes because the total coverage is given as 100%.

```
+--------+------+--------+--------+
| Module | Line | Branch | Method |
+--------+------+--------+--------+

+---------+------+--------+--------+
|         | Line | Branch | Method |
+---------+------+--------+--------+
| Total   | 100% | 100%   | 100%   |
+---------+------+--------+--------+
| Average | NaN% | NaN%   | NaN%   |
+---------+------+--------+--------+
```

You can also see the average is reported as `NaN%`.

See https://github.com/reactiveui/refit/issues/1073 for a scenario where this behaviour masked an issue with code coverage generation because tests still passed.

Coverlet should fail the build and draw the developer's attention to the issue, rather than give a false impression that code coverage is being maintained above previous thresholds configured by the repository.

daveMueller:
closes #1083

### DIFF
--- a/src/coverlet.core/CoverageDetails.cs
+++ b/src/coverlet.core/CoverageDetails.cs
@@ -13,6 +13,6 @@ namespace Coverlet.Core
             internal set { _averageModulePercent = value; }
         }
 
-        public double Percent => Total == 0 ? 100D : Math.Floor((Covered / Total) * 10000) / 100;
+        public double Percent => Total == 0 ? 0 : Math.Floor((Covered / Total) * 10000) / 100;
     }
 }
--- a/src/coverlet.core/CoverageDetails.cs
+++ b/src/coverlet.core/CoverageDetails.cs
@@ -5,20 +5,19 @@ namespace Coverlet.Core
     internal class CoverageDetails
     {
         private double _averageModulePercent;
-        private double _covered;
 
-        public double Covered
+        public Modules Modules { get; internal set; }
+        public double Covered { get; internal set; }
+        public int Total { get; internal set; }
+        public double Percent
         {
-            get => _covered;
-            internal set
+            get
             {
-                _covered = value;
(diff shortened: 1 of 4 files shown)


daveMueller:
An output like the following 

```
+--------+------+--------+--------+
| Module | Line | Branch | Method |
+--------+------+--------+--------+

+---------+------+--------+--------+
|         | Line | Branch | Method |
+---------+------+--------+--------+
| Total   | 100% | 100%   | 100%   |
+---------+------+--------+--------+
| Average | NaN% | NaN%   | NaN%   |
+---------+------+--------+--------+
```

currently only happens when for what ever reason the ```Modules``` are empty. In the first part of the summary there is no entry because no Module exists in the result. That is also the reason why ```Average``` is ```NaN``` because there is a division by zero (Modules.Count=0). 
As the issue already stated a coverage of 100% is the wrong value to return in this case.

There are two solution I can think of. One would throw an exception. Something like "No coverage could be calculated because the result doesn't contain any modules." However this might break some CI pipelines just because coverage has an issue.

The other solution would be to return a coverage of 0 when this happens. This is what I prepared in this PR. But personally I'm a bit divided about this. What do you think, @MarcoRossignoli @martincostello ?

image

martincostello:
Yeah I think 0% is better, as then the coverage threshold will take effect if it's configured and fail the build that way.

MarcoRossignoli:
I agree with Martin, go for 0% and fail if there is a threshold check

MarcoRossignoli (approved):
LGTM

MarcoRossignoli:
@daveMueller remember to update the changelog thx a lot!

PR #1120, opened 2021-03-09
[NOMERGE]Relative paths lcov

phaniva:


MarcoRossignoli:
Thanks @phaniva I'll wait to merge this one because I think it's better let to user decide if use or not relative path in reports.
There are some security concerns on it so shouldn't be the default.
As soon as this guy https://github.com/coverlet-coverage/coverlet/pull/1113 will be merged we'll add new parameter like `--useRelativePathsReport` or something like that to drive report creation.

cc: @daveMueller

phaniva:
@MarcoRossignoli Any idea on how to debug the above failure? The error message appears to indicate some failure reading a csproj file. Is there way to kick off the pipeline one more time?

MarcoRossignoli:
@phaniva can you open an issue for this one?
Can you explain your scenario and why you need relative path?
Can you point me to some docs related to lcov relative path support for reports generators?

phaniva:
@MarcoRossignoli This PR is for addressing the [issue 263](https://github.com/coverlet-coverage/coverlet/issues/263)
This is for addressing issues where coverage reports are generated in one place and the reports are viewed in another place.

Juancanico5 (changes requested):
@MarcoRossignoli

MarcoRossignoli:
There're some conflicts we can give it a try if someone solve it, we need anyway to do a review too cc: @daveMueller @Bertk

Bertk:
This PR was closed because it is outdated.

PR #1159, opened 2021-05-03
Skip assigned auto properties and records

### LINKED ISSUE #1139: Fix SkipAutoProps to work for records as well.
```csharp
    public sealed record RouteResponse
    {
        public string Id { get; init; } = string.Empty;
    }
```

I'm still experiencing that autoprops like these in records are still counting against coverage even though SkipAutoProps is set to true.

I originally was under the impression that #328  would fix it for both classes and records; however, I was told to open this new issue since it doesn't.  Please fix so that we don't have to keep manually exclude them via the [ExcludedFroCodeCoverage] attribute.

daveMueller:
closes #1139


src/coverlet.core/Symbols/CecilSymbolHelper.cs
```diff
@@ -1274,6 +1274,61 @@ private bool SkipExpressionBreakpointsSequences(MethodDefinition methodDefinitio
...
 
+        public bool SkipInlineAssignedAutoProperty(bool skipAutoProps, MethodDefinition methodDefinition, Instruction instruction)
+        {
+            if (!skipAutoProps || !methodDefinition.IsConstructor) return false;
+
+            return SkipGeneratedBackingFieldAssignment(methodDefinition, instruction) ||
+                   SkipDefaultInitializationSystemObject(instruction);
+        }
+
+        private static bool SkipGeneratedBackingFieldAssignment(MethodDefinition methodDefinition, Instruction instruction)
```
MarcoRossignoli, line 1285: do you mind to move all these static to instance method? Now that is a "service" with public contract we can move to instances method.

src/coverlet.core/Symbols/CecilSymbolHelper.cs
```diff
@@ -1274,6 +1274,61 @@ private bool SkipExpressionBreakpointsSequences(MethodDefinition methodDefinitio
...
+                    IL_0000: ldarg.0
+                    IL_0001: ldsfld    string[System.Runtime] System.String::Empty
+                    IL_0006: stfld     string TestRepro.ClassWithPropertyInit::'<SurName>k__BackingField'
+                    ...
+                }
+                ...
+            */
+            var autogeneratedBackingFields = methodDefinition.DeclaringType.Fields.Where(x =>
+                x.CustomAttributes.Any(ca => ca.AttributeType.FullName.Equals(typeof(CompilerGeneratedAttribute).FullName)) &&
+                x.FullName.Contains("k__BackingField"));
```
MarcoRossignoli, line 1308: `EndsWith`?Instead of `Contains`

src/coverlet.core/Symbols/CecilSymbolHelper.cs
```diff
@@ -1274,6 +1274,61 @@ private bool SkipExpressionBreakpointsSequences(MethodDefinition methodDefinitio
...
+            return instruction.OpCode == OpCodes.Ldarg &&
+                   instruction.Next?.Next?.OpCode == OpCodes.Stfld &&
+                   instruction.Next?.Next?.Operand is FieldReference fr &&
+                   autogeneratedBackingFields.Select(x => x.FullName).Contains(fr.FullName);
+        }
+
+        private static bool SkipDefaultInitializationSystemObject(Instruction instruction)
+        {
+        /*
+            A type always has a constructor with a default instantiation of System.Object. For record types these
```
MarcoRossignoli, line 1319: If in future they'll provide record `struct` based we would revisit this. For now I think it's ok.

MarcoRossignoli (approved):
Some nits, LGTM, thanks a lot!

src/coverlet.core/Symbols/CecilSymbolHelper.cs
daveMueller: Sure but those two methods don't access instance data or call other instance methods. Thus I can make them ```private static```. I was told that this can have a performance benefit?

src/coverlet.core/Symbols/CecilSymbolHelper.cs
daveMueller: Done

src/coverlet.core/Symbols/CecilSymbolHelper.cs
daveMueller: Yes then we have to check again if there are strange sequence points in the constructor.

src/coverlet.core/Symbols/CecilSymbolHelper.cs
MarcoRossignoli: Yep I think small improvement(no instance pointer to pass), I don't like too much `static` when I see it every time I need to check if there is some shared state somewhere 😄 , anyway it's not fundamental, we can go with it.

MarcoRossignoli:
REMINDER: changelog @daveMueller

PR #1073, opened 2021-01-28
Path is empty for VB My namespace

daveMueller:
closes #775

PR #1165, opened 2021-05-18
Skip fsharp construct with `unknown` source

### LINKED ISSUE #1145: Coverlet fails to instrument F# projects that have "unknown" source (inline functions, anonymous records etc)
I have a F# solution with a web sdk api project, two shared libraries and some windows services.
Coverlet sucessfully instrument and analyze one of the shared libraries and the windows services but fails to instrument the api project and one of the libraries.

I've seen that there are similar issues here in the repo and while this could be a duplicate of one of those I have two detail that I think sets this apart.
* it works for some projects in my solution and not for others. See below.
* I get the "unknown" file reference in the Instrumentation log entry for the failing projects none of the other issues does that. See below.

I've tested with all three variants of coverlet 3.0.3: msbuild, collector and console and all three give the same result.
Inspecting the logs I see this for the failed instrumentations:

Unsuccessful library and unsuccessful api project,
```
[coverlet] Unable to instrument module: C:\myproject\src\tests\bin\Debug\net5.0\coreLibA.dll, pdb without local source files, [C:\myproject\src\coreLibA\unknown]
[coverlet] Unable to instrument module: C:\myproject\src\tests\bin\Debug\net5.0\myapi.dll, pdb without local source files, [C:\myproject\src\api\unknown]
```

Things I've tried and checked:
* Creating a new solution with same kind of projects - unable to reproduce.
* Does changing the name of the assembly with the <AssemblyName> property have any impact? - No, i have one changed and one unchanged assembly name that works and one changed and one unchanged that doesn't work.

I have two questions:
* What else can I do to find what the significant difference is between the projects that work and those that don't?
* What is the significance of the strange "unknown" file reference at the end that I get `[C:\myproject\src\api\unknown]`. The word "unknown" is not present in the coverlet repo so I assume it comes from something external to coverlet.

daveMueller:
closes #1145


src/coverlet.core/Helpers/InstrumentationHelper.cs
```diff
@@ -121,12 +121,13 @@ public bool EmbeddedPortablePdbHasLocalSource(string module, out string firstNot
...
-
+                                Guid languageGuid = metadataReader.GetGuid(document.Language);
                                 // We verify all docs and return false if not all are present in local
                                 // We could have false negative if doc is not a source
                                 // Btw check for all possible extension could be weak approach
-                                // We exlude from the check the autogenerated source file(i.e. source generators)
-                                if (!_fileSystem.Exists(docName) && !docName.EndsWith(".g.cs"))
+                                // We exclude from the check the autogenerated source file(i.e. source generators)
+                                // We exclude special F# construct https://github.com/coverlet-coverage/coverlet/issues/1145
+                                if (!_fileSystem.Exists(docName) && !docName.EndsWith(".g.cs") && !IsUnknownModuleInFSharpAssembly(languageGuid, docName))
```
petli, line 130: This logic could be extracted into a method, since it is identical with the logic in `PortablePdbHasLocalSource`. This helps if we need to add even more exceptions, e.g. for the null documents in VB (#1073).

daveMueller: Sure, done.

petli (approved):

test/Directory.Build.props
```diff
@@ -1,7 +1,11 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project>
   <Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.props', '$(MSBuildThisFileDirectory)../'))" />
-  <PropertyGroup>
-    <IsTestProject>true</IsTestProject>
-  </PropertyGroup>
+  <Choose>
+    <When Condition="$(MSBuildProjectName)!='coverlet.tests.projectsample.fsharp'">
+      <PropertyGroup>
+        <IsTestProject>true</IsTestProject>
```
MarcoRossignoli, line 7: Can you force false directly on new project `coverlet.tests.projectsample.fsharp.fsproj `?
Or in future we need to add every time projects here.

test/coverlet.core.tests/Helpers/InstrumentationHelperTests.cs
```diff
@@ -3,14 +3,15 @@
...
+using Castle.Core.Internal;
 using Moq;
 using Coverlet.Core.Abstractions;
 
 namespace Coverlet.Core.Helpers.Tests
 {
     public class InstrumentationHelperTests
     {
-        private readonly InstrumentationHelper _instrumentationHelper =
+        private InstrumentationHelper _instrumentationHelper =
```
MarcoRossignoli, line 14: I prefer keep this `readonly` and define new one only for that test wdym?
Like you did in `CanInstrumentFSharpAssemblyWithAnonymousRecord`

MarcoRossignoli (commented):
Some nits but here we go thx a lot!

test/Directory.Build.props
daveMueller: Yes sorry didn't think about this. I was getting a error when ```dotnet test``` the fsharp project and I didn't know why.
image
Took me a good amount of time to figure out it is the ```Directory.Build.pros```. The other c# class library projects don't have this issue even when ```<IsTestProject>true</IsTestProject>```.

test/coverlet.core.tests/Helpers/InstrumentationHelperTests.cs
daveMueller: Done

PR #1169, opened 2021-05-28
Incorrect branch coverage async pattern for targetframework net472

### LINKED ISSUE #1167: Incorrect branch coverage with .NET 472
I am having issues with the branch coverage on a .NET 472 based project. In particular some awaited methods result in two or more branches in the coverage report. I was able to reproduce on a small scale [here](https://github.com/alessandroste/coverlet-test). The project on .NET standard looks good, but the project on .NET 472 results in some suspicious behavior.
The class is exactly the same, but the coverage is different.
image
image
The await on the semaphore ` await this.semaphore.WaitAsync();` results in two branches for .NET 472 and one branch for .NET standard, and the same happens to `await Task.Delay(TimeSpan.FromMilliseconds(1));`. In the second case, both the branches are visited, but, in the first case, only one is visited. This is probably related to the state machine of the task framework.
Is this something that can be solved on coverlet side? I tested with versions back to 3.0.1 and I see the same behavior.

daveMueller:
closes #1167


test/coverlet.tests.projectsample.netframework/coverlet.tests.projectsample.netframework.csproj
```diff
@@ -0,0 +1,12 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net472</TargetFramework>
```
MarcoRossignoli, line 4: Add 
```
    <IsPackable>false</IsPackable>
    <IsTestProject>false</IsTestProject>
```

like https://github.com/coverlet-coverage/coverlet/blob/master/test/coverlet.tests.projectsample.empty/coverlet.tests.projectsample.empty.csproj

MarcoRossignoli (commented):
Some nits, LGTM

test/coverlet.tests.projectsample.netframework/coverlet.tests.projectsample.netframework.csproj
daveMueller: Done

MarcoRossignoli:
@daveMueller small conflict to handle and we can merge a HUGE thanks!

PR #1206, opened 2021-08-03
Return non zero exit code for ThresholdStatistic.Minimum even when no coverage

### LINKED ISSUE #1205: Option --threshold  doesn't work, coverage.json is empty
Hi there!

I am trying to implement Coverlet in my pre-commit script, unit tests are still to be written. I am using this option:
` coverlet CBPm.Validations.Test --target "dotnet" --targetargs "test . --no-build" --threshold 80`

This snippet sholud retrun non zero value if code coverage is less than 80%.
In my case code coverge is 0% and I still get zero exit code instead of non zero exit code.

In addition to this coverage.json file is generated but it is empty.

Any help would be appreciated.

Kind regards! :)

daveMueller:
closes #1205

### DIFF
--- a/src/coverlet.core/CoverageResult.cs
+++ b/src/coverlet.core/CoverageResult.cs
@@ -118,6 +118,9 @@ namespace Coverlet.Core
             {
                 case ThresholdStatistic.Minimum:
                     {
+                        if (!Modules.Any())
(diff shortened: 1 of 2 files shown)


MarcoRossignoli (approved):
LGTM thanks! Remember the changelog.

PR #1230, opened 2021-10-02
Improve logging in case of exception inside static ctor of NetstandardAwareAssemblyResolver

MarcoRossignoli:
close https://github.com/coverlet-coverage/coverlet/issues/1196

image


KirillOsenkov:
Thanks!

PR #1243, opened 2021-10-29
Skip ResolvedFileToPublish list during publish

MarcoRossignoli:
image

During publish skip to copy coverlet files if are inside the `ResolvedFileToPublish` list.

fix https://github.com/coverlet-coverage/coverlet/issues/1045
fix https://github.com/coverlet-coverage/coverlet/issues/1226
fix https://github.com/coverlet-coverage/coverlet/issues/1106
fix https://github.com/coverlet-coverage/coverlet/issues/1131

cc: @daveMueller @Haplois


daveMueller:
Thanks @MarcoRossignoli. I totally see that this fixes the issue. But why doesn't this happen with v1.3. There we already had the target `CopyCoverletDataCollectorFiles`? Also I don't really get why `CopyCoverletDataCollectorFiles` is needed at all? I would assume that nuget copies the resolved files by its own. E.g. in `coverlet.msbuild.targets` we don't have something like this target and the files are copied nevertheless.

Haplois:
I haven't find a chance to look at this yet, I'll do it on this weekend. :)

MarcoRossignoli:
@daveMueller this scenario should run only when we `publish` so with `dotnet publish` and after a user can run test/coverage with `dotnet test file.dll` in msbuild case coverlet is resolved using nuget folder.

PR #1245, opened 2021-11-02
Fix coverage issue for finally block with await

### LINKED ISSUE #1233: Uncovered branch at the of "finally" block when "await" appears within it
I wrote a method in which there was an `await` statement in a `finally` block.  The code was fully covered, but Coverlet reported an uncovered branch at the end of the `finally` block.  I had a few things going on in that method -- `ConfigureAwait` was used in multiple places, there were multiple `await` statements in the `finally` block, including one in an `if` statement -- but I was able to narrow the problem down to something much simpler.

Suppose that we have these methods:
image

Suppose, also, that we have one test that calls into them, which should fully cover them, since there's no (user-written) branching logic in the code under test:
image

If I run that test under Coverlet 3.1.0 (in .NET 5 or .NET 6 RC 1), here's what I see:
image

The missing branch coverage is here (which turns out to be the curly brace closing the `finally` block):
image

This is the IL instruction that's left uncovered:
image

I will attach a repro solution, as well.

Many thanks!

daveMueller:
closes #1233

### DIFF
--- a/Documentation/Changelog.md
+++ b/Documentation/Changelog.md
@@ -7,9 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Fixed 
+-Fix branch coverage issue for finally block with await [#1233](https://github.com/coverlet-coverage/coverlet/issues/1233)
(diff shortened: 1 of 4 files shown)


MarcoRossignoli (approved):
LGTM

PR #1264, opened 2021-12-03
Invariant CLI output

### LINKED ISSUE #1263: Coverage result uses comma instead of dot for decimals
When I run `dotnet coverlet` on an **english Linux** machine, the result uses a dot:

```
| Total   | 35.41% | 24.31% | 51.41% |
```

When I run the same on a **german Windows**, the result uses a comma:

```
| Total   | 35,41% | 24,31% | 51,41% |
```

I tried to set `$Env:DOTNET_CLI_UI_LANGUAGE = "en"`, but it still uses the comma.

Is there any way to configure the output format?

daveMueller:
closes #1263


MarcoRossignoli:
Thx I'll take a look asap

MarcoRossignoli:
@daveMueller some reason to not simply update call sites with `val.ToString(CultureInfo.InvariantCulture)` and avoid new service?

daveMueller:
I also thought about this and you convinced me now :wink:. It's just a habbit of me to write adapters for things like `FormattableString.Invariant(value)` to increase testability. But I also had in mind that in this case we don't really benefit much from it.

MarcoRossignoli (approved):
LGTM

PR #1267, opened 2021-12-14
Negative coverage when integer overflows

### LINKED ISSUE #1266: Negative sequence point coverage when line executions exceed int.MaxValue
_Package_: coverlet.msbuild
_Version_: 3.1.0
_Format_: opencover

In a project where a particular line of code gets called _a lot_, the sequence point coverage value for that line can become negative if the line is hit more than `int.MaxValue` times. The value in the XML is negative and shows in the generated report as uncovered.

I created a simple repro of this because I cannot show the actual code that caused the problem (it's from my job).

XML:
image

Report:
image

Repro code: https://github.com/Malivil/Coverlet1266Repro

daveMueller:
closes #1266


daveMueller:
I'm not so sure about the test. The loop to `int.MaxValue` takes quite some time and slows down the build job.

src/coverlet.core/Coverage.cs
```diff
@@ -420,11 +420,15 @@ private void CalculateCoverage()
                         var hitLocation = result.HitCandidates[i];
                         var document = documentsList[hitLocation.docIndex];
                         int hits = br.ReadInt32();
+                        hits = hits < 0 ? int.MaxValue : hits;
```
petli, line 424: No need to do the work below if there were not hits at the current location (which links up with the next suggestion):

```suggestion
if (hits == 0)
{
    continue;
}
```

src/coverlet.core/Coverage.cs
```diff
@@ -420,11 +420,15 @@ private void CalculateCoverage()
...
 
                         if (hitLocation.isBranch)
                         {
                             var branch = document.Branches[new BranchKey(hitLocation.start, hitLocation.end)];
                             branch.Hits += hits;
+
+                            if (branch.Hits < 0)
```
petli, line 430: There's a tiny risk that we add `int.MaxValue` to 1 resulting in 0, but that can be handled here by checking for any non-positive sum (assuming we've already checked that `hits != 0`, as suggested above):

```suggestion
                            if (branch.Hits <= 0)
```

src/coverlet.core/Coverage.cs
```diff
@@ -437,6 +441,9 @@ private void CalculateCoverage()
 
                                 var line = document.Lines[j];
                                 line.Hits += hits;
+
+                                if (line.Hits < 0)
```
petli, line 445: As above:

```suggestion
                                if (line.Hits <= 0)
```

petli (commented):
I agree the test seems overkill. The interesting logic is in the collection of the hit results, can that be tested by a unit test that just reads negative hits, rather than a full integration test?

petli:
I have deleted two review comments (and updated the third) as I realised I did my sums wrong and the situation I was thinking about couldn't happen.

daveMueller:
Thanks @petli. I will try to somehow mock a `InstrumenterResult` instead but I'm always having a hard time when I want to do this.

src/coverlet.core/Coverage.cs
daveMueller: done

petli (approved):
Good unit test!

MarcoRossignoli:
Thanks to all!

PR #1290, opened 2022-02-05
Put AppContext.OnProcessExit body inside a try/finally

MarcoRossignoli:
fixes https://github.com/coverlet-coverage/coverlet/issues/1287

image

cc: @ViktorHofer @stephentoub @jkotas

### DIFF
--- a/src/coverlet.core/Instrumentation/Instrumenter.cs
+++ b/src/coverlet.core/Instrumentation/Instrumenter.cs
@@ -297,10 +297,27 @@ namespace Coverlet.Core.Instrumentation
                         var onProcessExitMethod = new MethodReference("OnProcessExit", module.TypeSystem.Void, appContextType).Resolve();
                         var onProcessExitIl = onProcessExitMethod.Body.GetILProcessor();
 
-                        lastInstr = onProcessExitIl.Body.Instructions.Last();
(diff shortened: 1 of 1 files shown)


stephentoub:
Thank you for fixing it.

stephentoub:
When do we expect a new official nuget release with the corelib-related fixes to published?

MarcoRossignoli:
I'd like a test before release a new version, to be sure that is working as expected...is it possible that someone with more "knowledge" and a ready runtime env give it a try with the dogfood?
I'm a bit rusty with the new layout/build of runtime repo.

stephentoub:
> I'd like a test before release a new version, to be sure that is working as expected...is it possible that someone with more "knowledge" and a ready runtime env give it a try with the dogfood?

Sure, I can try on Monday.

stephentoub:
Is the code you're generating actually what's shown in the image above?  return'ing out of a finally is invalid.

MarcoRossignoli:
I've fixed...pair of issues, missing end finally and patch every return. Looks like ILSpy validation is not complete, sometimes it gives errors...not this time. Too much time far from IL generation 🙂

Also we don't have specific tests for the corelib custom case, it doesn't help.

PR #1301, opened 2022-02-14
Enable nullable for new files while preserving non-nullable for existing

Evangelink:


Bertk:
This PR was closed because it is outdated.

PR #1304, opened 2022-02-17
Improves unexpected packages folder state error

Evangelink:


test/coverlet.integration.tests/BaseTest.cs
```diff
@@ -44,11 +44,24 @@ private protected string GetPackageVersion(string filter)
...
-            return manifest.Metadata.Version.OriginalVersion;
+            string packagesPath = $"../../../../../bin/{GetAssemblyBuildConfiguration()}/Packages";
+            var files = Directory.GetFiles(packagesPath, filter).ToList();
+            if (files.Count == 0)
+            {
+                throw new InvalidOperationException($"Could not find any package using filter '{filter}' in folder '{Path.GetFullPath(packagesPath)}'. Make sure 'dotnet pack' was called.");
+            }
+            else if (files.Count > 1)
+            {
+                throw new InvalidOperationException($"Found more than one package using filter '{filter}' in folder '{Path.GetFullPath(packagesPath)}'. Make sure 'dotnet pack' was only called once.");
```
Evangelink, line 55: Are there other conditions that would lead to the same state? Also, in such case, wouldn't it be better to sort by name and take last (i.e. biggest version or suffix).

MarcoRossignoli: I prefer fail and have the same "CI" experience, I don't want to waste time because the state between local and CI is "sightly" different and I need to investigate what's happening.

MarcoRossignoli (approved):

PR #1305, opened 2022-02-17
Increase severity of some rules and fix instances

Evangelink:
@MarcoRossignoli Please let me know if there are parts of the "default" state you would like to tweak.


src/coverlet.collector/Utilities/TestPlatformEqtTrace.cs
```diff
@@ -8,17 +8,17 @@ namespace Coverlet.Collector.Utilities
     /// <summary>
     /// Test platform eqttrace
     /// </summary>
-    internal class TestPlatformEqtTrace
+    internal static class TestPlatformEqtTrace
```
Evangelink, line 11: Updated the type to be `static` as all members could be `static`.  If you prefer an instance with the idea of one day have a different instance for tests/prod then let me know I can revert. We could also wait for such need and refactor and that time.

src/coverlet.core/CoverageSummary.cs
```diff
@@ -7,17 +7,17 @@
 
 namespace Coverlet.Core
 {
-    internal class CoverageSummary
+    internal static class CoverageSummary
```
Evangelink, line 10: Same remark as for the eqt logger. All members could be put `static` so I have updated everything.

src/coverlet.console/Logging/ConsoleLogger.cs
```diff
@@ -9,7 +9,8 @@ namespace Coverlet.Console.Logging
 {
     class ConsoleLogger : ILogger
     {
-        private static readonly object _sync = new object();
+        private static readonly object s_sync = new();
```
Evangelink, line 12: `s_` is the default that was approved but if you want to change it to make no diff or use a different pattern, let me know.

src/coverlet.core/Instrumentation/ReachabilityHelper.cs
```diff
@@ -169,7 +169,9 @@ public override string ToString()
         /// OpCodes that transfer control code, even if they do not
         /// introduce branch points.
         /// </summary>
+#pragma warning disable IDE1006 // Naming Styles
```
Evangelink, line 172: I thought you would prefer not to change the name of the field but if you are ok with changing, please let me know.

src/coverlet.core/Instrumentation/ReachabilityHelper.cs
```diff
@@ -224,7 +226,9 @@ public override string ToString()
         /// OpCodes that unconditionally transfer control, so there
         /// is not "fall through" branch target.
         /// </summary>
+#pragma warning disable IDE1006 // Naming Styles
```
Evangelink, line 229: I thought you would prefer not to change the name of the field but if you are ok with changing, please let me know.

src/coverlet.collector/Utilities/TestPlatformEqtTrace.cs
MarcoRossignoli: Yep pls revert I want to able to test/inject everything

src/coverlet.core/CoverageSummary.cs
MarcoRossignoli: the same

src/coverlet.core/Instrumentation/ReachabilityHelper.cs
MarcoRossignoli: I'm ok with field name update

src/coverlet.core/Instrumentation/ReachabilityHelper.cs
MarcoRossignoli: I'm ok with it

src/coverlet.console/Logging/ConsoleLogger.cs
MarcoRossignoli: I'm ok with field name update

MarcoRossignoli:
@Evangelink pls keep the instances(no static) and I'm ok with the `s_` renaming

MarcoRossignoli (approved):
thx!

PR #1306, opened 2022-02-27
ExcludeFromCodeCoverage attribute on local functions ignores lambda expression

### LINKED ISSUE #1302: `ExcludeFromCodeCoverage` attribute is ignored on local function
Having the following code:

```csharp
public class ExcludedLocalFunctionReproduction
{
    public void SomethingThatIsUsingALocalFunction()
    {
        MethodThatTakesALambda(NewMethod());

        [ExcludeFromCodeCoverage(Justification = "Not testable")]
        static Func<string, bool> NewMethod()
        {
            return myString => myString.Length == 10;
        }
    }

    private void MethodThatTakesALambda(Func<string, bool> function)
    {
        // Do something - the actual function is not invoked to show the issue
        bool _ = function != null;
    }
}
```

Coverlet reports the line `return myString => myString.Length == 10;` as uncovered even though the local method is excluded from code coverage:

image

I would expect the local function to be shown as excluded.

The full code (along with other reproductions for already reported issues) is here: https://github.com/meggima/coverlet-reproductions

daveMueller:
closes #1302

### DIFF
--- a/Documentation/Changelog.md
+++ b/Documentation/Changelog.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+-ExcludeFromCodeCoverage attribute on local functions ignores lambda expression [#1302](https://github.com/coverlet-coverage/coverlet/issues/1302)
+
 ## Release date 2022-02-06
 ### Packages  
 coverlet.msbuild 3.1.2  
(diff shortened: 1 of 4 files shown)


daveMueller:
The issue is only for lambda methods inside local functions. The algorithm that checks if the function name is synthesized by the compiler doesn't work here because the base for the lambda method name is the 'parent' function and not the local function. 

For example when we have the following code:

```csharp
public void ParentFunction()
{
    [System.Diagnostics.CodeAnalysis.ExcludeFromCodeCoverage]
    static Func<string, bool> LocalFunction()
    {
        return myString => myString.Length == 10;
     }
}
```
The name of the local function is sth like `<ParentFunction>g__LocalFunction|0_0` and the name of the lambda method sth like `<ParentFunction>b__0_1`. The current algorithem compares only the name of the 'parent' function which is `ParentFunction` and looks up if it is already excluded.
Unfortunately `ParentFunction` isn't the one with the `ExcludeFromCodeCoverage` attribute and thus the lambda method isn't excluded.

I couldn't think of any other solution besides this PR because from a name like `<ParentFunction>b__0_1` it seems impossible to figure out if the lambda method is inside a local function or not.

PR #1312, opened 2022-03-15
Wrong coverage for await foreach with generic method

### LINKED ISSUE #1210: Wrong coverage for "await foreach" when method is generic
I'm using coverlet.msbuild v3.1.0 (and tried also with the current nightly build) and there is an uncovered statement reported.
image

I tried to reproduce it but I didn't managed to get an reproduction.
Any ideas what the problem could be and how I can provide an reproduction?

```csharp
public class AsyncForeachReproduction
{
    private static async IAsyncEnumerable<int> RangeAsync(int start, int count)
    {
        for (int i = 0; i < count; i++)
        {
            await Task.Delay(i);
            yield return start + i;
        }
    }

    public async Task Execute()
    {
        await foreach (var a in RangeAsync(1, 5))
        {
            Console.WriteLine(a);
            await Task.CompletedTask;
        }

        await Task.CompletedTask;
    }
}
```

daveMueller:
closes #1210 

grafik

### DIFF
--- a/Documentation/Changelog.md
+++ b/Documentation/Changelog.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+-Await foreach has wrong branch coverage when method is generic [#1210](https://github.com/coverlet-coverage/coverlet/issues/1210)
+
 ## Release date 2022-02-06
 ### Packages  
 coverlet.msbuild 3.1.2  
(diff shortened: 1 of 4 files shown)


daveMueller:
@MarcoRossignoli Can you run the build again? The failed test has nothing to do with my changes. The job already was green before I updated the changelog.

MarcoRossignoli (approved):
Thanks a lot ❤️

daveMueller:
What I've analysed is that in the case of generics, the operand isn't a `FieldDefinition`. So every symbol check that is looking for a `FieldDefinition` will not match. There are places in `CecilSymbolHelper` where this is already considered, e.g. [https://github.com/coverlet-coverage/coverlet/blob/master/src/coverlet.core/Symbols/CecilSymbolHelper.cs#L114-L115](https://github.com/coverlet-coverage/coverlet/blob/master/src/coverlet.core/Symbols/CecilSymbolHelper.cs#L114-L115).But there are also a couple of places where it isn't. 
In this PR I only changed the necessary places to fix the bug but maybe it would make sense to change this everywhere in `CecilSymbolHelper`?

MarcoRossignoli:
I think it's ok as-is for now, if we don't have feedback.
If you feel safe to update also around go for it.

PR #1320, opened 2022-04-08
Add project name to the CoverletSourceRootsMapping file to avoid file locking

MarcoRossignoli:
fixes https://github.com/coverlet-coverage/coverlet/issues/1155


zplan:
Can someone of  the contributers or authors looks why this PR can not get in (@tonerdo ?). 
I think a lot of people are waiting for this PR, including me : )

MarcoRossignoli:
I'm working on it, but I'm bit busy atm.

MarcoRossignoli:
You can dogfood the fix here for now https://github.com/coverlet-coverage/coverlet/issues/1155#issuecomment-1073289885

something weird with CI...

daveMueller:
Hi @MarcoRossignoli ,

so the issue is that after changing the `sourceRootMappingFilePath`, the `InstrumenterHelper` throws an exception when trying 
to load the type of the tracker.

```csharp
tracker = asm.GetTypes().Single(n => n.FullName.Contains("Coverlet.Core.Instrumentation.Tracker"));
```

The reason is that the tracker isn't injected (or not correct) into the assembly (coverlet.core.tests). It can't be found
with the code snipped above and thus `Single(n => ...)` throws an exception.

It's super weird that if changing the `sourceRootMappingFilePath` back to the way it was before, the tracker is injected in the 
assembly again. I also can't reproduce this locally and it only happens on the build server.

Any idea what could cause the tracker to not beeing injected into the assembly?

MarcoRossignoli:
But if we disable the coverage for our tests(we don't dogfood) does it work?

daveMueller:
Not I tried this and it didn't help. Basically every unit test that is using `TestInstrumentationHelper.Run<AwaitUsing>` fails. If I remove all those tests (most of them are in `coverlet.core.tests\Coverage`) I can get build and the remaining test green.

daveMueller:
The issue is that `Coverlet.Core.Instrumentation.Tracker` isn't injected into the assembly anymore and up to now I couldn't figure out why.

MarcoRossignoli:
closed here https://github.com/coverlet-coverage/coverlet/pull/1456

PR #1323, opened 2022-04-18
Sourcelink for code generators

### LINKED ISSUE #1322: Incorrect coverage - only one module detected
When using the MSBuild collector I get an incorrect coverage result. Coverlet only detects one module while the test project references multiple projects that are being tested.

[Here](https://github.com/klemmchr/MvvmBlazor/runs/5995940063?check_suite_focus=true#step:5:35) you can find a sample run that shows this issue. The project references can be found [here](https://github.com/klemmchr/MvvmBlazor/blob/workflow-adjust/src/MvvmBlazor.Tests/MvvmBlazor.Tests.csproj#L28).

daveMueller:
closes #1322

### DIFF
--- a/src/coverlet.core/Coverage.cs
+++ b/src/coverlet.core/Coverage.cs
@@ -513,8 +513,12 @@ namespace Coverlet.Core
             string replacement = Path.Combine(relativePathOfBestMatch, Path.GetFileName(document));
             replacement = replacement.Replace('\\', '/');
 
-            url = sourceLinkDocuments[keyWithBestMatch];
-            return url.Replace("*", replacement);
+            if (sourceLinkDocuments.TryGetValue(keyWithBestMatch, out url))
+            {
+                return url.Replace("*", replacement);
+            }
+
+            return document;
         }
     }
 }


daveMueller:
While analyzing #1322 I stumbled across the issue that getting the sourcelink url for a code generator class throws. As a quick fix I just returned the relative path so that it doesn't fail anymore. But I could also think of something like removing those files from the coverage result?

I think there is still a general problem with code generators. Sure we can collect hits et cetera, but we can't really map this to a source file because initially no physical file exists. And I think there are a couple of  places in coverlet where this causes problems. E.g. the `SourceRootTranslator` can't resolve the file path because it doesn't exist.
Maybe it would be the best to exclude those files from the instrumentation? Was there already a discussion about this that I missed? 
cc: @MarcoRossignoli @petli @tonerdo

MarcoRossignoli:
I think that at the moment we have this one https://github.com/coverlet-coverage/coverlet/issues/1164 on the subject

MarcoRossignoli (approved):

Issue #1164, opened 2021-05-17
New parameter ExcludeAssembliesWithoutSources to control automatic assembly exclusion


petli:
## Problem

Coverlet automatically excludes third-party assemblies from the coverage analysis, since they are usually not interesting to the user. The heuristic for determining if an assembly is a third-party dependency is if any of its source documents does not have a corresponding source file.

(Ref: https://github.com/coverlet-coverage/coverlet/blob/3ae19761876f2ac809c227ca3563495e28993b59/src/coverlet.core/Helpers/InstrumentationHelper.cs#L125-L129)

C# source generators and some F# constructs add source documents to firsrt-party assemblies, causing them to be excluded by this heuristic, and compiler bugs may also trigger this. See e.g. #1084 #1145 #1100 #1297

The C# source generators are handled by identifying them using a best practice that the document names should end in ".g.cs", and similar logic can be added for the other issues. But the source generator naming is but a best practice and not all projects will follow it, and new compiler issues will likely turn up in the future, so this heuristic will continue to cause issues.

## Suggested solution

Add a new parameter `ExcludeAssembliesWithoutSources` with the values `MissingAny`, `MissingAll` and `None`.  `MissingAny` is the current behaviour, but `MissingAll` would be the new default behaviour to avoid these and future issues with generators. `None` would be a new setting to never filter on source files, for those rare cases where it may not make sense.  Explicit include/exclude settings would override any filtering from this parameter.  

Changing the default to include an assembly in the coverage if at least one source document exist may cause some third-party documents to be included, if the same file name exists in the project being analysed with coverlet (pointed out by @daveMueller). Users affected by this can then either change the setting to `MissingAny` to revert to the old behaviour, or explicitly exclude those assemblies.

Coverlet should still have a heuristic to ignore generated source files according to conventions, but now to avoid logging those cases.

(Edit: change parameter name in feature description to `ExcludeAssembliesWithoutSources`)

petli:
A more consistent parameter name may be `ExcludeAssembliesWithoutSources`.

TomasMorton:
Ran into this because we heavily use this library: https://github.com/mcintyre321/OneOf with the source generator. It generates `GenerateOneOfAttribute.cs` in all of our projects. This causes coverlet to fail:
[...]

Issue #1084, opened 2021-02-10
Code coverage stops working if assembly contains interfaces from source generators


martincostello:
_Originally posted as https://github.com/reactiveui/refit/issues/1073_

**Describe the bug**

In a project using Refit 6.0.8 and coverlet.msbuild 3.0.2, upon updating from Refit 5.2.4, code coverage is no longer being collected for the assembly containing Refit interfaces.

Using coverlet's troubleshooting guide [here](https://github.com/coverlet-coverage/coverlet/blob/master/Documentation/Troubleshooting.md#msbuild-integration), I narrowed my investigation down to this line of interest:

```
[coverlet] Unable to instrument module: C:\Coding\Projects\MyProject\tests\MyProject.Tests\bin\debug\net5.0\MyProject.dll, pdb without local source files, [C:\Coding\Projects\MyProject\src\MyProject\InterfaceStubGenerator.Core\Refit.Generator.InterfaceStubGenerator\IMyApi.g.cs]
```

As there's no file on disk here, that sort of makes sense, but causes an issue as it breaks all code coverage for the project for the sake of the one interface.

I'm guessing this is something to do with the usage of source generators in Refit.

The failure then slipped through several CI jobs for Refit updates until it was noticed due to https://github.com/coverlet-coverage/coverlet/issues/1083.

**Steps To Reproduce**

Clone https://github.com/martincostello/alexa-london-travel and run `./build.ps1`.

The build script will eventually output something like this:

```
+--------+------+--------+--------+
| Module | Line | Branch | Method |
+--------+------+--------+--------+

+---------+------+--------+--------+
|         | Line | Branch | Method |
+---------+------+--------+--------+
| Total   | 100% | 100%   | 100%   |
+---------+------+--------+--------+
| Average | NaN% | NaN%   | NaN%   |
+---------+------+--------+--------+
```

Example build: https://github.com/martincostello/alexa-london-travel/runs/1869767942?check_suite_focus=true#step:4:78

This seems to be affecting multiple repos I have using Refit, not just this one.

Here's another example: https://github.com/martincostello/alexa-london-travel-site/runs/1870035564?check_suite_focus=true#step:4:160

**Expected behavior**

Assemblies containing Refit interfaces are instrumented correctly.

For the referenced repository, it should output something like this from `build.ps1`:

```
+--------------------+-------+--------+--------+
| Module             | Line  | Branch | Method |
+--------------------+-------+--------+--------+
[...]

PR #1352, opened 2022-06-16
Add some more Roslyn generated file suffixes

georgevanburgh:
As discussed in #1084, coverlet fails to generate metrics for assemblies that contain source generators. This was improved in #1091 by skipping files that end in `.g.cs`. However, some source generators produce files which do not match this pattern - for example [`protobuf-net.BuildTools`](https://github.com/protobuf-net/protobuf-net/blob/683d112113fd251ce9eae0a806ac931b79166b20/src/protobuf-net.BuildTools/Generators/ProtoFileGenerator.cs#L233). Obviously coverlet is never going to be able to match against every possible generated file naming convention - which is where the suggestion in #1164 comes from. I understand if you'd rather use the ideas from that issue in preference to this change, however, I thought it might make sense to add the list of [conventions used internally by `Microsoft.CodeAnalysis`](https://sourceroslyn.io/#Microsoft.CodeAnalysis/InternalUtilities/GeneratedCodeUtilities.cs,104) to the logic introduced in #1091 for detecting auto-generated files (selfishly, this would also fix the issues I'm having with projects that reference `protobuf-net.BuildTools` 😇).


petli (approved):
Thank you! I think it should be pretty safe to add these additional conditions, it is quite unlikely that it will result in a module being instrumented that wasn't before.

PR #1360, opened 2022-07-10
Added InstrumentModulesWithoutLocalSources setting 

TFTomSun:
There are already several issues regarding unexpected empty coverage results. 
https://github.com/coverlet-coverage/coverlet/issues/1089
https://github.com/coverlet-coverage/coverlet/issues/1121

We found out, that there's an implicit filter that checks whether a pdb source location can be mapped to a local file. However that logic is not optimal for some use cases, for example when the build and test run happens on a different machines or when the sources are embedded in the assembly. Since the source files are not used anywhere in the whole instrumentation logic except for that filter logic, we added an option to disable that implicit filter. The previous behavior stays the default behavior. With that setting activated, the user has full control over the filtering and can still filter using Include / Exclude to filter out 3rdParty or include only the assemblies he explicitly wants to measure.

### DIFF
--- a/src/coverlet.console/Program.cs
+++ b/src/coverlet.console/Program.cs
@@ -70,6 +70,7 @@ namespace Coverlet.Console
             CommandOption mergeWith = app.Option("--merge-with", "Path to existing coverage result to merge.", CommandOptionType.SingleValue);
             CommandOption useSourceLink = app.Option("--use-source-link", "Specifies whether to use SourceLink URIs in place of file system paths.", CommandOptionType.NoValue);
             CommandOption doesNotReturnAttributes = app.Option("--does-not-return-attribute", "Attributes that mark methods that do not return.", CommandOptionType.MultipleValue);
+            CommandOption instrumentModulesWithoutLocalSources = app.Option("--instrument-modules-without-local-sources", "Specifies whether modules should be instrumented even if the sources from the PDBs can't be found locally.", CommandOptionType.MultipleValue);
 
             app.OnExecute(() =>
             {
@@ -97,7 +98,8 @@ namespace Coverlet.Console
                     MergeWith = mergeWith.Value(),
                     UseSourceLink = useSourceLink.HasValue(),
                     SkipAutoProps = skipAutoProp.HasValue(),
-                    DoesNotReturnAttributes = doesNotReturnAttributes.Values.ToArray()
+                    DoesNotReturnAttributes = doesNotReturnAttributes.Values.ToArray(),
+                    InstrumentModulesWithoutLocalSources = instrumentModulesWithoutLocalSources.HasValue(),
                 };
 
                 ISourceRootTranslator sourceRootTranslator = serviceProvider.GetRequiredService<ISourceRootTranslator>();
--- a/src/coverlet.console/Program.cs
+++ b/src/coverlet.console/Program.cs
@@ -70,7 +70,7 @@ namespace Coverlet.Console
             CommandOption mergeWith = app.Option("--merge-with", "Path to existing coverage result to merge.", CommandOptionType.SingleValue);
             CommandOption useSourceLink = app.Option("--use-source-link", "Specifies whether to use SourceLink URIs in place of file system paths.", CommandOptionType.NoValue);
             CommandOption doesNotReturnAttributes = app.Option("--does-not-return-attribute", "Attributes that mark methods that do not return.", CommandOptionType.MultipleValue);
-            CommandOption instrumentModulesWithoutLocalSources = app.Option("--instrument-modules-without-local-sources", "Specifies whether modules should be instrumented even if the sources from the PDBs can't be found locally.", CommandOptionType.MultipleValue);
+            CommandOption instrumentModulesWithoutLocalSources = app.Option("--instrument-modules-without-local-sources", "Specifies whether modules should be instrumented even if the sources from the PDBs can't be found locally.", CommandOptionType.NoValue);
 
             app.OnExecute(() =>
             {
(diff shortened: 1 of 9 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/coverlet-coverage/coverlet?pullRequest=1360) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/coverlet-coverage/coverlet?pullRequest=1360) before we can accept your contribution.<br/><br/>:x: TFTomSun [sign now](https://cla.dotnetfoundation.org/coverlet-coverage/coverlet?pullRequest=1360)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/coverlet-coverage/coverlet?pullRequest=1360) it.</sub>

searoz:
Hi. I've just created a similar PR: https://github.com/coverlet-coverage/coverlet/pull/1363
The major difference is that I allow more precise control over which modules are going to be force-instrumented

TFTomSun:
@searoz there are already include / exclude parameters available. I'd prefer the option allow all and use the existing filter possibilities instead. Are the existing filter options not sufficient for your case?

petli:
@TFTomSun @searoz Thanks for both your PRs adressing this longstanding issue! I also think the approach of having a flag that controls the heuristic and then rely on normal include/exclude filtering is less complex both to understand as a user and to maintain in code.

Have you seen #1164 which also discusses the topic? Does it make more sense to you to have a flag that overrides the automatic exclusion of supposed third-party assemblies, as in both your PRs, or have a flag that instead allows finetuning the behaviour of that behaviour as in that issue?

TFTomSun:
@petli  I found the default filter based on local sources confusing, We have for example build packages that link additional files into the project, which are not available if only the sources are cloned. Additionally we usually don't do test runs in dev environments. We want to avoid issues in production that are caused by dev environment expectations. Therefore turning off that local source based filter completely is what we need. We need to distinguish internal and external assemblies based on their names.

petli:
@TFTomSun I think the current behaviour works well for straightforward projects out-of-the-box, but as soon as you start having a more complex build process there's a need for being able to turn it off.  

Since there is also the issue that this behaviour easily gets thrown off by generated sources, I was thinking it could be handled with a single parameter with multiple options, but I'm coming around to that it would be clearer for the user to have the option in this PR that completely disables the automatic filtering, and addressing the generated source files by relaxing the behaviour to include assemblies as long as there is at least one local source file. Any opinion, @MarcoRossignoli?

src/coverlet.console/Program.cs
```diff
@@ -70,6 +70,7 @@ static int Main(string[] args)
             CommandOption mergeWith = app.Option("--merge-with", "Path to existing coverage result to merge.", CommandOptionType.SingleValue);
             CommandOption useSourceLink = app.Option("--use-source-link", "Specifies whether to use SourceLink URIs in place of file system paths.", CommandOptionType.NoValue);
             CommandOption doesNotReturnAttributes = app.Option("--does-not-return-attribute", "Attributes that mark methods that do not return.", CommandOptionType.MultipleValue);
+            CommandOption instrumentModulesWithoutLocalSources = app.Option("--instrument-modules-without-local-sources", "Specifies whether modules should be instrumented even if the sources from the PDBs can't be found locally.", CommandOptionType.MultipleValue);
```
petli, line 73: Should this be a simple boolean flag like `--use-source-link` abouve?

```suggestion
            CommandOption instrumentModulesWithoutLocalSources = app.Option("--instrument-modules-without-local-sources", "Specifies whether modules should be instrumented even if the sources from the PDBs can't be found locally.", CommandOptionType.NoValue);
```

TFTomSun: yes I committed the suggested change

MarcoRossignoli:
@petli I'm with the idea to have the option to exclude heuristic and move the "any" to "all".

The heuristic was in place to automatically exclude for instance xunit, test adapters etc...that are close to the user dll test.

@TFTomSun I'm oof for a pair of week I'll take a look when I'll be back from the vacation. @petli feel free to review and approve and I'll merge.

petli (changes requested):
I think the code looks good, but can you take a look at the comment https://github.com/coverlet-coverage/coverlet/pull/1360#issuecomment-1185285557 about the cobertura generator and see if that needs addressing with this solution too.

It would also be golden if you can add this new flag to the documentation for the different drivers too. :)

TFTomSun:
@petli What's the expected behavior in the reporter? Do we expect the sources files to be available there? Our plan was to execute the tests in a close-to-production environment and create the report in a dev environment. As far as I know, the reporter needs the source files, right? So maybe that debug.assert still makes sense?

petli:
@TFTomSun I'm not sure myself, I haven't worked with or used the cobertura reporter. 

@searoz Do you have any input on this?

searoz:
@TFTomSun @petli AFAIU, cobertura reporter only uses sources to "fix" file paths in the report. IMHO, this is a pretty minor feature that could be safely ignored. The only reason I brought this up is the fact that cobertura reporter has `Debug.Assert` right here: https://github.com/coverlet-coverage/coverlet/blob/2bb04f21ec30978335ad00b5d52b473c45181070/src/coverlet.core/Reporters/CoberturaReporter.cs#L231

During my tests this `Debug.Assert` caused `dotnet test` to fail, so it should be either ignored when `InstrumentModulesWithoutLocalSources` option is specified or just removed entirely

petli:
Sounds like the `Assert` can be removed, given that circumstances have changed here.

TFTomSun:
> Sounds like the `Assert` can be removed, given that circumstances have changed here.

done

petli (approved):
The code looks good to me. It would be useful with adding the new parameters to the documentation for the three drivers, but I can also help out with that later in a separate PR.

searoz:
@petli Do you know when can we expect new Coverlet version with these changes included? Our CI/CD depends on them (that's why I created my own PR in the first place)

petli:
@searoz Once this is merged it would be included in the nightly builds, so you could target those until a release is published by @MarcoRossignoli.

MarcoRossignoli:
@searoz I've merged, you can dogfood directly from your CI since tomorrow

PR #1392, opened 2022-10-16
New parameter to control automatic assembly exclusion

### LINKED ISSUE #1164: New parameter ExcludeAssembliesWithoutSources to control automatic assembly exclusion

daveMueller:
closes #1164 

There are two things I wasn't really sure about.

I aligned the parameter description to the description in the feature #1164. But in the rest of the documentation the 
values for the other parameters are always lowercased. So I would suggest to also describe them in the documentation 
as `missingall,missingany,none`.
In the actual code it doesn't  matter because it is compared case insensitive.

The second thing is that in the feature description it says that `MissingAny`should be the current behavior. Personally I 
think that after changing the default behavior of the heuristic to `MissingAll`, the exclusion of generated document names 
and the exclusion of the `unknown` F-sharp module don't make much sense anymore. Thus I removed the functions `IsGeneratedDocumentName` and `IsUnknownModuleInFSharpAssembly` from the `InstrumentationHelper`.
But I can also add them back again?

### DIFF
--- a/Documentation/Changelog.md
+++ b/Documentation/Changelog.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 -ExcludeFromCodeCoverage attribute on local functions ignores lambda expression [#1302](https://github.com/coverlet-coverage/coverlet/issues/1302)
 
 ### Improvements  
--Added new parameter to control automatic assembly exclusion and changed the default behavior [#1164](https://github.com/coverlet-coverage/coverlet/issues/1164)
+-Added new parameter to control automatic assembly exclusion and changed the default behavior [#1164](https://github.com/coverlet-coverage/coverlet/issues/1164) 
 
 ## Release date 2022-02-06
 ### Packages  
--- a/Documentation/Changelog.md
+++ b/Documentation/Changelog.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Breaking changes
--New parameter ExcludeAssembliesWithoutSources to control automatic assembly exclusion [1164](https://github.com/coverlet-coverage/coverlet/issues/1164)
+-New parameter ExcludeAssembliesWithoutSources to control automatic assembly exclusion [1164](https://github.com/coverlet-coverage/coverlet/issues/1164) 
 
 ## Release date 2022-10-29
 ### Packages  
--- a/Documentation/Changelog.md
+++ b/Documentation/Changelog.md
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 -Await foreach has wrong branch coverage when method is generic [#1210](https://github.com/coverlet-coverage/coverlet/issues/1210)
 -ExcludeFromCodeCoverage attribute on local functions ignores lambda expression [#1302](https://github.com/coverlet-coverage/coverlet/issues/1302)
 
+### Improvements  
+-Added new parameter to control automatic assembly exclusion and changed the default behavior [#1164](https://github.com/coverlet-coverage/coverlet/issues/1164)
+
 ## Release date 2022-02-06
 ### Packages  
 coverlet.msbuild 3.1.2  
--- a/Documentation/Changelog.md
+++ b/Documentation/Changelog.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Breaking changes
+-New parameter ExcludeAssembliesWithoutSources to control automatic assembly exclusion [1164](https://github.com/coverlet-coverage/coverlet/issues/1164)
+
 ## Release date 2022-10-29
 ### Packages  
 coverlet.msbuild 3.2.0  
--- a/Documentation/Changelog.md
+++ b/Documentation/Changelog.md
@@ -7,7 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Breaking changes
--New parameter ExcludeAssembliesWithoutSources to control automatic assembly exclusion [1164](https://github.com/coverlet-coverage/coverlet/issues/1164) 
+- New parameter `ExcludeAssembliesWithoutSources` to control automatic assembly exclusion [1164](https://github.com/coverlet-coverage/coverlet/issues/1164). The parameter `InstrumentModulesWithoutLocalSources` has been removed. since it can be handled by setting `ExcludeAssembliesWithoutSources` to `None`.
+- The default heuristics for determining whether to instrument an assembly has been changed. In previous versions any missing source file was taken as a signal that it was a third-party project that shouldn't be instrumented, with exceptions for some common file name patterns for source generators. Now only assemblies where no source files at all can be found are excluded from instrumentation, and the code for detecting source generator files have been removed. To get back to the behaviour that at least one missing file is sufficient to exclude an assembly, set `ExcludeAssembliesWithoutSources` to `MissingAny`, or use assembly exclusion filters for more fine-grained control.
 
 ## Release date 2022-10-29
 ### Packages  
--- a/Documentation/Changelog.md
+++ b/Documentation/Changelog.md
@@ -24,9 +24,6 @@ coverlet.collector 3.2.0
 ### Added  
 -Added InstrumentModulesWithoutLocalSources setting [#1360](https://github.com/coverlet-coverage/coverlet/pull/1360) by @TFTomSun  
 
-### Improvements  
--Added new parameter to control automatic assembly exclusion and changed the default behavior [#1164](https://github.com/coverlet-coverage/coverlet/issues/1164) 
-
 ## Release date 2022-02-06
 ### Packages  
 coverlet.msbuild 3.1.2  
--- a/Documentation/MSBuildIntegration.md
+++ b/Documentation/MSBuildIntegration.md
@@ -236,8 +236,8 @@ This parameter has three different values to control the automatic assembly excl
 
 | Parameter | Description |
 |-----------|-------------|
-| MissingAny | Includes the assembly if at least one document is matching. In case the `ExcludeAssembliesWithoutSources` parameter is not specified the default value is `MissingAny`. |
-| MissingAll | Includes the assembly only if all documents can be matched to corresponding source files. |
+| MissingAll | Includes the assembly if at least one document is matching. In case the `ExcludeAssembliesWithoutSources` parameter is not specified the default value is `MissingAll`. |
+| MissingAny | Includes the assembly only if all documents can be matched to corresponding source files. |
 | None | No assembly is excluded. |
 
 Here is an example of how to specifiy the parameter:
--- a/Documentation/MSBuildIntegration.md
+++ b/Documentation/MSBuildIntegration.md
@@ -232,7 +232,7 @@ To generate deterministc report the parameter is:
 ## Exclude assemblies without sources from coverage
 
 The heuristic coverlet uses to determine if an assembly is a third-party dependency is based on the matching of the assembly`s source documents and the corresponding source files.
-This parameter has three different values to control the this automatic assembly exclusion. 
+This parameter has three different values to control the automatic assembly exclusion. 
 
 | Parameter | Description |
 |-----------|-------------|
@@ -240,7 +240,7 @@ This parameter has three different values to control the this automatic assembly
 | MissingAll | Includes the assembly only if all documents can be matched to corresponding source files. |
 | None | No assembly is excluded. |
 
-Here is an example how specifiy the parameter:
+Here is an example of how to specifiy the parameter:
 ```
 /p:ExcludeAssembliesWithoutSources="MissingAny"
 ```
--- a/Documentation/MSBuildIntegration.md
+++ b/Documentation/MSBuildIntegration.md
@@ -228,3 +228,19 @@ To generate deterministc report the parameter is:
 ```
 /p:DeterministicReport=true
 ```
+
+## Exclude assemblies without sources from coverage
+
+The heuristic coverlet uses to determine if an assembly is a third-party dependency is based on the matching of the assembly`s source documents and the corresponding source files.
+This parameter has three different values to control the this automatic assembly exclusion. 
+
+| Parameter | Description |
+|-----------|-------------|
+| MissingAny | Includes the assembly if at least one document is matching. In case the `ExcludeAssembliesWithoutSources` parameter is not specified the default value is `MissingAny`. |
+| MissingAll | Includes the assembly only if all documents can be matched to corresponding source files. |
+| None | No assembly is excluded. |
+
+Here is an example how specifiy the parameter:
+```
+/p:ExcludeAssembliesWithoutSources="MissingAny"
+```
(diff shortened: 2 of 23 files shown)


Documentation/MSBuildIntegration.md
```diff
@@ -228,3 +228,19 @@ To generate deterministc report the parameter is:
...
+
+## Exclude assemblies without sources from coverage
+
+The heuristic coverlet uses to determine if an assembly is a third-party dependency is based on the matching of the assembly`s source documents and the corresponding source files.
+This parameter has three different values to control the automatic assembly exclusion. 
+
+| Parameter | Description |
+|-----------|-------------|
+| MissingAny | Includes the assembly if at least one document is matching. In case the `ExcludeAssembliesWithoutSources` parameter is not specified the default value is `MissingAny`. |
+| MissingAll | Includes the assembly only if all documents can be matched to corresponding source files. |
```
petli, line 240: The parameters got mixed up in the documentation, this seems to be the intended meaning:

```suggestion
| MissingAll | Includes the assembly if at least one document is matching. In case the `ExcludeAssembliesWithoutSources` parameter is not specified the default value is `MissingAll`. |
| MissingAny | Includes the assembly only if all documents can be matched to corresponding source files. |
```

src/coverlet.core/Helpers/InstrumentationHelper.cs
```diff
@@ -189,25 +182,57 @@ public bool PortablePdbHasLocalSource(string module, out string firstNotFoundDoc
...
-                // Btw check for all possible extension could be weak approach
-                // We exlude from the check the autogenerated source file(i.e. source generators)
-                // We exclude special F# construct https://github.com/coverlet-coverage/coverlet/issues/1145
-                if (!_fileSystem.Exists(docName) && !IsGeneratedDocumentName(docName) &&
-                    !IsUnknownModuleInFSharpAssembly(languageGuid, docName))
+                bool anyDocumentMatches = MatchDocumentsWithSourcesMissingAll(metadataReader);
+                if (!anyDocumentMatches)
                 {
-                    return (false, docName);
+                    _logger.LogVerbose($"Unable to instrument module: {module}, pdb without any local source files");
```
petli, line 193:
```suggestion
                    _logger.LogVerbose($"Excluding module from instrumentation: {module}, pdb without any local source files");
```

src/coverlet.core/Helpers/InstrumentationHelper.cs
```diff
@@ -189,25 +182,57 @@ public bool PortablePdbHasLocalSource(string module, out string firstNotFoundDoc
...
             }
+
+            if (excludeAssembliesWithoutSources.Equals(AssemblySearchType.MissingAny))
+            {
+                (bool allDocumentsMatch, string notFoundDocument) = MatchDocumentsWithSourcesMissingAny(metadataReader);
+
+                if (!allDocumentsMatch)
+                {
+                    _logger.LogVerbose(
+                        $"Unable to instrument module: {module}, pdb without local source files, [{FileSystem.EscapeFileName(notFoundDocument)}]");
```
petli, line 205:
```suggestion
                        $"Excluding module from instrumentation: {module}, pdb without local source files, [{FileSystem.EscapeFileName(notFoundDocument)}]");
```

src/coverlet.core/Helpers/InstrumentationHelper.cs
```diff
@@ -189,25 +182,57 @@ public bool PortablePdbHasLocalSource(string module, out string firstNotFoundDoc
...
+                Document document = metadataReader.GetDocument(docHandle);
+                string docName = _sourceRootTranslator.ResolveFilePath(metadataReader.GetString(document.Name));
+                return (docName, _fileSystem.Exists(docName));
+            });
+        }
+
+        private bool MatchDocumentsWithSourcesMissingAll(MetadataReader metadataReader)
+        {
+            var documentSourceMap = DocumentSourceMap(metadataReader).ToList();
+            return documentSourceMap.Any(x => x.documentExists.Equals(true));
```
petli, line 226: Minor detail, this could work on the enumeration directly, but at the price of making debugging harder:

```suggestion
            return DocumentSourceMap(metadataReader).Any(x => x.documentExists);
```

src/coverlet.core/Helpers/InstrumentationHelper.cs
```diff
@@ -189,25 +182,57 @@ public bool PortablePdbHasLocalSource(string module, out string firstNotFoundDoc
...
+                    return false;
+                }
+            }
+
+            return true;
+        }
+
+        private IEnumerable<(string documentName, bool documentExists)> DocumentSourceMap(MetadataReader metadataReader)
+        {
+            return metadataReader.Documents.ToList().Select(docHandle =>
```
petli, line 215: Is it necessary to do the `ToList()` here?

src/coverlet.core/Helpers/InstrumentationHelper.cs
```diff
@@ -189,25 +182,57 @@ public bool PortablePdbHasLocalSource(string module, out string firstNotFoundDoc
...
+            return documentSourceMap.Any(x => x.documentExists.Equals(true));
+        }
+
+        private (bool allDocumentsMatch, string notFoundDocument) MatchDocumentsWithSourcesMissingAny(
+            MetadataReader metadataReader)
+        {
+            var documentSourceMap = DocumentSourceMap(metadataReader).ToList();
+
+            if (documentSourceMap.Any(x => x.documentExists.Equals(false)))
+                return (false, documentSourceMap.FirstOrDefault(x => x.documentExists.Equals(false)).documentName);
```
petli, line 235: Same as above, can work on the enumerable:

```suggestion
            var missingDoc = DocumentSourceMap(metadataReader).FirstOrDefault(x => !x.documentExists);
            if (missingDoc != null)
            {
                return (false, missingDoc.documentName);
            }
```

test/coverlet.core.tests/Helpers/InstrumentationHelperTests.cs
```diff
@@ -34,23 +35,23 @@ public void TestGetDependenciesWithTestAssembly()
...
 
-            Assert.False(instrumentationHelper.PortablePdbHasLocalSource(typeof(InstrumentationHelperTests).Assembly.Location, out string notFoundDocument));
-            Assert.False(notFoundDocument.IsNullOrEmpty());
+            Assert.False(instrumentationHelper.PortablePdbHasLocalSource(typeof(InstrumentationHelperTests).Assembly.Location, AssemblySearchType.MissingAny));
+            Assert.False(instrumentationHelper.PortablePdbHasLocalSource(typeof(InstrumentationHelperTests).Assembly.Location, AssemblySearchType.MissingAll));
         }
 
         [Fact]
-        public void EmbeddedPortablePDPHasLocalSource_AllDocumentsExist_ReturnsTrue()
+        public void EmbeddedPortablePDPWithAssemblySearchTypeMissingAny_AllDocumentsExist_ReturnsTrue()
```
petli, line 51: Since both `MissingAny` and `MissingAll` are tested:

```suggestion
        public void EmbeddedPortablePDPHasLocalSource_AllDocumentsExist_ReturnsTrue()
```

It would also be good with a testcase where one file is missing, to test the difference in behaviour betweeen the two cases.

petli (commented):
Thanks for having a go at this, @daveMueller! Sorry I didn't have time to review it until now.

It looks good, and I agree with changing the default behaviour and removing the old heuristics for generated code. Might that require a major version bump, since it might change the behaviour of instrumentation for existing projects?

I only have some minor suggestions for code and logging tweaks, and the documentation seems inverted.

daveMueller:
Thanks for the review @petli 🙏 . I'll have a look in the next days.

petli:
Just a note that we shouldn't merge this before a minor fix with #1395 has been released, so people don't have to onboard the new behaviour at the same time.

src/coverlet.core/Helpers/InstrumentationHelper.cs
daveMueller: I'm not so sure about this one. `FirstOrDefault` isn't meant to check whether a value exists or not, although we all (ab)use it this way. In this case `FirstOrDefault` would return a `ValueTuple<string, bool>` which defaults to `(documentName: null, documentExists: false)`.
I now could adapt it to sth. like `if (missingDoc.documenName != null){...}` but now it somehow reads like: "If there is a document whose name can't be resolved lets include the assembly." 
Even tho' this is just because all documents have a matching source file.

petli: I missed the tuple data type here, so ignore this comment :)

MarcoRossignoli:
@daveMueller  we merged this one in past(released on 3.2.0) https://github.com/coverlet-coverage/coverlet/pull/1360 

does this PR invalidate that one?

Documentation/MSBuildIntegration.md
daveMueller: done

src/coverlet.core/Helpers/InstrumentationHelper.cs
daveMueller: done

src/coverlet.core/Helpers/InstrumentationHelper.cs
daveMueller: done

src/coverlet.core/Helpers/InstrumentationHelper.cs
daveMueller: done

src/coverlet.core/Helpers/InstrumentationHelper.cs
daveMueller: done

test/coverlet.core.tests/Helpers/InstrumentationHelperTests.cs
daveMueller: added test with one missing document

src/coverlet.core/Helpers/InstrumentationHelper.cs
daveMueller: simplified the predicate

daveMueller:
> @daveMueller we merged this one in past(released on 3.2.0) #1360
> 
> does this PR invalidate that one right? I think we should add a new document where we list the breaking changes between versions so we can point the user there in case of questions, wdym? We can after link to the main readme as a new section.

Yes you are right. The parameter from #1360 was replaced by setting the `ExcludeAssembliesWithoutSources` to `none`.
https://github.com/coverlet-coverage/coverlet/blob/70fe3191205e01126dbb981a77db787dbf85b3bc/src/coverlet.core/Instrumentation/Instrumenter.cs#L109

Maybe we could add/mark the breaking changes in our changelog like other projects are doing it, e.g. [https://github.com/cake-build/cake/releases/tag/v2.0.0](https://github.com/cake-build/cake/releases/tag/v2.0.0). Or were you thinking about sth. like a table that lists the different versions and their breaking changes?

daveMueller:
I personally would prefer to align the description of the parameter values and make them lower case in these two documentations? Something like `missingall,missingany,none`, what do you think?
https://github.com/coverlet-coverage/coverlet/blob/70fe3191205e01126dbb981a77db787dbf85b3bc/Documentation/VSTestIntegration.md?plain=1#L123
https://github.com/coverlet-coverage/coverlet/blob/70fe3191205e01126dbb981a77db787dbf85b3bc/Documentation/MSBuildIntegration.md?plain=1#L237-L241

MarcoRossignoli:
> Maybe we could add/mark the breaking changes in our changelog like other projects are doing it, e.g. https://github.com/cake-build/cake/releases/tag/v2.0.0. Or were you thinking about sth. like a table that lists the different versions and their breaking changes?

It's fine to me, I was wondering if we have around other BC for the next version, pls bump the version inside the version.json and coverlet.core.csproj like here https://github.com/coverlet-coverage/coverlet/blob/master/Documentation/ReleasePlan.md#how-to-manually-release-packages-to-nugetorg

I remember some issue with the default msbuild property like /Include etc... and the idea to change to /CoverletInclude etc...to avoid clash with "standard msbuild" prop, but I cannot find the issue.

daveMueller:
> I personally would prefer to align the description of the parameter values and make them lower case in these two documentations? Something like `missingall,missingany,none`, what do you think?

I thought about this again and decided to leave it like it currently is (`MissingAll,MissingAny,None`).

daveMueller:
> I remember some issue with the default msbuild property like /Include etc... and the idea to change to `/Coverlet[Coverage]Include` `/CoverletCoverage` etc...to avoid clash with "standard msbuild" prop, but I cannot find the issue.

It's this here:
https://github.com/coverlet-coverage/coverlet/issues/580
https://github.com/coverlet-coverage/coverlet/discussions/1252

Documentation/Changelog.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Breaking changes
+-New parameter ExcludeAssembliesWithoutSources to control automatic assembly exclusion [1164](https://github.com/coverlet-coverage/coverlet/issues/1164)
```
petli, line 10: Can add:
```suggestion
- New parameter `ExcludeAssembliesWithoutSources` to control automatic assembly exclusion [1164](https://github.com/coverlet-coverage/coverlet/issues/1164). The parameter `InstrumentModulesWithoutLocalSources` has been removed. since it can be handled by setting `ExcludeAssembliesWithoutSources` to `None`.
- The default heuristics for determining whether to instrument an assembly has been changed. In previous versions any missing source file was taken as a signal that it was a third-party project that shouldn't be instrumented, with exceptions for some common file name patterns for source generators. Now only assemblies where no source files at all can be found are excluded from instrumentation, and the code for detecting source generator files have been removed. To get back to the behaviour that at least one missing file is sufficient to exclude an assembly, set `ExcludeAssembliesWithoutSources` to `MissingAny`, or use assembly exclusion filters for more fine-grained control.
```

petli (approved):

MarcoRossignoli:
Thanks @petli @daveMueller going to merge!

PR #1444, opened 2023-02-08
Update collector docs with info about passing format without runsettings file

### LINKED ISSUE #1441: Clarify how to supply format on the command line in the documentation
The [documentation](https://github.com/coverlet-coverage/coverlet/blob/master/Documentation/VSTestIntegration.md#default-option-if-you-dont-specify-a-runsettings-file) specifies that the only option that one can provide without using a dedicated `coverlet.runsettings` file is the **Format** option. Unfortunately the documentation fails to detail exactly how to use this option on the command line.

Can this be amended to the section above the options table, please?

daveMueller:
closes #1441 

I tried out all the documented combinations.

grafik

grafik

### DIFF
--- a/Documentation/VSTestIntegration.md
+++ b/Documentation/VSTestIntegration.md
@@ -77,9 +77,20 @@ We're working to fill the gaps.
 
 ### Default option (if you don't specify a runsettings file)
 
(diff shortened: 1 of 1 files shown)


petli (approved):
👍

PR #1465, opened 2023-04-05
Bump versions

MarcoRossignoli:


daveMueller:
I think we need to adapt the azure devops pipeline to install .NET Core SDK 6.x.xxx

MarcoRossignoli:
Yep forgot it

Issue #1505, opened 2023-07-28
Package 'Newtonsoft.Json' 9.0.1 has a known moderate severity (coverlet.core)


Bertk:
Project coverlet.core uses [vulnerable Package 'Newtonsoft.Json 9.0.1](https://github.com/advisories/GHSA-5crp-9r3c-p9vr)'. This assembly is also added coverlet Nuget packages.

<img width="624" alt="image" src="https://github.com/coverlet-coverage/coverlet/assets/2315215/428ac34e-36fb-47cd-a9fb-7d375d22dcdd">

The version is defined in [Directory.Build.targets](https://github.com/coverlet-coverage/coverlet/blob/master/Directory.Build.targets) with comment "Do not upgrade".
```xml
    <!-- Do not upgrade this version or we won't support old SDK -->
    <PackageReference Update="Newtonsoft.Json" Version="9.0.1" />
    <PackageReference Update="NuGet.Packaging" Version="5.4.0" />
    <PackageReference Update="ReportGenerator.Core" Version="4.6.0" />
```

When will the **Newtonsoft.Json** package version be upgraded to **V13.0.3**?

SeijiSuenaga:
@Bertk do you know when a version of Coverlet with this security patch might be published?  I see the PR was merged 5 months ago, but the latest published version 6.0.0 was 9 months ago.  Sorry for the ping if the plan is already documented somewhere, but I couldn't find it.

Bertk:
coverlet nuget packages will be released soon. By,the way the Newtonsoft.Json package will be replaced (https://github.com/coverlet-coverage/coverlet/pull/1513).

Issue #1361, opened 2022-07-11
Newtonsoft.Json vulnerability to DoS attacks in versions before 13.0.1


SiwinskiK:
Newtonsoft.Json prior to version 13.0.1 is vulnerable to Insecure Defaults due to improper handling of StackOverFlow exception (SOE) whenever nested expressions are being processed. Exploiting this vulnerability results in Denial Of Service (DoS), and it is exploitable when an attacker sends 5 requests that cause SOE in time frame of 5 minutes. This vulnerability affects Internet Information Services (IIS) Applications. The vulnerability is severe and renders coverlet unusable.
Coverlet in version 2.0.9 references an older version of aforementioned library. Upgrade to version Newtonsoft.Json - 13.0.1

petli:
@SiwinskiK coverlet packages should only be referenced by unit test projects, which generally ought not to be run in IIS, at least not in production. Given that, is this vulnerability still relevant for coverlet? I think, but am not sure, that the older version of Newtonsoft.Json is referenced because of compatiblity with .NET Framework.

MarcoRossignoli:
@petli is true, anyway we're running inside CI so I think that the bump will be good, I don't know but I think that we'll need to bump also the object model used by collectors because test platform already did the bump and so we could break something but it's ok.

Issue #1332, opened 2022-05-09
datacollector.exe Unable to instrument module: Could not load file or assembly 'System.Reflection.Metadata, Version=1.4.2.0'


thoma01:
Test project created with xUnit v2.4.1 and Microsoft.NET.Test.Sdk v17.1.0 and coverlet.collector v3.1.2

Error from log.datacollector.txt
```
TpTrace Information: 0 : 17336, 1, 2022/05/09, 14:42:45.797, 23045101538449, datacollector.exe, AssemblyResolver.OnResolve: System.Reflection.Metadata, Version=1.4.2.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a: Resolving assembly.
TpTrace Information: 0 : 17336, 1, 2022/05/09, 14:42:45.797, 23045101539974, datacollector.exe, AssemblyResolver.OnResolve: System.Reflection.Metadata, Version=1.4.2.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a: Searching in: 'REST.Test\publish'.
TpTrace Information: 0 : 17336, 1, 2022/05/09, 14:42:45.797, 23045101544437, datacollector.exe, AssemblyResolver.OnResolve: System.Reflection.Metadata, Version=1.4.2.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a: File exists but version/public key is wrong. Try next extension.
TpTrace Information: 0 : 17336, 1, 2022/05/09, 14:42:45.797, 23045101545465, datacollector.exe, AssemblyResolver.OnResolve: System.Reflection.Metadata, Version=1.4.2.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a: Assembly path does not exist: 'publish\System.Reflection.Metadata.exe', returning.
TpTrace Information: 0 : 17336, 1, 2022/05/09, 14:42:45.797, 23045101545985, datacollector.exe, AssemblyResolver.OnResolve: System.Reflection.Metadata, Version=1.4.2.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a: Searching in: 'C:\Program Files\Microsoft Visual Studio\2022\Professional\Common7\IDE\Extensions\TestPlatform\Extensions'.
[...]

PR #1510, opened 2023-07-31
Update Newtonsoft.Json  version and other nuget versions (use CPM)

### LINKED ISSUE #1351: Update Mono.Cecil dependency version to 0.11.4
Hello.

The latest (at the time of this post - 3.1.2) version of Coverlet uses Mono.Cecil 0.11.1. Unfortunately, this is an outdated version of Mono.Cecil that contains some bugs. For example, const null arrays are not processed correctly:
```
using Mono.Cecil;
using Mono.Cecil.Rocks;

await using var moduleStream = File.OpenRead(typeof(TestClass).Assembly.Location);

using var module = ModuleDefinition.ReadModule(moduleStream, new ReaderParameters { ReadSymbols = true });
var type = module.GetType(string.Empty, nameof(TestClass));
var method = type.GetMethods().Single();

try
{
    var debugInformation = method.DebugInformation; // fails on Mono.Cecil 0.11.1, works fine on 0.11.4
    Console.WriteLine("Success");
}
catch (Exception ex)
{
    Console.WriteLine("Error: {0}", ex);
}

public class TestClass
{
    public void TestMethod()
    {
        const object[]? testVariable = null;
    }
}
```
This prevents Coverlet from instumenting modules and calculating code coverage correctly.

It would be nice if you could update Mono.Cecil dependency to 0.11.4. Thanks in advance.

Bertk:
- Update version of Newtonsoft.Json (https://github.com/coverlet-coverage/coverlet/issues/1505)
- Update documentation (TFM and nuget package versions )
- Remove unnecessary UseDotNet tasks from pipeline
- Remove unnecessary using statements
- Use Central Package Management (CPM)
- resolves #1351 #1505 #1361 #1332

alternative: #1513


Bertk:
@MarcoRossignoli @daveMueller Please review PR.

eng/build.yml
```diff
@@ -1,14 +1,4 @@
...
-- task: UseDotNet@2
-  inputs:
-    version: 3.1.404
-  displayName: Install .NET Core SDK 3.1.404
-
-- task: UseDotNet@2
-  inputs:
-    version: 5.0.401
-  displayName: Install .NET Core SDK 5.0.401
-
```
daveMueller, line 11: @Bertk Thanks a lot for your time-consuming contributions 🙏 . I've seen your comment in [https://github.com/coverlet-coverage/coverlet/issues/1511#issuecomment-1663517315](https://github.com/coverlet-coverage/coverlet/issues/1511#issuecomment-1663517315) regarding this topic but PR was already merged at that time. I think you are right that we don't need them anymore. Especially if we migrate all the test projects to `net6.0` like you have done in this PR.
Personally I'm not sure if we shouldn't keep 1-2 legacy `netcoreapp3.1` projects to have a wider range of test scenarios regarding instrumentation... What do you think @MarcoRossignoli?
Thanks again, I really like the changes in this PR and will try to do a proper review in the next days.

MarcoRossignoli: I suggest to test and ship only supported versions announced here https://learn.microsoft.com/en-us/lifecycle/products/microsoft-net-and-net-core

Bertk: This is shows ***supported versions*** much better :wink:
https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core

MarcoRossignoli:
@daveMueller  is this good to go?

daveMueller:
> @daveMueller is this good to go?

Sorry didn't find time yet to look into it. Most likely on the weekend.

eng/publish-coverlet-result-files.yml
```diff
@@ -0,0 +1,44 @@
...
+- task: Powershell@2
+  displayName: Prepare coverage files to Upload
+  inputs:
+    targetType: inline
+    pwsh: true
+    script: |
+      New-Item -ItemType Directory "$(Build.SourcesDirectory)/artifacts/TestLogs/"
+      function Copy-FileKeepPath {
+      param (
+          $filter,$FileToCopy,$des,$startIndex
```
daveMueller, line 11: Can we stay consistent with the naming of the parameters. I think the default in powershell is capitalized e.g. `$Filter, $FileToCopy, $Destination, $StartIndex`.

eng/publish-coverlet-result-files.yml
```diff
@@ -0,0 +1,44 @@
...
+          )
+      Get-ChildItem -Path $FileToCopy -Filter $filter -Recurse -File | ForEach-Object {
+          $fileName = $_.FullName
+          #Remove the first part to ignore from the path.
+          $newdes=Join-Path -Path $des -ChildPath $fileName.Substring($startIndex)
+          $folder=Split-Path -Path $newdes -Parent
+          $err=0
+
+          #check if folder exists"
+          $void=Get-Item $folder -ErrorVariable err  -ErrorAction SilentlyContinue
```
daveMueller, line 21: I'm not that familiar with azure devops pipelines but what is this `$void` variable for? In general this seems a bit complicated to me. What about
```powershell
      if (!(Test-Path -Path $folder)) {New-Item $folder -Type Directory -Force -Verbose}
      Copy-Item -Path $fileName -Destination $newdes -Recurse -Force -Verbose
```

eng/publish-coverlet-result-files.yml
```diff
@@ -0,0 +1,44 @@
...
+            #create when it doesn't
+            $void=New-Item -Path $folder -ItemType Directory -Force -Verbose
+          }
+
+          $void=Copy-Item -Path $fileName -destination $newdes -Recurse -Container -Force -Verbose
+          }
+      }
+      $logfiles = "coverage.opencover.xml", "coverage.cobertura.xml", "coverage.json", "log.txt", "log.datacollector.*.txt", "log.host.*.txt"
+      foreach ($logfile in $logfiles ) {
+      Copy-FileKeepPath -FileToCopy "$(Build.SourcesDirectory)/test/coverlet.integration.tests/bin/*" -des "$(Build.SourcesDirectory)/artifacts/TestLogs/" -filter $logfile -startIndex "$(Build.SourcesDirectory)/test/coverlet.integration.tests/bin/".Length
```
daveMueller, line 32: Why collect those logs only for `coverlet.integration.tests`? I think it would be beneficial to collect them from all test projects.

eng/publish-coverlet-result-files.yml
```diff
@@ -0,0 +1,44 @@
...
+    script: |
+      New-Item -ItemType Directory "$(Build.SourcesDirectory)/artifacts/TestLogs/"
+      function Copy-FileKeepPath {
+      param (
+          $filter,$FileToCopy,$des,$startIndex
+          )
+      Get-ChildItem -Path $FileToCopy -Filter $filter -Recurse -File | ForEach-Object {
+          $fileName = $_.FullName
+          #Remove the first part to ignore from the path.
+          $newdes=Join-Path -Path $des -ChildPath $fileName.Substring($startIndex)
```
daveMueller, line 16: Personally I'm not a big fan of abbreviations (`$newdes`). And are those 3 inline comments really needed for 5 lines powershell?

eng/build.yml
```diff
@@ -34,3 +24,11 @@ steps:
     command: test
     arguments: -c $(BuildConfiguration) --no-build /p:CollectCoverage=true /p:CoverletOutputFormat=opencover /p:Include="[coverlet.collector]*%2c[coverlet.core]*%2c[coverlet.msbuild.tasks]*" /p:Exclude="[coverlet.core.tests.samples.netstandard]*%2c[coverlet.tests.xunit.extensions]*"
     testRunTitle: $(Agent.JobName)
+
+- template: publish-coverlet-result-files.yml
```
daveMueller, line 28: I really like that we collect logs here etc... Will make it easier to troubleshoot issues that only happen on the build server. ❤️ 👍

eng/build.yml
```diff
@@ -34,3 +24,11 @@ steps:
...
+
+#- template: publish-coverage-results.yml
```
daveMueller, line 30: Why is this commented? I also like this idea of having an html report that can be looked into. Unfortunately I don't have much control over the pipeline and thus I can't tell anything about file restrictions...

Directory.Packages.props
```diff
@@ -0,0 +1,58 @@
...
+    <CentralPackageTransitivePinningEnabled>true</CentralPackageTransitivePinningEnabled>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <GlobalPackageReference  Include="Microsoft.SourceLink.GitHub" Version="1.1.1" />
+    <GlobalPackageReference  Include="Nerdbank.GitVersioning" Version="3.6.133" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageVersion Include="McMaster.Extensions.CommandLineUtils" Version="2.3.4" />
```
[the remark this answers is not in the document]
Bertk: This is correct. I will remove it ;-)

Directory.Packages.props
```diff
@@ -0,0 +1,58 @@
...
+         vstest 17.8 version
+         NuGetFrameworksVersion is defined here https://github.com/microsoft/vstest/blob/9a0c41811637edf4afe0e265e08fdd1cb18109ed/eng/Versions.props#L94C1-L94C1
+    -->
+    <PackageVersion Include="NuGet.Frameworks" Version="5.11.5" />
+    <PackageVersion Include="Mono.Cecil" Version="0.11.5" />
+    <PackageVersion Include="Moq" Version="4.18.4" />
+    <PackageVersion Include="Newtonsoft.Json" Version="13.0.1" />
+    <PackageVersion Include="NuGet.Packaging" Version="6.6.1" />
+    <PackageVersion Include="ReportGenerator.Core" Version="5.1.23" />
+    <PackageVersion Include="LinqKit.Microsoft.EntityFrameworkCore" Version="7.1.4" />
```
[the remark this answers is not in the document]
Bertk: The version for this packages are explicitly configured in the csproj file with `VersionOverride` and if the versions are not significant the comments can be moved and VersionOverride removed.
https://github.com/Bertk/coverlet/blob/4ebed44c91b3d36464b59928c5dee022826904ea/test/coverlet.core.tests/coverlet.core.tests.csproj#L37C1-L42C77 

I did not look into this.

Documentation/Examples/MSBuild/DeterministicBuild/HowTo.md
```diff
@@ -32,22 +33,24 @@ Copyright (C) Microsoft Corporation. All rights reserved.
...
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.5.0" />
-    <PackageReference Include="xunit" Version="2.4.0" />
-    <PackageReference Include="xunit.runner.visualstudio" Version="2.4.0" />
-    <PackageReference Include="coverlet.collector" Version="1.2.0" />
-    <!-- version comes from  coverlet.msbuild.2.9.0-preview.6.ga0e22ec622.nupkg -->
-    <PackageReference Include="coverlet.msbuild" Version="2.9.0-preview.6.ga0e22ec622" >
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.5.0" />
+    <PackageReference Include="xunit" Version="2.5.0" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.5.0" />
+    <PackageReference Include="coverlet.collector" Version="6.0.0" />
```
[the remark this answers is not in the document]
Bertk: Removed ` <PackageReference Include="coverlet.collector" Version="6.0.0" />`

eng/build.yml
Bertk: The template uses `reportgenerator@5` task which is currently not available on Azure DevOps **tonerdo** organization
(see #1516).

Directory.Build.props
```diff
@@ -14,6 +14,7 @@
     <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
     <LangVersion>preview</LangVersion>
     <NoWarn>$(NoWarn);NU5105;CS1591</NoWarn>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
```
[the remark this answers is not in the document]
Bertk: I added this because `unnecessary using` check require it. By the way, the new .NET SDK 8 generates warnings for `unnecessary using` which will break the build in CI.

eng/publish-coverlet-result-files.yml
Bertk: $void is used to suppress some PSScriptAnalyzer warnings. I like your 2 lines but there is a small deatsil. The folder is not created if it already exists. I will check this with a separte commit.

test/coverlet.core.tests/coverlet.core.tests.csproj
```diff
@@ -30,7 +30,10 @@
     <ProjectReference Include="$(RepoRoot)test\coverlet.core.tests.samples.netstandard\coverlet.core.tests.samples.netstandard.csproj" />
     <ProjectReference Include="$(RepoRoot)test\coverlet.tests.xunit.extensions\coverlet.tests.xunit.extensions.csproj" />
     <ProjectReference Include="$(RepoRoot)test\coverlet.tests.projectsample.netframework\coverlet.tests.projectsample.netframework.csproj" />
-    <ProjectReference Include="$(RepoRoot)test\coverlet.tests.projectsample.vbmynamespace\coverlet.tests.projectsample.vbmynamespace.vbproj" />
+  </ItemGroup>
+  
+  <ItemGroup Condition="$([MSBuild]::IsOSPlatform('windows'))">
+      <ProjectReference Include="$(RepoRoot)test\coverlet.tests.projectsample.vbmynamespace\coverlet.tests.projectsample.vbmynamespace.vbproj" />
```
daveMueller, line 36: Why did you add this condition here? It is possible to compile the assembly on linux.
grafik

Directory.Packages.props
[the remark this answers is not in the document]
daveMueller: Those packages are only used for tests and are not really relevant for `coverlet` at all. Thus the version number can also be moved to `Directory.Packages.props`. In addition I don't think the version is significant for those tests and updating shouldn't be a problem if we need one of the packages some where else.

test/coverlet.core.tests/coverlet.core.tests.csproj
Bertk: coverlet.tests.projectsample.vbmynamespace.dll targets net48 platform which is only available for windows.  Therefore compilation is not necessary for Linux and MacOS. 

```Text
Testhost process for source(s) '/home/vsts/work/1/s/test/coverlet.tests.projectsample.vbmynamespace/bin/Debug/net48/coverlet.tests.projectsample.vbmynamespace.dll' exited with error: Unhandled Exception:
System.TypeInitializationException: The type initializer for 'Microsoft.VisualStudio.TestPlatform.ObjectModel.EqtTrace' threw an exception. ---> System.TypeLoadException: Could not load type of field 'Microsoft.VisualStudio.TestPlatform.ObjectModel.PlatformEqtTrace:TraceSourceLevelsMap' (1) due to: Could not resolve type with token 01000019 from typeref (expected class 'System.Diagnostics.TraceLevel' in assembly 'System, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089') assembly:System, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089 type:System.Diagnostics.TraceLevel member:(null)
   --- End of inner exception stack trace ---
  at Microsoft.VisualStudio.TestPlatform.TestHost.Program.Main (System.String[] args) [0x00033] in <985811b3ce274022a2072f51d4b88753>:0 
[ERROR] FATAL UNHANDLED EXCEPTION: System.TypeInitializationException: The type initializer for 'Microsoft.VisualStudio.TestPlatform.ObjectModel.EqtTrace' threw an exception. ---> System.TypeLoadException: Could not load type of field 'Microsoft.VisualStudio.TestPlatform.ObjectModel.PlatformEqtTrace:TraceSourceLevelsMap' (1) due to: Could not resolve type with token 01000019 from typeref (expected class 'System.Diagnostics.TraceLevel' in assembly 'System, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089') assembly:System, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089 type:System.Diagnostics.TraceLevel member:(null)
   --- End of inner exception stack trace ---
  at Microsoft.VisualStudio.TestPlatform.TestHost.Program.Main (System.String[] args) [0x00033] in <985811b3ce274022a2072f51d4b88753>:0 

```

daveMueller: OK so `net48` can only be run on windows, but you can build a `net48` assembly on linux. The test `Instrumenter_MethodsWithoutReferenceToSource_AreSkipped ` also doesn't run this assembly, it only instruments it. And thus the test also executes on linux.

Sure it currently isn't a real scenario that `coverlet` instruments on a different operating system than the test assemblies are being executed. But technically it would be possible.

Nevertheless this wasn't really the objective here. In the end we wanted to test a scenario that an assembly/symbole contains a method definition that has no reference to a source (https://github.com/coverlet-coverage/coverlet/blob/master/src/coverlet.core/Instrumentation/Instrumenter.cs#L533-L535). This is a rather odd scenario but this `visual basic` `net48` assembly contains exactly such a scenario. So the test wasn't only meant for a `net48` assembly executed on windows.

This often is the way we analyze/fix bugs. Create an assembly that has the **problematic** construct inside, make coverlet that resilient that it can handle the problem and in the end write a unit test that checks the correct instrumentation. We actually do this because it is a pretty fast process. But I also agree that this process of writing tests could be optimized. E.g. by writing more actual unit tests instead of integration tests.

As this test currently doesn't make any problems on the other operating systems, can we agree to just leave it as it is?

Bertk: I removed the condition but the rational should be documented somewhere.

daveMueller (approved):

PR #1522, opened 2023-08-13
add issue-untriaged.yml and issue-inactive.yml

Bertk:
- automatically close inactive issues after 12 month period
- automatically add labels "triage" and "stale"
- run issue-inactive.yml on demand and scheduled

❗ issue-inactive.yml action uses variable `secrets.GITHUB_TOKEN` 

refine options:
- https://github.com/actions/stale#any-of-issue-labels
- https://github.com/actions/stale#exempt-issue-labels


Bertk:
Latest Visual Studio versions complain about "restore loop" in **coverlet.core.tests**. 

image

<br><br>Reporters.cs should be relocated to a separate unit test for **coverlet.msbuild.tasks** tests. 
I will create an issue for this activity.

.github/workflows/issue-triage.yml
```diff
@@ -0,0 +1,28 @@
+# label management triage | good first issue
+#                      -> untriaged|duplicate|invalid
+#                         ->bug|enhancement|feature-request|documentation|question|wontfix|Known issue
+#                           -> up-for-grabs | in progress | with repo | needs repo | tracking-external-issue
+# https://github.com/coverlet-coverage/coverlet/blob/master/Documentation/Roadmap.md
```
daveMueller, line 5: I think those comments don't really describe our current workflow. Usually we add the label `untriaged` to new issues and then adjust the labels if somebody found time to analyze it. If you wan't to change the current workflow then we should discuss it and document it consistently. E.g. the workflow in the comment doesn't really match the description in the roadmap document.

Directory.Packages.props
```diff
@@ -50,6 +50,7 @@
     System.TypeInitializationException : The type initializer for 'Tmds.Utils.ExecFunction' threw an exception.
     System.NotSupportedException : Application is running as testhost, unable to determine parent 'dotnet' process.
     -->
+    <PackageVersion Include="System.Text.Json" Version="7.0.3" />
```
daveMueller, line 53: I think this isn't needed in this PR yet?

.github/workflows/issue-triage.yml
Bertk: I will keep the list of labels and the (Roadmap.md )(https://github.com/coverlet-coverage/coverlet/blob/master/Documentation/Roadmap.md) URL . This document has some details for labels but not for all of them.

By the way, I can see the list of labels only since I am a co-maintainer.

Directory.Packages.props
Bertk: Removed `PackageVersion`. This was required before I removed the unnecessary `PrivateAssets="All"`.

daveMueller (approved):
Thanks a lot @Bertk 👍

PR #1530, opened 2023-08-30
more coverlet.msbuild.tasks tests

Bertk:
- increase source code coverage
- unit test for coverlet.msbuild.tasks
- use ILogger instead of Console.Write


Bertk:
The updates allow to run coverlet tests code coverage (first time without errors) but the CI fails ...

`dotnet test /p:CollectCoverage=true /p:CoverletOutputFormat=opencover`

image

Bertk:
@daveMueller @MarcoRossignoli I need some help and maybe you have already experience with this situation.

- The local build has no error but the CI fails and the test design is a little bit tricky. 
- The`Assert.Single(prepareResult.Results)` statement in **InstrumenterHelper.cs:line 318** fails.  
 
Any idea how to get more details?

Bertk:
Replaced with PR #1534

PR #1532, opened 2023-09-03
Update issue-inactive.yml

Bertk:
set operations-per-run = 100


Bertk:
The issue-inactive action did only a partially check and should be executed multiple times until all 137 issues are processed.

daveMueller (approved):

daveMueller:
> By the way, I am not able to start the action manually

I can start it manually. Maybe @MarcoRossignoli can check what permission you need to start actions.

PR #1541, opened 2023-10-10
fix deadlog in BaseTest

Bertk:
- 
- use project names for test artifacts (trx, html) 
- use CopyFiles instead of powershell script (publish-coverlet-result-files.yml)
- disable CPM for coverlet.core.tests.csproj (VS reports "restore loop")


daveMueller (approved):
All cool, LGTM 👍

PR #1551, opened 2023-11-11
add SignClient (pwsh)

Bertk:
:warning: Do not merge without pipeline tests

A variable group has to be created which has the secrets for code signing. This variables use :lock: and are not visible.

image

By the way, the parameter `/p:PublicRelease=xxxx` might not be active in this build pipeline. I think this is only active for build pipelines using using _Microsoft Arcade build tooling_.

=> Why not always sign the nuget packages created by nightly build ?!?

**Does not include:**
> Upload *.nupkg files to Nuget.org site. Check all metadata(url links, deterministic build etc...) before "Submit"


MarcoRossignoli:
> Why not always sign the nuget packages created by nightly build ?!?

We could do that but at the moment the rate of release is not huge we're doing it 3/5 per year and less we put that pwd around better is imo, BUT for start we could have this pipeline that we can run "manually" with some "enable" for the step and we "also" sign and publish the version inside the artifacts in case of "official release". We need to understand if worth it or not the "manual" release workflow is matter of 15/20 today.

PR #1559, opened 2023-11-22
use .NET SDK 8.0 for tests

Bertk:
- use net8.0 for tests
- use [Simplified output paths](https://learn.microsoft.com/en-us/dotnet/core/sdk/artifacts-output)
- improve log-data for deterministic test
- coverlet.msbuild.task targets netstandard2.0 and net6.0
- activate workaround [MSBUILDDISABLENODEREUSE=1](https://github.com/dotnet/sdk/issues/9452)


src/coverlet.msbuild.tasks/coverlet.msbuild.tasks.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Library</OutputType>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0;net6.0</TargetFrameworks>
```
MarcoRossignoli, line 5: is there a reason for multitfm here?

MarcoRossignoli:
Some question but generally looks good wait for @daveMueller double check

src/coverlet.msbuild.tasks/coverlet.msbuild.tasks.csproj
Bertk: The tasks for net6.0 only has 1 system.* dependency "System.Text.Json"  and `coverlet.msbuild.tasks.deps.json` uses net6.0 runtime. 

image

image

Bertk: Yes, this can be done. We only need to add target net6.0 and use a ItemGroup with condition for netstandard2.0.
This ItemGroup will have the PackageReference for System.Reflection.Metadata and System.Collections.Immutable for coverlet.collector (vstest compatibility).

src/coverlet.collector/lib/netstandard2.0/_._
daveMueller, line 1: What's with this? I guess just some leftover?

src/coverlet.core/Helpers/SourceRootTranslator.cs
```diff
@@ -48,6 +48,8 @@ public SourceRootTranslator(string moduleTestPath, ILogger logger, IFileSystem f
       string assemblyName = assemblyAdapter.GetAssemblyName(moduleTestPath);
       _mappingFileName = $"CoverletSourceRootsMapping_{assemblyName}";
 
+      // _logger.LogInformation($"_mapping file name: '{_mappingFileName}'", true);
```
daveMueller, line 51: I would suggest to uncomment this as it could be usefull for troubleshooting?

src/coverlet.core/Helpers/SourceRootTranslator.cs
```diff
@@ -94,9 +96,11 @@ public SourceRootTranslator(string moduleTestPath, ILogger logger, IFileSystem f
           _logger.LogWarning($"Malformed mapping '{mappingRecord}'");
           continue;
         }
+#pragma warning disable IDE0057 // Use range operator
         string projectPath = mappingRecord.Substring(0, projectFileSeparatorIndex);
         string originalPath = mappingRecord.Substring(projectFileSeparatorIndex + 1, pathMappingSeparatorIndex - projectFileSeparatorIndex - 1);
         string mappedPath = mappingRecord.Substring(pathMappingSeparatorIndex + 1);
+#pragma warning restore IDE0057 // Use range operator
```
daveMueller, line 103: I see disabling this warning in a couple of files. Can we adapt `.editorconfig` instead?

src/coverlet.msbuild.tasks/coverlet.msbuild.tasks.csproj
daveMueller: Does this also make sense for the `collector`?

test/coverlet.core.tests/Coverage/CoverageTests.AutoProps.cs
```diff
@@ -105,7 +106,7 @@ public void SkipAutoPropsInRecords(bool skipAutoProps)
       }
     }
 
-    [Theory]
+    [Theory(Skip = "fails reason unknown (no session debug possible)")]
```
daveMueller, line 109: I think I know what the problem is. Will fix it once this PR is merged.

src/coverlet.core/Helpers/SourceRootTranslator.cs
Bertk: Yes, this could be done but this will hide also new findings. The range operators are pretty need and with the current source suppression this is easy to find and could be resolved in another PR. This one is already huge.

src/coverlet.msbuild.tasks/coverlet.msbuild.tasks.csproj
Bertk: Yes, this can also be done for coverlet.collector 👍 but should be done in a separate PR. This requires an update of csproj 
 and coverlet.collector.targets file. I will create a PR after this one is merged (currently: framework 6.0).

> Extend DataCollector
The very first thing you will need to create is a Class Library project and add reference to Microsoft.TestPlatform.ObjectModel nuget package. Class Library project can target Desktop clr or dotnet core clr or both frameworks.

daveMueller (approved):
Thanks, LGTM 👍

PR #1575, opened 2024-01-01
Fix issues with primary constructors in records

### LINKED ISSUE #1555: Uncovered lines in .NET 8
**Describe the bug**
When creating a record inheriting another record in a .NET 8 project, the initial line containing `public record` is reported as not covered by tests even though the record is used in the tests.

**To Reproduce**
1. Create a record.
2. Create another record inheriting that record.
3. Create a test using the second record.
4. Run the test with test coverage.
5. The line containing `public record` on the second record is reported as not tested.

Here is an example project showcasing the problem: https://github.com/samuel-utbult-oborgen/coverlet-missing-line-example

**Expected behavior**
The uncovered line should either be covered or ignored.

**Actual behavior**
The line is reported as uncovered.

**Configuration**
The following environment it used:
- coverlet 6.0.0
- .NET 8
- macOS Sonoma 14.0
- x64
- I have not encountered this problem in .NET 7 but coverlet 3.2.0 seems to have the same problem.

### LINKED ISSUE #1561: C# record constructors are not covered when SkipAutoProps is true
**Describe the bug**
C# record constructors are marked as uncovered whenever _SkipAutoProps_ is true. But if _SkipAutoProps_ is false, the coverage is correct.

**To Reproduce**

1. Create a file with a simple record
```public record MyRecord(int A);```

2. Create a test that constructs the record
```var myRecord = new MyRecord(1);```

Minimal Example: https://github.com/jlawrence/coverlet-missing-line-example/tree/main

**Expected behavior**
The record constructor should be covered regardless of the value of _SkipAutoProps_.

**Actual behavior**
Observed behavior: If _SkipAutoProps_ is _true_, the record constructor will not be covered. If it is _false_, the record constructor will be covered.

**Configuration:**
    * Coverlet package: "coverlet.collector" Version="6.0.0"
    * .NET 8
    * Windows 10 x64

daveMueller:
closes #1555, closes #1561


daveMueller:
Lets merge PR #1559 first. Then no changes to the test are needed.

MarcoRossignoli (approved):
LGTM(testing is failing)

Issue #1579, opened 2024-01-08
Generate CHANGELOG.md using changelog-from-release tool


Bertk:
The CHANGELOG.md can be generated using available tools and should be updated within the PR. This might require addition conventions for PR and git commits. 

- [**changelog-from-release**](https://github.com/rhysd/changelog-from-release) 
  -  uses git releases to generate CHANGELOG.md
  -  GitHub action [rhysd/changelog-from-release/action@v3](https://github.com/rhysd/changelog-from-release/blob/master/action/README.md)
- [**Versionize**](https://github.com/versionize/versionize) using [conventional commit messages](https://conventionalcommits.org/).
   - This might require addition conventions for PR and git commits. 
   -  Versionize is a dotnet tool 
- [**github-changelog-generator**](https://github.com/github-changelog-generator/github-changelog-generator)
  - allows to add existing manual created CHANGELOG.md 
- ...

There are many tools and my favored for coverlet is [**changelog-from-release**]. I created a CHANGELOG.md for coverlet with this tools (see attachment) with command `changelog-from-release -r https://github.com/coverlet-coverage/coverlet`.

_Proposal:_ 
Create GitHub action which created CHANGELOG.md  and upload the file as build artifact. This can be used for release activity or using draft releases even for PR.

@daveMueller @MarcoRossignoli 
Please share your thoughts on this topic .  

[CHANGELOG.md](https://github.com/coverlet-coverage/coverlet/files/13858058/CHANGELOG.md)

see also: 

- https://github.com/marketplace/actions/release-drafter

Draft release is updated every time the pipeline with the release-drafter action is called. 🚀 
<img width="1035" alt="image" src="https://github.com/coverlet-coverage/coverlet/assets/2315215/8dc5cba7-a910-49d0-a157-ed962063496d">

MarcoRossignoli:
looks interesting, @daveMueller what you think?

daveMueller:
Nothing against that. As long as it is just `release-drafter` and `changelog-from-release` and I don't need to look into additional tools to fix something when it breaks. What I think is a bit annoying that you need `changelog-from-release` to create the markdown as `release-drafter` doesn't have a option to also create it.
What I'm also missing a bit is the option to manually add something to the markdown, e.g. description of breaking changes etc... Do I then need to update the draft release first with my manual information so that it is included in the changelog?

Bertk:
[...]

PR #1585, opened 2024-01-12
create draft release (triggered by master branch update)

Bertk:
This pipeline will reduce the effort for release documentation (#1579)

Chapters of release note are specified in configuration file `.github\release-drafter.yml` and PR labels are used to populate release note chapters. 

Current label list of release-drafter.yml:
    - **breaking-change**
    - **enhancement**
    - **bug**
    - maintenance
    - chore
    - **documentation**
    - dependencies
    - security
    - skip-changelog
   
 _Bold labels already exist in [coverlet repository](https://github.com/coverlet-coverage/coverlet/labels)_


daveMueller:
I think with release drafter we would be doing a fundamental switch on how we document our changes. Currently we focus on Github issues that we closed with our changes. But we don't reference the actual PR in the change log. 
Release drafter does this differently. It references PRs without an explicit reference to the issues a PR closes. This means for us we must be much more precise on the title of PRs and also start tagging our PRs.

Personally I like it when there is an issue to a PR because it contributes to transparency for the community. When I work in other open source projects I often wonder about commits that I can't link to an issue. Release drafter doesn't exclude this option as issues can still be linked to PRs, but we need to be more careful with our PRs.

I think release drafter looks interesting and we can give it a try. But we also need to document the workflow for creating PRs then. On the other hand, I don't think maintaining our manual change log is much work. Let's see what @MarcoRossignoli thinks about it.

Bertk:
I used  issues as a source for release notes in the past. We created software for a regulated environment and followed internal quality regulations. There was no PR without a issue with defined content. This is not required here and not all PRs need a issue e.g. maintenance/chore.

**release drafter** will create information based on PRs but you are allowed to ignore it. 
- Nevertheless it creates a nice overview and which kann be reused for CHANGELOG.md.
- The content can be tweaked during development using labels.

Issue #1583, opened 2024-01-10
[BUG] Coverlet Nightly Crashing when merging JSON results 


pinkfloydx33:
**Describe the bug**
When running the coverlet nightly builds the generation of results fails when combining multiple test runs using JSON format. Individual runs generate reports successfully. The exception thrown is:

```
/home/AzDevOps/.nuget/packages/coverlet.msbuild/6.0.1-preview.31.g0db8de77cb/build/coverlet.msbuild.targets(72,5): error : Deserialization of types without a parameterless constructor, a singular parameterized constructor, or a parameterized constructor annotated with 'JsonConstructorAttribute' is not supported. Type 'Coverlet.Core.Method'. Path: $['MyProject.Auth.dll']['/datadisks/disk1/work/1/s/src/MyProject.Auth/Utils/AuthExtensions.cs']['MyProject.Auth.Utils.AuthExtensions']['System.Threading.Tasks.Task MyProject.Auth.Utils.AuthExtensions::ClearRolesForUser(MyProject.Auth.Services.IExternalAuthService,System.String,System.Threading.CancellationToken)'] | LineNumber: 4 | BytePositionInLine: 204. [/datadisks/disk1/work/1/s/test/MyProject.Distributed.Extensions.Tests/MyProject.Distributed.Extensions.Tests.csproj]
/home/AzDevOps/.nuget/packages/coverlet.msbuild/6.0.1-preview.31.g0db8de77cb/build/coverlet.msbuild.targets(72,5): error :    at System.Text.Json.ThrowHelper.ThrowNotSupportedException(ReadStack& state, Utf8JsonReader& reader, NotSupportedException ex) [/datadisks/disk1/work/1/s/test/MyProject.Distributed.Extensions.Tests/MyProject.Distributed.Extensions.Tests.csproj]
/home/AzDevOps/.nuget/packages/coverlet.msbuild/6.0.1-preview.31.g0db8de77cb/build/coverlet.msbuild.targets(72,5): error :    at System.Text.Json.ThrowHelper.ThrowNotSupportedException_DeserializeNoConstructor(Type type, Utf8JsonReader& reader, ReadStack& state) [/datadisks/disk1/work/1/s/test/MyProject.Distributed.Extensions.Tests/MyProject.Distributed.Extensions.Tests.csproj]
[...]

PR #1588, opened 2024-01-17
update access  modifier "public Methode()"

Bertk:
related #1583


daveMueller:
A bit annoying that we need to change the access modifier but I also think this is currently the best solution. 👍

daveMueller (approved):

Issue #1587, opened 2024-01-15
[BUG] Using MSBuild, the coverage report is not being generated if the file does not exist previously


mikibakaiki:
**Describe the bug**
I was trying to use the `coverlet.msbuild` to make the command `dotnet test` always run as `dotnet test /p:CollectCoverage=true` to make life easier for my work colleagues - and myself!

However, this was not working properly. After hours of trying different things, and also running the command `dotnet test /p:CollectCoverage=true` without any success, i tried creating an empty file called `coverage.cobertura.xml` and there it was, the report was generated, and it had overwritten the empty file.

The first report with `ReportGenerator` was empty, because it read the empty file, but in the subsequent runs, it was reading the correct data - I'm currently unsure if it is reading the "freshest" data or just the n-1 run attempt.

**To Reproduce**
Here's my `test.csproj` setup: 

```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>net7.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
    <IsPackable>false</IsPackable>
    <IsTestProject>true</IsTestProject>
  </PropertyGroup>

<!-- here, i'm defining the the arguments -->
  <PropertyGroup>
    <CollectCoverage>true</CollectCoverage>
    <CoverletOutput>$(MSBuildProjectDirectory)\coverage\</CoverletOutput>
    <CoverletOutputFormat>cobertura</CoverletOutputFormat>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="coverlet.msbuild" Version="6.0.0">
      <PrivateAssets>all</PrivateAssets>
      <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
    </PackageReference>
    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.6.0" />
    <PackageReference Include="moq" Version="4.20.70" />
    <PackageReference Include="xunit" Version="2.6.5" />
    <PackageReference Include="ReportGenerator" Version="5.2.0" />
     <!-- other packages -->
  </ItemGroup>

  <ItemGroup>
    <ProjectReference Include="..\base_location\<the actual code>.csproj" />
  </ItemGroup>

  <Target Name="GenerateCoverageReport" AfterTargets="VSTest">
[...]

PR #1601, opened 2024-01-23
fix MergeWith option

Bertk:
- adds JsonSerializerOptions for JsonSerializer.Deserialize (#1513)
- removes src/coverlet.collector/Resources/Resources.resx
- updates nuget package versions for coverlet.core.csproj
- add tests in coverlet.core.tests/Coverage/CoverageTests.cs
- add message for invalid mergeWith parameter 
  `LogInformation($"MergeWith: file '{_parameters.MergeWith}' does not exist.");`

#1587 #1583


pinkfloydx33:
Is there any chance this will be reviewed soon? We'd really like to start using the nightly build again and this is the blocker

Bertk:
@daveMueller @MarcoRossignoli Please review this PR.

src/coverlet.core/coverlet.core.csproj
```diff
@@ -7,18 +7,23 @@
...
-    <PackageReference Include="Microsoft.Extensions.FileSystemGlobbing" VersionOverride="6.0.0" />
-    <PackageReference Include="Microsoft.Extensions.DependencyInjection" VersionOverride="6.0.1"/>
+    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" VersionOverride="8.0.0"/>
+    <PackageReference Include="Microsoft.Extensions.DependencyModel" VersionOverride="8.0.0"/>
+    <PackageReference Include="Microsoft.Extensions.FileSystemGlobbing" VersionOverride="8.0.0" />
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection" VersionOverride="8.0.0"/>
     <PackageReference Include="Mono.Cecil" />
     <PackageReference Include="NuGet.Versioning" />
-    <PackageReference Include="System.Text.Json" VersionOverride="6.0.9"/>
+    <PackageReference Include="System.Text.Json" VersionOverride="8.0.1"/>
```
daveMueller, line 16: What is the actual fix now? Moving this to `8.0.1`? And with the version upgrade, can we now change the access modifier of `CoverageResult.Method` ctor back to `internal`?

Directory.Build.props
```diff
@@ -53,6 +53,23 @@
...
+       .NET 8.0.3xx core-sdk-tasks https://github.com/dotnet/installer/blob/release/8.0.3xx/src/core-sdk-tasks/core-sdk-tasks.csproj
+
+       align package versions coverlet.core with coverlet.collector (vstest) and coverlet.msbuild.tasks (.NET SDK MSBuild)
+
+            <MicrosoftBuildTaskVersion>15.7.179</MicrosoftBuildTaskVersion>
+            <MicrosoftBuildTaskUtilitiesCoreVersion>15.7.179</MicrosoftBuildTaskUtilitiesCoreVersion>
+            <NuGetBuildTasksPackageVersion>6.9.0-rc.86</NuGetBuildTasksPackageVersion>
+            <MicrosoftBuildTaskSystemReflectionMetaData>1.4.2</MicrosoftBuildTaskSystemReflectionMetaData> 
+            <MicrosoftBuildTaskSystemCollectionImmutable>1.5.0</MicrosoftBuildTaskSystemCollectionImmutable>  >= 1.3.1
+    -->
```
daveMueller, line 72: Could you elaborate on this comment a bit? Did you add this to make sure that the versions in `coverlet.core` and the other projects align.

src/coverlet.collector/coverlet.collector.csproj
```diff
@@ -49,33 +49,6 @@
     </None>
   </ItemGroup>
 
-  <ItemGroup>
```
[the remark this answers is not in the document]
Bertk: I removed it because it was generating some weird entries in csproj file and it was also useless.

src/coverlet.core/coverlet.core.csproj
Bertk: The actual fix is update of the nuget versions of coverlet.core to 8.0.x. This resolves version conflicts with net7.0 and net8.0.
We should be able to use internal again but this should be checked by @pinkfloydx33.

Directory.Build.props
Bertk: Done. The list will remind me to check versions of ALL dependencies (.NET 8.0 MSBuild)

src/coverlet.core/coverlet.core.csproj
pinkfloydx33: Let me know when there's a new build and I can try it out

Bertk: @pinkfloydx33 Please use the [package](https://dev.azure.com/tonerdo/coverlet/_build/results?buildId=2833&view=artifacts&pathAsName=false&type=publishedArtifacts) from CI build.

pinkfloydx33: It doesn't throw any errors BUT it's also not merging.

It actually looks like it may be _partially_ merging that execution order of the projects might matter. 

I've re-run using the same preview package that I previously reported as working--as it turns out I was mistaken. It seems like even there that the results don't get fully merged.  This seems to be exacerbated by a CLI bug (?) that _sometimes_ reports 100% on an empty result after the merge (which is probably why I thought it was working)

After merging, the coverage data for _previously_ executed projects is just the following (spammed over and over):

```json
      "SomeNamespace.SomeClass": {
        "System.Void SomeNamespace.SomeClass::.ctor()": {
          "Lines": {},
          "Branches": []
        }
      },
```

I don't know what made me think/report that the previous updates were working successfully. They clearly aren't. Don't you guys have tests and/or local dummy projects to help validate the merging behavior? I'm unfortunately not going to be able to spend much more time validating (at least in the near term anyways)

pinkfloydx33: Just as an example, here's the CLI output from two consecutive runs:

```csharp
Passed!  - Failed:     0, Passed:   473, Skipped:     0, Total:   473, Duration: 10 s - MyProject.Things.Web.API.Tests.dll (net7.0)
Results File: D:\Source\MyProject.Things\test\MyProject.Things.Web.API.Tests\TestResults\TestResults.xml
  [coverlet]
  [coverlet] MergeWith: file 'D:/Source/MyProject.Things/test/Coverage/coverage.json' does not exist.
   Generating report 'D:\Source\MyProject.Things\test\Coverage\coverage.json'


+-----------------------------------------+--------+--------+--------+
| Module                                  | Line   | Branch | Method |
+-----------------------------------------+--------+--------+--------+
| MyProject.Things.API.Contracts          | 97.56% | 100%   | 100%   |
+-----------------------------------------+--------+--------+--------+
| MyProject.Things.Auth                   | 8.37%  | 1.85%  | 5.19%  |
+-----------------------------------------+--------+--------+--------+
| MyProject.Things.Database               | 0%     | 0%     | 0%     |
+-----------------------------------------+--------+--------+--------+
| MyProject.Things.Distributed.Extensions | 26.75% | 17.3%  | 21.15% |
+-----------------------------------------+--------+--------+--------+
| MyProject.Things.Distributed.Work       | 0%     | 0%     | 0%     |
+-----------------------------------------+--------+--------+--------+
| MyProject.Things.Service                | 18.96% | 1.5%   | 11.15% |
+-----------------------------------------+--------+--------+--------+
| MyProject.Things.Web.API                | 99%    | 95.23% | 100%   |   
+-----------------------------------------+--------+--------+--------+
```

and the second:

```csharp
Passed!  - Failed:     0, Passed:   181, Skipped:     0, Total:   181, Duration: 926 ms - MyProject.Things.Auth.Tests.dll (net7.0)
Results File: D:\Source\MyProject.Things\test\MyProject.Things.Auth.Tests\TestResults\TestResults.xml
  [coverlet]
  [coverlet] MergeWith: 'D:/Source/MyProject.Things/test/Coverage/coverage.json'.
   Generating report 'D:\Source\MyProject.Things\test\Coverage\coverage.json'


+-----------------------------------------+--------+--------+--------+
| Module                                  | Line   | Branch | Method |
+-----------------------------------------+--------+--------+--------+
| MyProject.Things.API.Contracts          | 0%     | 0%     | 0%     |
+-----------------------------------------+--------+--------+--------+
| MyProject.Things.Auth                   | 99.73% | 96.29% | 98.7%  |  
+-----------------------------------------+--------+--------+--------+
| MyProject.Things.Database               | 0%     | 0%     | 0%     |
+-----------------------------------------+--------+--------+--------+
| MyProject.Things.Distributed.Extensions | 0%     | 0%     | 0%     |
+-----------------------------------------+--------+--------+--------+
| MyProject.Things.Distributed.Work       | 0%     | 0%     | 0%     |
+-----------------------------------------+--------+--------+--------+
| MyProject.Things.Service                | 0%     | 0%     | 0%     |
+-----------------------------------------+--------+--------+--------+
| MyProject.Things.Web.API                | 100%   | 100%   | 100%   |
+-----------------------------------------+--------+--------+--------+
```

Note that the last line shows as 100% and that there's a lot of zeros on the other lines. It's coverage scores should not have **increased** as the "Auth" project doesn't reference the API project. That bottom project is an example of one showing up in the merged results with the empty branch/lines properties I mentioned above

pinkfloydx33: I will try again in the morning. Do you have an updated package or should I use the same one?

The only thing different when I ran them is that I also had the project open in visual studio. I'm wondering if that was somehow doing something to me because I swear when I tested last week I had a successful merge. (the results I output above clearly didn't merge and the result file was mostly blank)

Bertk: @pinkfloydx33 Please use the latest [package](https://dev.azure.com/tonerdo/coverlet/_build/results?buildId=2838&view=artifacts&pathAsName=false&type=publishedArtifacts).

Yes, Visual Studio sometimes blocks file access and should be closed.

pinkfloydx33: Ok so I figured it out! 

After a lot of debugging/reconfiguring options I *was* able to get a final report generated using your sample commands above. But I *think* that's because report-generator is using the opencover files and *not* the merged json output. (Though perhaps that proves it's working?) 

Even if that's not the case and the `MergeWith` is working properly _in this case_, there are still some issues that prevent me from using it the way we have been. Basically there's some breaking changes still. 

First is that the command line output still only shows the results as-if the latest project was run in isolation. The percentages aren't being reported in aggregate, which they used to do with `MergeWith`. But I can live with that one. 

The bigger problem is that files are now being overwritten (or discarded?) 

In your examples above your `CoverletOutput` path changes per project AND your `MergeWith` points back to the original/first run. However when I run it, every project uses the *same* `CoverletOutput` path, which also happens to be the same folder that contains the `MergeWith` file. 

For example, try running your projects above but setting the variables like this for EVERY project:

```
/p:CoverletOutput="C:/GitHub/coverlet/artifacts/Reports/Coverage/" 
/p:MergeWith="C:/GitHub/coverlet/artifacts/Reports/Coverage/coverage.json"
```

With this configuration for every project, I believe that the second run is overwriting the original data files and so when it merges it only has the latest project data.

This scenario has always been supported and is a MUST in CI since I need all files to go to a well-known location. I don't know which project runs first so I can't handcraft a custom MergeWith pointing back to it. And even if I could I need to run the same exact command for every single test project because that's how the CI tasks work. 

This usage of a static set of paths has always worked without things overwriting each other.

Bertk: @pinkfloydx33 Thank you for the update. I think there are 2 topics which interfere MergeWith workflow in CI

- using CoverletOutput and store the coverage result file in one folder will not work in CI with multiple parallel tasks
- not using CoverletOutput or use it with relative path will not work because the current location of coverage file is not known for multiple parallel task CI scenario

I use code coverage file with opencover or cobertura format and report-generator which accumulates code coverage files for holistic report in CI since years.

pinkfloydx33: To be clear they don't run in parallel. They run one after the other. It's just part of one single command that can't change per each.

We need one single opencover report file because that is what we are expected to upload into our CI system (which then runs report generator for us, we don't do it). MergeWith in this fashion gets us that. 

Otherwise we have to add steps into our CI to run report generator ourselves, along with commands to disable the auto generation... Which is pointless if the CI tool does it automatically.

Point being is that coverlet has worked this way for years. If it's a breaking change, then that's what it is... and we (and others) will need to adapt to it. It seems unnecessary though

Bertk: @pinkfloydx33 OK, I see. I will analyse the `JsonSerializer.Deserialize` in this code section and create some additional tests.

Bertk: @pinkfloydx33 Sorry, my bad. I added missing JsonSerializerOptions for deserialization and have now following results.

Without merge:

image

With merge:

image

Please verify this on your project. 
[Nuget packages](https://dev.azure.com/tonerdo/coverlet/_build/results?buildId=2847&view=artifacts&pathAsName=false&type=publishedArtifacts)

src/coverlet.msbuild.tasks/ReportWriter.cs
```diff
@@ -30,9 +30,10 @@ public string WriteReport()
 
       if (filename == string.Empty)
       {
-        // empty filename for instance only directory is passed to CoverletOutput c:\reportpath
-        // c:\reportpath\coverage.reportedextension
+        // empty filename for instance only directory is passed to CoverletOutput 'c:/reportpath/'
+        // note: use always '/' and not backslash => 'c:\reportpath\coverage.cobertura.xml'
         filename = $"coverage.{_coverletMultiTargetFrameworksCurrentTFM}{separatorPoint}{_reporter.Extension}";
+        _directory = Path.GetFullPath(_output);
```
daveMueller, line 36: Sorry maybe a bit blind here but I don't see how this fixes #1610? And why is the full path needed again? In https://github.com/coverlet-coverage/coverlet/blob/master/src/coverlet.msbuild.tasks/CoverageResultTask.cs#L89 the directory path is already resolved and passed into the ctor of the `ReportWriter` as `directory` parameter.

Also I've seen that we have the same issue (#1610) with `coverlet.console`. But I don't think we need to tackle this in this PR here. I'll add it to the issue.

src/coverlet.core/Coverage.cs
```diff
@@ -60,6 +60,15 @@ internal class Coverage
 
     public string Identifier { get; }
 
+    readonly JsonSerializerOptions _options = new()
+    {
+      PropertyNameCaseInsensitive = true,
+      DictionaryKeyPolicy = JsonNamingPolicy.CamelCase,
+      IncludeFields = true,
+      //NumberHandling = JsonNumberHandling.AllowReadingFromString,
```
daveMueller, line 68: Is this uncommented code still needed?

src/coverlet.msbuild.tasks/ReportWriter.cs
Bertk: `Path.GetFullPath(_output)` sanitizes the output path which is used for information message e.g. `/p:CoverletOutput="../Reports/coverlet.collector/"` will be

```
  [coverlet] MergeWith: 'C:\GitHub\coverlet\artifacts\Reports\coverlet.core\coverage.json'.
   Generating report 'C:\GitHub\coverlet\test\Reports\coverlet.collector\coverage.net8.0.json'
   Generating report 'C:\GitHub\coverlet\test\Reports\coverlet.collector\coverage.net8.0.opencover.xml'
   
   and not 
   
  [coverlet] MergeWith: 'C:\GitHub\coverlet\artifacts\Reports\coverlet.core\coverage.json'.
   Generating report '../Reports/coverlet.collector/coverage.net8.0.json'
   Generating report '../Reports/coverlet.collector/coverage.net8.0.opencover.xml'
   
```

src/coverlet.core/Coverage.cs
Bertk: Removed comment.

src/coverlet.core/coverlet.core.csproj
pinkfloydx33: Thanks will try and test this morning

src/coverlet.msbuild.tasks/ReportWriter.cs
daveMueller: OK now I see what you want to do with this. But wouldn't it make more sense to sanitize the path somewhere later and not in the `(filename == string.Empty)` branch?
The paths with file name can also be relative and are currently also displayed like this in the information message.

grafik

I would probably do this right before the result gets written to the file.
```csharp
string report = Path.Combine(Path.GetFullPath(_directory), filename);
_fileSystem.WriteAllText(report, _reporter.Report(_result, _sourceRootTranslator));
```

Bertk: That is OK. I limited it to folder case only and kept the unrealistic  `Reporters.Msbuild_ReportWriter` tests. Removed the mix-up tests for folder and file name because the **default** for `_directory` is the current working directory of the test.

src/coverlet.core/coverlet.core.csproj
pinkfloydx33: @Bertk sorry yesterday was crazy. Today is too but will do my best to get a test run in this morning if I can

daveMueller (approved):

daveMueller:
LGTM 👍

PR #1603, opened 2024-01-28
Adaptions to exclude by attribute feature

### LINKED ISSUE #1589: [BUG] Documentation states you can exclude by attribute with short name or full name - cannot by long name
**Describe the bug**
From https://github.com/coverlet-coverage/coverlet/blob/master/Documentation/GlobalTool.md

> You can also ignore additional attributes by using the ExcludeByAttribute property (short name or full name supported):

**To Reproduce**
Create a custom attribute, apply it at any level.
Supply fully qualified through one of the drivers

**Expected behavior**
Is it excluded

**Actual behavior**
Is not excluded

**Additional context**
https://github.com/coverlet-coverage/coverlet/blob/a0bd87d70f899d955891956ce4173e8bf31c16b0/src/coverlet.core/Instrumentation/Instrumenter.cs#L775

None of the Instrumenter tests supply the attributesToIgnore as fully qualified
e.g 
https://github.com/coverlet-coverage/coverlet/blob/a0bd87d70f899d955891956ce4173e8bf31c16b0/test/coverlet.core.tests/Instrumentation/InstrumenterTests.cs#L150

### LINKED ISSUE #1591: Poor documentation around attribute-based exclusion
I am trying to find somewhere in the documentation which states the default list of attributes that cause things to be excluded from coverage.

I'm guessing the answer is simply "System.Diagnostics.CodeAnalysis.ExcludeFromCodeCoverageAttribute", but I can't see that documented anywhere.

I'm using the "dotnet test" approach if that matters. I think the documentation for that is https://github.com/coverlet-coverage/coverlet/blob/master/Documentation/VSTestIntegration.md

The only mention on there of anything close to this is the example which shows
          <ExcludeByAttribute>Obsolete,GeneratedCodeAttribute,CompilerGeneratedAttribute</ExcludeByAttribute>
but there is no explanation for what that does. For example, does that setting replace the default attributes that are used to exclude things, or does it just add to it?

I have read elsewhere that these are just the "name" of the attribute, without needing to be fully-qualified... it would be nice if that was documented here as well.

And I saw a comment somewhere saying don't forget the "Attribute" on the end. If that's true, if would be useful to have that here too, and if that's the case, then is the "Obsolete" example faulty?

daveMueller:
closes #1589, closes #1591


Bertk (approved):
LGTM

MarcoRossignoli (approved):

PR #1630, opened 2024-02-23
Add regex evaluation timeout

mokarchi:
[Regex - Deny of Service (ReDoS)](https://www.meziantou.net/regex-deny-of-service-redos.htm)


MarcoRossignoli:
I don't think we need this anti-DoS timeout...the inputs are under the control of the user that's using coverlet and not from "unknown users" like in case of a web app.
But if we want to add if in case someone is using it under some other tool where users can add arbitrary input we should at least increase a bit, 1 sec looks too few to me, we could put 10.

mokarchi:
@MarcoRossignoli Changed to 10 seconds

PR #1636, opened 2024-02-28
Fix TypeInitializer issue

### LINKED ISSUE #1629: [BUG] Type initializer errors after updating from 6.0.0 to 6.0.1
**Describe the bug**
After upgrading from 6.0.0 to 6.0.1 the tests fail with errors initializing Coverlet

**To Reproduce**
Updated from 6.0.0 to 6.0.1.
Project being tested targets netstandard2.0. Only external reference is Newtonsoft.Json 13.0.3
Test project targets 4.7.2 and has the following references:

```
    <PackageReference Include="coverlet.msbuild" Version="6.0.1" PrivateAssets="all" IncludeAssets="runtime; build; native; contentfiles; analyzers; buildtransitive" />
    <PackageReference Include="GitVersion.MsBuild" Version="5.12.0" PrivateAssets="all" IncludeAssets="runtime; build; native; contentfiles; analyzers; buildtransitive" />
    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.9.0" />
    <PackageReference Include="NSubstitute" Version="5.1.0" />
    <PackageReference Include="Nunit" Version="3.14.0" />
    <PackageReference Include="NUnit3TestAdapter" Version="4.5.0" />
    <PackageReference Include="WireMock.Net" Version="1.5.40" />
```

**Expected behavior**
Tests to run without crashing

**Actual behavior**

```
   System.TypeInitializationException : The type initializer for 'Coverlet.Core.Instrumentation.Tracker.OutSystems.Auditing_cbebb53f-99cc-461e-8124-db81357fd8a1' threw an exception.
  ----> System.IO.FileNotFoundException : Could not load file or assembly 'System.Runtime, Version=6.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The system cannot find the file specified.
  Stack Trace:
     at Coverlet.Core.Instrumentation.Tracker.OutSystems.Auditing_cbebb53f-99cc-461e-8124-db81357fd8a1.RecordHit(Int32 )
   at Auditing.Tests.AuditConfigTests.AuditConfigRequestTimeoutTest() in D:\a\1\s\tests\UnitTests\Auditing.Tests\AuditConfigTests.cs:line 12
--FileNotFoundException
   at Coverlet.Core.Instrumentation.Tracker.OutSystems.Auditing_cbebb53f-99cc-461e-8124-db81357fd8a1..cctor()
```

    * Which coverlet package and version was used? 6.0.1
    * Which version of .NET is the code running on? 4.7.2
    * What OS and version, and what distro if applicable? Windows 2022 (Azure CI agent windows-latest)
    * What is the architecture (x64, x86, ARM, ARM64)? 
    * Do you know whether it is specific to that configuration? Don't know

**Additional context**
Current command being used to execute the tests:

dotnet test D:\a\1\s\tests\UnitTests\Auditing.Tests --configuration Release --no-restore --no-build --logger trx --framework net472  /p:CollectCoverage=true /p:CoverletOutputFormat=opencover /p:ExcludeByFile=**/GitVersionInformation.* /p:CoverletOutput=D:\a\TestResults\coverage.net472.1.xml

Currently I'm not able to switch execution mode to the --collect version due to the test CI templates being shared by all our or organizational repositories. So cannot test with the --collect:"XPlat Code Coverage" option

daveMueller:
closes #1629 

OK I guess I need a review here. Changing the tfm of core to only `netstandard2.0` fixes the problem. But from my perspective the `CoverletToolsPath` should also be resolved properly. The change of the condition for `CoverletToolsPath` I did in this PR still doesn't seem to work properly. No matter what I do it always point to netstandard2.0
```cmd
CoverletToolsPath = C:\Users\...\.nuget\packages\coverlet.msbuild\6.0.2-preview.7.ga760715d51\build\..\tasks\netstandard2.0\
```

But what I think, for e.g. `net6` it should point to 
```cmd
CoverletToolsPath = C:\Users\...\.nuget\packages\coverlet.msbuild\6.0.2-preview.7.ga760715d51\build\..\tasks\net6.0\
```
@MarcoRossignoli @Bertk Am I missing something? Can you help out here?

### DIFF
--- a/src/coverlet.msbuild.tasks/coverlet.msbuild.props
+++ b/src/coverlet.msbuild.tasks/coverlet.msbuild.props
@@ -19,7 +19,7 @@
     <ExcludeAssembliesWithoutSources Condition="$(ExcludeAssembliesWithoutSources) == ''"></ExcludeAssembliesWithoutSources>
   </PropertyGroup>
   <PropertyGroup>
-    <CoverletToolsPath Condition="'$(CoverletToolsPath)' == '' and '$(MSBuildRuntimeType)' == 'Core'">$(MSBuildThisFileDirectory)..\tasks\net6.0\</CoverletToolsPath>
-    <CoverletToolsPath Condition="'$(CoverletToolsPath)' == '' and '$(MSBuildRuntimeType)' != 'Core'">$(MSBuildThisFileDirectory)..\tasks\netstandard2.0\</CoverletToolsPath>
+    <CoverletToolsPath Condition="$(TargetFrameworkIdentifier) == '.NETCoreApp'">$(MSBuildThisFileDirectory)..\tasks\net6.0\</CoverletToolsPath>
+    <CoverletToolsPath Condition="$(TargetFrameworkIdentifier) != '.NETCoreApp'">$(MSBuildThisFileDirectory)..\tasks\netstandard2.0\</CoverletToolsPath>
   </PropertyGroup>
 </Project>
--- a/src/coverlet.msbuild.tasks/coverlet.msbuild.props
+++ b/src/coverlet.msbuild.tasks/coverlet.msbuild.props
@@ -19,7 +19,6 @@
     <ExcludeAssembliesWithoutSources Condition="$(ExcludeAssembliesWithoutSources) == ''"></ExcludeAssembliesWithoutSources>
   </PropertyGroup>
   <PropertyGroup>
-    <CoverletToolsPath Condition="$(TargetFrameworkIdentifier) == '.NETCoreApp'">$(MSBuildThisFileDirectory)..\tasks\net6.0\</CoverletToolsPath>
-    <CoverletToolsPath Condition="$(TargetFrameworkIdentifier) != '.NETCoreApp'">$(MSBuildThisFileDirectory)..\tasks\netstandard2.0\</CoverletToolsPath>
+    <CoverletToolsPath Condition=" '$(CoverletToolsPath)' == '' ">$(MSBuildThisFileDirectory)..\tasks\netstandard2.0\</CoverletToolsPath>
   </PropertyGroup>
 </Project>
--- a/Documentation/Changelog.md
+++ b/Documentation/Changelog.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Fixed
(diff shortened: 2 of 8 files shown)


src/coverlet.msbuild.tasks/coverlet.msbuild.props
```diff
@@ -19,7 +19,7 @@
     <ExcludeAssembliesWithoutSources Condition="$(ExcludeAssembliesWithoutSources) == ''"></ExcludeAssembliesWithoutSources>
   </PropertyGroup>
   <PropertyGroup>
-    <CoverletToolsPath Condition="'$(CoverletToolsPath)' == '' and '$(MSBuildRuntimeType)' == 'Core'">$(MSBuildThisFileDirectory)..\tasks\net6.0\</CoverletToolsPath>
-    <CoverletToolsPath Condition="'$(CoverletToolsPath)' == '' and '$(MSBuildRuntimeType)' != 'Core'">$(MSBuildThisFileDirectory)..\tasks\netstandard2.0\</CoverletToolsPath>
+    <CoverletToolsPath Condition="$(TargetFrameworkIdentifier) == '.NETCoreApp'">$(MSBuildThisFileDirectory)..\tasks\net6.0\</CoverletToolsPath>
```
MarcoRossignoli, line 22: we're building multi-tfm https://github.com/coverlet-coverage/coverlet/blob/783c482bbb1d59e02e6859fd5397ebda95774f3c/src/coverlet.msbuild.tasks/coverlet.msbuild.tasks.csproj#L5

I think we load the net6.0 or the netstandard if the msbuild node is netfx @Bertk  can you confirm you did this change in past? I don't recall if we were compiling multi-tfm or not

daveMueller: Yes the `coverlet.msbuild` package contains the assemblies for both. But it doesn't look like the `net6` assemblies are used at all. No matter if the using application is core or netfx. Thus we then could also move `coverlet.msbuild.tasks` back to only being netstandard.

MarcoRossignoli: agree let's stay compat

MarcoRossignoli:
conflict but LGTM

src/coverlet.msbuild.tasks/coverlet.msbuild.props
valadas: I am happy to test it in .net framework projects if there is a beta or something like that...

PR #1637, opened 2024-03-02
Remove usage of System.Text.Json for now

### LINKED ISSUE #1625: [BUG] 6.0.1 requires dotnetsdk 8 but doesn't say that
**Describe the bug**
Had a build tied to 6.* for a version using sdk 6 and it failed with 

```
Data collector 'XPlat code coverage' message: [coverlet]System.TypeLoadException: Could not load type 'Microsoft.Extensions.DependencyInjection.ServiceCollection' from assembly 'Microsoft.Extensions.DependencyInjection.Abstractions, Version=2.2.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'.
   at Coverlet.Collector.DataCollection.CoverletCoverageCollector.GetDefaultServiceCollection(TestPlatformEqtTrace eqtTrace, TestPlatformLogger logger, String testModule)
   at Coverlet.Collector.DataCollection.CoverletCoverageCollector.OnSessionStart(Object sender, SessionStartEventArgs sessionStartEventArgs) in /_/src/coverlet.collector/DataCollection/CoverletCoverageCollector.cs:line 135.
```

**To Reproduce**
build with dotnet6

**Expected behavior**
Expect coverlet to run and report successfully

**Actual behavior**
fails to generate report

    * Which coverlet package and version was used?
      * reproduceable with 6.0.1
    * Which version of .NET is the code running on?
      * 6.0.x (tried a few versions...all reproduced) 
    * What OS and version, and what distro if applicable?
      * window 10 and linux ubuntu (whatever azdo is using now a days)
    * What is the architecture (x64, x86, ARM, ARM64)?
      * x64
    * Do you know whether it is specific to that configuration?
      * seems not to be

**Additional context**
the changes in this PR https://github.com/coverlet-coverage/coverlet/pull/1601 seems to force having dotnet 8 installed so that thinks like System.Text.Json will load the 8.0 version from the shared framework.

doesn't appear to be any

overall, seems like mistake to break backwards compatability between 6.0.0 and 6.0.1 like this...doesn't seem to match semver...but maybe that was your intent?   understand 6 is EOL at end of year so we should be moving to 8 sdk soon...

daveMueller:
closes #1625

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -44,7 +44,6 @@
     <PackageVersion Include="System.Configuration.ConfigurationManager" Version="8.0.0" />
     <PackageVersion Include="System.Linq.Async" Version="6.0.1" />
     <PackageVersion Include="System.Reflection.Metadata" Version="8.0.0" />
-    <PackageVersion Include="System.Text.Json" Version="8.0.2" />
     <PackageVersion Include="Tmds.ExecFunction" Version="0.7.1" />
     <PackageVersion Include="xunit" Version="2.6.6" />
     <PackageVersion Include="xunit.assemblyfixture" Version="2.2.0" />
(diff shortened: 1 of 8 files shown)


MarcoRossignoli (approved):
LGTM

PR #1667, opened 2024-06-25
Extra logging in module selection

Tubeliar:
I had some trouble debugging why my assembly wasn't picked up by coverlet. I enabled verbose logging but only saw expected modules being excluded. The module I was interested in didn't appear in the output at all. I now know that the pdb file was missing, but having some log output on it would have saved me a lot of debugging time.


Tubeliar:
@dotnet-policy-service agree

Bertk:
Hi, this PR uses master branch. I will close the PR because CI does not start the checks.

PR #1671, opened 2024-07-21
fix issue-inactive action

Bertk:
Introduce separate action for issue close and avoid updates for issue with stale label.


daveMueller (approved):

daveMueller:
Thanks @Bertk

PR #1675, opened 2024-07-26
Use System.Text.Json 8.0.5

Bertk:
coverlet.collector performance on .NET8.0 is reduced. 

- **coverlet.core** has _Microsoft.Extensions.DependencyModel_ which depends on _System.Text.Json_

!!! requires .NET 8.0 SDK !!!

!!!Do not merge!!! => only for performance testing


Bertk:
@daveMueller Do you have any idea why the tests are failing?
https://github.com/coverlet-coverage/coverlet/pull/1675/commits/42c08f2930b3be24159e893ccacc8100375b8110

PR #1676, opened 2024-07-26
Solution based merging for data collector

### LINKED ISSUE #1307: Implement solution based merging for data collector
VSTest allows to post process artifacts, now we can implement solution wide report merge.

RFC https://github.com/microsoft/vstest-docs/blob/main/RFCs/0031-Test-Run-Attachments-Processing.md#how-to-register-an-idatacollectorattachmentprocessor

Contributes to https://github.com/coverlet-coverage/coverlet/issues/357

daveMueller:
closes #1307

### DIFF
--- a/Documentation/MSBuildIntegration.md
+++ b/Documentation/MSBuildIntegration.md
@@ -1,6 +1,6 @@
 # Coverlet integration with MSBuild
 
-In this mode, Coverlet doesn't require any additional setup other than including the NuGet package in the unit test project. It integrates with the `dotnet test` infrastructure built into the .NET Core CLI and when enabled, will automatically generate coverage results after tests are run.
+In this mode, Coverlet doesn't require any additional setup other than including the `coverlet.msbuild` NuGet package in the unit test project. It integrates with the `dotnet test` infrastructure built into the .NET Core CLI and when enabled, will automatically generate coverage results after tests are run.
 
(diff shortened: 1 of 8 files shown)


daveMueller:
I guess I need some additional feedback/help here. The main problem is that I still don't fully understand how it is intended to implement `IDataCollectorAttachmentProcessor`. I see it beeing executed for every test project where the attachment then only is the projects test result. And also it gets executed once for the whole solution where the attachments are all the test runs results. Thus I could only check if `attachments > 1` and then merge all of them together.
But from previous [discussions](https://github.com/microsoft/vstest/issues/5133#issuecomment-2219722040) it seems like there could be intermediate results that need to merged. 
Therfore I added now a new `runsettings` option called `MergeDirectory` to store intermediate results. The idea behind that is that everytime the attachment processor triggers all attachments are merged plus the intermediate result in the `MergeDirectory`. The result of this merge is the new intermediate result.
Unfortunately the `MergeDirectory` path must be an absolute path so that this works reliable. If it is a relative path, the location changes depending on the if the processor is triggered from a project or from a solution.
cc: @MarcoRossignoli @bert2

MarcoRossignoli:
> Therfore I added now a new runsettings option called MergeDirectory to store intermediate results. The idea behind that is that everytime the attachment processor triggers all attachments are merged plus the intermediate result in the MergeDirectory. The result of this merge is the new intermediate result.

Why you need intermediate? take the directory from the first of the 2 attachments you receive and save in the same directory removing the 2 and send back. Something like a reduce.

PR #1681, opened 2024-08-06
fix filename update for activated UseSourceFile option

### LINKED ISSUE #1679: [BUG] Different filenames with UseSourceLink after .NET 8
**Describe the bug**
Issue is connected to this and partially explained: [https://github.com/coverlet-coverage/coverlet/issues/1658](https://github.com/coverlet-coverage/coverlet/issues/1658)

From .NET 8 source links are included in the SDK: [https://learn.microsoft.com/en-us/dotnet/core/compatibility/sdk/8.0/source-link](https://learn.microsoft.com/en-us/dotnet/core/compatibility/sdk/8.0/source-link), [https://github.com/dotnet/sourcelink](https://github.com/dotnet/sourcelink). This means enabling **UseSourceLink** option results in different filenames compared to previous versions. However if someone has a referenced assembly for a test project which content is fully uncovered filenames could be inconsistent inside a single generated report files. Meaning for example a `coverage.cobertura.xml` could contain filenames like `c:\users\me\...\ProjectA\ClassA.cs` and `https://github.com/.../ProjectA/ClassA.cs` at the same time.

Further problem with this that additional tools like [ReportGenerator](https://github.com/danielpalme/ReportGenerator) calculates invalid coverage if filenames are inconsistent through multiple report files.

**To Reproduce**
Issue can be easily reproduced in a .NET 8 solution if a tesproject has a reference for a covered and non-covered assembly.

Small example can be found here: [https://github.com/tungi52/CoverletBug](https://github.com/tungi52/CoverletBug). Running the following command the report XML will be wrong:

`dotnet test CoverletBug_net8.sln --collect:"XPlat Code Coverage" --settings TestProject/test.runsettings`

Problem is here:

[https://github.com/tungi52/CoverletBug/blob/main/TestResults/with_sdk8/coverage.cobertura.xml#L9](https://github.com/tungi52/CoverletBug/blob/main/TestResults/with_sdk8/coverage.cobertura.xml#L9)

[https://github.com/tungi52/CoverletBug/blob/main/TestResults/with_sdk8/coverage.cobertura.xml#L25](https://github.com/tungi52/CoverletBug/blob/main/TestResults/with_sdk8/coverage.cobertura.xml#L25)

**Expected behavior**
Filenames should be consistent inside and across multiple report files using the same configuration.

**Actual behavior**
Filenames are inconsistent inside and across multiple a report files with the same configuration. Possible root cause of the error is here:  [https://github.com/coverlet-coverage/coverlet/blob/master/src/coverlet.core/Coverage.cs#L368](https://github.com/coverlet-coverage/coverlet/blob/master/src/coverlet.core/Coverage.cs#L368) between line 368 and 387. As I mentioned the issue is only reproducable if there is a fully uncovered referenced assembly. The linked code part is exactly shows this problem: if there is no hit for an assembly the source links are not processed and later the physical path will be used.

    * Coverlet version: v6.0.2, but reproducable with older, .NET 8 compatible versions too
    * .NET version: v8.0.303
    * OS: Windows 11; Alpine linux
    * Architecture: x64
    * I think it is not configuration specific issue

**Additional context**
There was a small discussion here: [https://github.com/coverlet-coverage/coverlet/issues/1658](https://github.com/coverlet-coverage/coverlet/issues/1658)

Bertk:
fixes #1679

### DIFF
--- a/src/coverlet.core/Coverage.cs
+++ b/src/coverlet.core/Coverage.cs
@@ -108,7 +108,7 @@ namespace Coverlet.Core
       _parameters.IncludeFilters = _parameters.IncludeFilters?.Where(f => _instrumentationHelper.IsValidFilterExpression(f)).ToArray();
 
       IReadOnlyList<string> validModules = _instrumentationHelper.SelectModules(modules, _parameters.IncludeFilters, _parameters.ExcludeFilters).ToList();
-      foreach (var excludedModule in modules.Except(validModules))
+      foreach (string excludedModule in modules.Except(validModules))
       {
         _logger.LogVerbose($"Excluded module: '{excludedModule}'");
       }
@@ -365,16 +365,6 @@ namespace Coverlet.Core
     {
       foreach (InstrumenterResult result in _results)
       {
-        if (!_fileSystem.Exists(result.HitsFilePath))
-        {
-          // Hits file could be missed mainly for two reason
-          // 1) Issue during module Unload()
-          // 2) Instrumented module is never loaded or used so we don't have any hit to register and
-          //    module tracker is never used
-          _logger.LogVerbose($"Hits file:'{result.HitsFilePath}' not found for module: '{result.Module}'");
-          continue;
-        }
-
         var documents = result.Documents.Values.ToList();
         if (_parameters.UseSourceLink && result.SourceLink != null)
         {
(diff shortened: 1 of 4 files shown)


src/coverlet.core/Coverage.cs
```diff
@@ -386,6 +376,16 @@ private void CalculateCoverage()
           }
         }
 
+        if (!_fileSystem.Exists(result.HitsFilePath))
```
MarcoRossignoli, line 379: Can you elaborate this change?

Bertk: The issue #1679 describes the problem and changing the sequence make sure the file name for activated SourceLink option is **always** updated.

MarcoRossignoli (approved):

Issue #1224, opened 2021-09-17
Coverlet.Collector.Utilities.CoverletDataCollectorException failing to access a DLL because it is being used by another process


jeofyoung:
Hi,

We're recently hitting this error when we made a change:

```
2021-09-16T02:13:33.6242181Z Data collector 'XPlat code coverage' message: [coverlet]Coverlet.Collector.Utilities.CoverletDataCollectorException: CoverletCoverageDataCollector: Failed to get coverage result
2021-09-16T02:13:33.6247821Z  ---> System.AggregateException: One or more errors occurred. (The process cannot access the file 'D:\a\_work\1\a\test\runners.test\Release\netcoreapp3.1\runners.dll' because it is being used by another process.) (The process cannot access the file 'D:\a\_work\1\a\test\runners.test\Release\netcoreapp3.1\runners.dll' because it is being used by another process.) (The process cannot access the file 'D:\a\_work\1\a\test\runners.test\Release\netcoreapp3.1\runners.dll' because it is being used by another process.) (The process cannot access the file 'D:\a\_work\1\a\test\runners.test\Release\netcoreapp3.1\runners.dll' because it is being used by another process.) (The process cannot access the file 'D:\a\_work\1\a\test\runners.test\Release\netcoreapp3.1\runners.dll' because it is being used by another process.) (The process cannot access the file 'D:\a\_work\1\a\test\runners.test\Release\netcoreapp3.1\runners.dll' because it is being used by another process.) (The process cannot access the file 'D:\a\_work\1\a\test\runners.test\Release\netcoreapp3.1\runners.dll' because it is being used by another process.) (The process cannot access the file 'D:\a\_work\1\a\test\runners.test\Release\netcoreapp3.1\runners.dll' because it is being used by another process.) (The process cannot access the file 'D:\a\_work\1\a\test\runners.test\Release\netcoreapp3.1\runners.dll' because it is being used by another process.) (The process cannot access the file 'D:\a\_work\1\a\test\runners.test\Release\netcoreapp3.1\runners.dll' because it is being used by another process.) (The process cannot access the file 'D:\a\_work\1\a\test\runners.test\Release\netcoreapp3.1\runners.dll' because it is being used by another process.) (The process cannot access the file 'D:\a\_work\1\a\test\runners.test\Release\netcoreapp3.1\runners.dll' because it is being used by another process.)
[...]

PR #1700, opened 2024-09-14
workaround for "Failed to instrument modules"

Bertk:
#1224

### DIFF
--- a/Documentation/KnownIssues.md
+++ b/Documentation/KnownIssues.md
@@ -249,3 +249,26 @@ Change [DebugType](https://learn.microsoft.com/en-us/dotnet/csharp/language-refe
 
 >[!IMPORTANT]
 >NET Core introduces a new symbol file (PDB) format - portable PDBs. Unlike traditional PDBs which are Windows-only, portable PDBs can be created and read on all platforms.
+
+## CoverletCoverageDataCollector: Failed to instrument modules
+
+*Affected drivers*: VSTest integration `dotnet test --collect:"XPlat Code Coverage"`
+
+*Symptoms:*
+
+```text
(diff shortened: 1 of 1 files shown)


Documentation/KnownIssues.md
```diff
@@ -249,3 +249,26 @@ Change [DebugType](https://learn.microsoft.com/en-us/dotnet/csharp/language-refe
...
+   at Coverlet.Core.Helpers.InstrumentationHelper.<>c__DisplayClass16_0.<RestoreOriginalModule>b__1() in /_/src/coverlet.core/Helpers/InstrumentationHelper.cs:line 277
+   at Coverlet.Core.Helpers.RetryHelper.<>c__DisplayClass0_0.<Retry>b__0() in /_/src/coverlet.core/Helpers/RetryHelper.cs:line 28
+   at Coverlet.Core.Helpers.RetryHelper.Do[T](Func`1 action, Func`1 backoffStrategy, Int32 maxAttemptCount) in /_/src/coverlet.core/Helpers/RetryHelper.cs:line 55
+   --- End of inner exception stack trace ---
+   ...
+   ```
+
+   *Solutions:*
+
+   use VSTest setting [-maxcpucount:1](https://learn.microsoft.com/en-us/visualstudio/msbuild/building-multiple-projects-in-parallel-with-msbuild?view=vs-2022#-maxcpucount-switch) which limits "worker processes".
```
MarcoRossignoli, line 274: I would also explain that it's unrelated to coverlet but related to the parallel nature of the build and the "specific" setup of the build that sometimes keep files loaded/locked.

This is not a coverlet issue.

Bertk: I updated the topic but maybe you have a better proposal.

PR #1709, opened 2024-11-09
Relocate tests using Tmds.ExecFunction to new test project coverlet.core.coverage.tests 

Bertk:
This pull request includes the following changes and enhancements:

**Testing:**
- unit tests using Tmds.ExecFunction relocated to coverlet.core.coverage.tests. 
- coverlet.core.tests execution will now run in VS 2022 without failing tests.

**Build pipeline:**
- test runs names use now target OS and configuration instead of generic name like 'VSTest_TestResults_3262_3'
- test results and code coverage are now also available after a test has failed


MarcoRossignoli (approved):

src/coverlet.console/coverlet.console.csproj
```diff
@@ -24,6 +24,7 @@
   </PropertyGroup>
  
   <ItemGroup>
+    <PackageReference Include="NuGet.Versioning" />
```
daveMueller, line 27: Why this needed here? Don't we already use `Nerdbank.GitVersioning` as a global refernce in all packable projects? The only real reference to `NuGet.Versioning` seems to be in `coverlet.core`. But maybe I'm missing something.

src/coverlet.msbuild.tasks/coverlet.msbuild.tasks.csproj
```diff
@@ -41,6 +41,10 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Build.Utilities.Core" VersionOverride="$(MicrosoftBuildUtilitiesCorePackageVersion)" PrivateAssets="all" ExcludeAssets="Runtime" />
+    <PackageReference Include="NuGet.Versioning" />
```
daveMueller, line 44: Basically the same like in the comment above.

daveMueller (approved):

daveMueller (commented):
I have one last question. Are you sure that `test/coverlet.core.coverage.tests/out.txt`, `test/coverlet.core.coverage.tests/public.snk` and `test/coverlet.core.coverage.tests/key.publickey` aren't needed anymore? Could you figure out for what they were used in the past?

Bertk:
> I have one last question. Are you sure that `test/coverlet.core.coverage.tests/out.txt`, `test/coverlet.core.coverage.tests/public.snk` and `test/coverlet.core.coverage.tests/key.publickey` aren't needed anymore? Could you figure out for what they were used in the past?

This are some intermediat files which are not necessary. I use _out.txt_ for test scripts and _public.snk, key.publickey_ for [new project using strong naming.](https://learn.microsoft.com/en-us/dotnet/standard/assembly/create-public-private-key-pair)

PR #1740, opened 2025-03-28
Use latest net8 SDK

Evangelink:


MarcoRossignoli (approved):
LGTM if tests are green

PR #1765, opened 2025-08-28
doc: update exclude from code coverage example (fixes #1764)

### LINKED ISSUE #1764: [DOC] Excluding From Coverage bad defaults from given example
**Describe the bug**
From [the documentation](https://github.com/coverlet-coverage/coverlet/blob/master/Documentation/MSBuildIntegration.md#excluding-from-coverage):

Image

The problem is that at least all async code is excluded from code coverage as it is stated in the [dotnet documentation](https://learn.microsoft.com/en-us/visualstudio/test/customizing-code-coverage-analysis?view=vs-2022#include-or-exclude-assemblies-and-members):

Image

**Expected behavior**
The given example should not add unwanted behavior. Removing `CompilerGeneratedAttribute` from the example could be a good start

**Actual behavior**
By using the given example, the code coverage is higher than expected.

cboudereau:
Fixes #1764


cboudereau:
@dotnet-policy-service agree

Bertk (approved):

Issue #1633, opened 2024-02-27
[BUG] Records show no coverage (without primary constructor)


tisonv:
**Describe the bug**
With .Net 8, records show no coverage
Tested with the nightly build too
I read #1561 #1607 #1576 

**To Reproduce**
```
public sealed record FolderInfoDto
{
    public bool IsHidden { get; init; }
    public long Creation { get; init; }
}
return new FolderInfoDto
{
    Creation = f.Creation,
    IsHidden = !string.IsNullOrEmpty(f.Mode) && 'h' == f.Mode[3]
};
```
```
internal abstract record BasicFolderDto
{
    public string? DfsMappedPath { get; init; }
    public string Path { get; init; } = null!;
}

internal sealed record CreateFolderDto : BasicFolderDto;

CreateFolderDto dto = new()
{
    Path = args != null && args.TryGetValue("path", out var path) ? (string)path : @"F:\rep1\rep2",
};
```
```
internal sealed record DeleteFolderDto : BasicFolderDto
{
    public bool Recurse { get; init; }
    public bool Force { get; init; }
}

DeleteFolderDto dto = new()
{
    Path = @"L:\path\",
};
```

**Expected behavior**
`FolderInfoDto`, `CreateFolderDto`, `DeleteFolderDto`, `BasicFolderDto` are marked as covered.

**Actual behavior**
They are marked as not covered.

**Configuration (please complete the following information):**
Please provide more information on your .NET configuration:
    * Which coverlet package and version was used? **6.0.1 and nightly** 
    * Which version of .NET is the code running on? **8.0.2**
    * What OS and version, and what distro if applicable? **Windows 11**
    * What is the architecture (x64, x86, ARM, ARM64)? **x64**
    * Do you know whether it is specific to that configuration? **No**

**Additional context**
 * Same result from Visual Studio and CLI
 * No difference with either `<TargetFramework>net8.0</TargetFramework>` or `<TargetFramework>net8.0-windows10.0.17763.0</TargetFramework>`
 * removing `abstract` or `sealed` makes no difference
 * SkipAutoProps makes no difference

test.runsettings:
```
<?xml version="1.0" encoding="utf-8"?>
<RunSettings>
	<RunConfiguration>
		<ResultsDirectory>.\TestResults</ResultsDirectory>
		<TargetPlatform>x64</TargetPlatform>
		<TargetFramework>net8.0</TargetFramework>
	</RunConfiguration>

	<DataCollectionRunSettings>
[...]

PR #1941, opened 2026-05-21
Relax auto-property skip logic and improve coverage for records

Bertk:
This pull request addresses an issue with code coverage for C# records that do not have a primary constructor, and updates the logic for skipping auto-property assignments. The main focus is to ensure that records without primary constructors are correctly instrumented and covered, and to simplify the logic for skipping inline assigned auto-properties.

**Test coverage improvements for records:**

* Added new test cases in `Instrumentation.AutoProps.cs` to cover records without primary constructors and abstract records, ensuring these scenarios are now properly handled and covered by instrumentation.

**Logic simplification:**

* Simplified the `SkipInlineAssignedAutoProperty` method in `CecilSymbolHelper.cs` to only skip generated backing field assignments when `skipAutoProps` is enabled, removing unnecessary checks and making the logic clearer.

This PR is similar to #1761 but without refactoring.

Resolves issue #1633


Bertk:
Please note that the experimental feature addressing issue #1633 is currently opt-in and must be enabled by setting the environment variable COVERLET_EXPERIMENTAL_AUTOPROP_BACKING_FIELD_CACHE=1.

PR #1987, opened 2026-07-17
Fix EndOfStreamException

### LINKED ISSUE #1981: [BUG] Unable to read beyond the end of the stream
**Describe the bug**
Running tests randomly fails with: Unable to read beyond the end of the stream. We're using xUnit v2 with the latest coverlet.collector NuGet package. Our projects multi-target .NET 8/9/10.

**To Reproduce**
- Install latest stable .NET 10 SDK (currently 10.0.301)
- Clone https://github.com/json-api-dotnet/JsonApiDotNetCore
- Checkout https://github.com/json-api-dotnet/JsonApiDotNetCore/pull/2009
- Run `./run-docker-postgres.ps1` (requires a container runtime such as Docker Desktop)
- Run the following commands in pwsh (I'm running on Windows):
  - `dotnet tool restore`
  - `dotnet build -c Release`
  - `1..25 | % { dotnet test -v quiet --no-build --configuration Release --collect:"XPlat Code Coverage" }`

**Expected behavior**
25 successful test runs without any errors or warnings.

**Actual behavior**
3 runs result in a warning, with varying TFMs. Output is similar to:

```
DapperTests test net9.0 succeeded with 1 warning(s) (48,4s)
    C:\Program Files\dotnet\sdk\10.0.301\Microsoft.TestPlatform.targets(48,5): warning
      Data collector 'XPlat code coverage' message: [coverlet]Coverlet.Collector.Utilities.CoverletDataCollectorException: CoverletCoverageDataCollector:
      Failed to get coverage result
       ---> System.IO.EndOfStreamException: Unable to read beyond the end of the stream.
         at System.IO.BinaryReader.ReadInt32()
         at Coverlet.Core.Coverage.CalculateCoverage() in /_/src/coverlet.core/Coverage.cs:line 432
         at Coverlet.Core.Coverage.GetCoverageResult() in /_/src/coverlet.core/Coverage.cs:line 169
         at Coverlet.Collector.DataCollection.CoverageWrapper.GetCoverageResult(Coverage coverage) in /_/src/legacy/coverlet.collector/DataCollection/Cove
      rageWrapper.cs:line 62
         at Coverlet.Collector.DataCollection.CoverageManager.GetCoverageResult() in /_/src/legacy/coverlet.collector/DataCollection/CoverageManager.cs:li
      ne 93
         --- End of inner exception stack trace ---
         at Coverlet.Collector.DataCollection.CoverageManager.GetCoverageResult() in /_/src/legacy/coverlet.collector/DataCollection/CoverageManager.cs:li
      ne 97
         at Coverlet.Collector.DataCollection.CoverageManager.GetCoverageReports() in /_/src/legacy/coverlet.collector/DataCollection/CoverageManager.cs:l
      ine 81
         at Coverlet.Collector.DataCollection.CoverletCoverageCollector.OnSessionEnd(Object sender, SessionEndEventArgs e) in /_/src/legacy/coverlet.colle
      ctor/DataCollection/CoverletCoverageCollector.cs:line 164.
```

    * Which coverlet package (collector, console, msbuild, ...) and version was used? coverlet.collector NuGet v10.0.1
    * Which version of .NET is the code running on? .NET SDK 10, runtime 8/9/10
    * What OS and version, and what distro if applicable? Windows 11
    * Do you know whether it is specific to that configuration? No

**Additional context**
[Known Issues](https://github.com/coverlet-coverage/coverlet/blob/master/Documentation/KnownIssues.md) lists this error with the following text:

> Solution:
> 
> The only way to get around this issue is to use collectors integration https://github.com/coverlet-coverage/coverlet#vstest-integration-preferred-due-to-known-issue-supports-only-net-core-application. With the collector, we're injected in test host through a in-proc collector that communicates with the VSTest platform so we can signal when we end our work.

However, that link is broken. It's unclear to me how to fix this.

### LINKED ISSUE #1983: [BUG] Race condition between ProcessExit hit-file write and out-of-proc coverage read causes EndOfStreamException
# Race condition between ProcessExit hit-file write and out-of-proc coverage read causes EndOfStreamException

## Summary

`System.IO.EndOfStreamException: Unable to read beyond the end of the stream` is thrown intermittently inside `Coverlet.Core.Coverage.CalculateCoverage()` when running tests under `dotnet test --collect:"XPlat Code Coverage"`. The failure is non-deterministic — it can go hundreds of runs without appearing, then suddenly reproduce.

Tested with:

| Package | Version |
|---|---|
| `coverlet.collector` | **10.0.1** |
| `Microsoft.NET.Test.Sdk` / `Microsoft.TestPlatform` | **18.7.0** |
| `xunit` | **2.9.3** |
| `xunit.runner.visualstudio` | **3.1.5** |
| Runtime | .NET 8 / 9 / 10 (multi-targeted) |

This is a detailed root cause analysis and proposed fix for the failure reported in #1981.

> **A note on history:** The race condition has always existed in the code — `CoverletInProcDataCollector.cs`, `ModuleTrackerTemplate.cs`, and the entire data-collection hot path are byte-for-byte identical between v8.0.1 and v10.0.1. That said, this is still a regression from a consumer perspective: something that worked reliably before now fails, and the failure surfaces through coverlet. Several coverlet changes between the two releases plausibly widen the race window (detailed in a dedicated section below), and timing shifts in the wider ecosystem — updated `xunit.runner.visualstudio` 3.x shutdown behaviour, updated vstest orchestration — may also contribute. We believe the fix clearly belongs here, and hope the analysis is useful regardless of how the history is framed.

---

## Stack trace

> **Directly observed** — reproduced once in a loop of 50 runs, with `--diag` enabled. The full diagnostic output was captured and is summarised in the next section. Subsequent runs with and without diagnostic flags were unable to reproduce the failure, consistent with the race window being sensitive to any additional I/O on the hot path.

```
[coverlet] CoverletCoverageDataCollector: Failed to get coverage result
 ---> System.IO.EndOfStreamException: Unable to read beyond the end of the stream.
   at System.IO.BinaryReader.ReadInt32()
   at Coverlet.Core.Coverage.CalculateCoverage()
   at Coverlet.Collector.DataCollection.CoverageWrapper.GetCoverageResult(Coverage coverage)
   at Coverlet.Collector.DataCollection.CoverageManager.GetCoverageResult()
   --- End of inner exception stack trace ---
   at Coverlet.Collector.DataCollection.CoverageManager.GetCoverageResult()
   at Coverlet.Collector.DataCollection.CoverletCoverageDataCollector.OnSessionEnd(Object sender, SessionEndEventArgs e)
```

The out-of-proc collector is trying to read the hits file and finds it empty or truncated.

---

## Captured diagnostic timeline

> **Directly observed** — the following timestamps come from `--diag` output files (`diag_run_1.datacollector.*.txt` and `diag_run_1.host.*.txt`) captured during the single reproduction. The affected test session was `MultiDbContextTests.dll` targeting `.NET 10`. All timestamps are UTC.

### Out-of-proc data collector process (PID 2352)

Extracted from `diag_run_1.datacollector.26-07-11_13-08-21_77445_5.txt`:

```
13:08:22.269  [coverlet] CoverletCoverageDataCollector: SessionStart received
13:08:37.100  SessionStart callback completed — took 14,832 ms (instrumenting assemblies)
13:08:45.838  [coverlet] CoverletCoverageDataCollector: SessionEnd received
13:08:45.980  [coverlet] Warning: CoverletCoverageDataCollector: Failed to get coverage result
              ---> System.IO.EndOfStreamException: Unable to read beyond the end of the stream.
                 at Coverlet.Collector.DataCollection.CoverletCoverageCollector.OnSessionEnd(...)
13:08:45.981  SessionEnd callback completed — took 143 ms
```

### Test host process (PID 12632)

Extracted from `diag_run_1.host.26-07-11_13-08-37_11933_5.txt`:

```
13:08:38.076  TestExecution.Initialize received (MultiDbContextTests, net10.0)
13:08:38.292  RaiseSessionStart — InProcDataCollectionExtensionManager — took 1 ms
13:08:38.743  xUnit.net Starting: MultiDbContextTests
13:08:43.462  xUnit.net Finished: MultiDbContextTests
13:08:43.472  RaiseSessionEnd callback 1/2 — ProxyOutOfProcDataCollectionManager — took 1 ms
              (fire-and-forget SessionEnd message sent to out-of-proc; see note below)
13:08:43.993  [coverlet] Calling UnloadModule for 'MultiDbContextExample'
13:08:44.981  [coverlet] Called  UnloadModule for 'MultiDbContextExample'  (988 ms)
13:08:44.981  [coverlet] Calling UnloadModule for 'JsonApiDotNetCore.Annotations'
13:08:45.212  [coverlet] Called  UnloadModule for 'JsonApiDotNetCore.Annotations'  (231 ms)
13:08:45.213  [coverlet] Calling UnloadModule for 'JsonApiDotNetCore'
13:08:45.493  [coverlet] Called  UnloadModule for 'JsonApiDotNetCore'  (280 ms)
13:08:45.699  RaiseSessionEnd callback 2/2 — InProcDataCollectionExtensionManager — took 2,226 ms
13:08:45.722  TestSession.Terminate received — test host begins shutdown
```

### Key observations from the trace

**Fire-and-forget delay (2,366 ms):** The session-end message was sent by the test host at 13:08:43.472 but only received by the out-of-proc collector at 13:08:45.838 — a 2,366 ms gap. During this time the out-of-proc was apparently busy completing unrelated work.

**Three assemblies found by in-proc scan:** The in-proc collector's `GetAssemblies()` scan found exactly three instrumented assemblies (`MultiDbContextExample`, `JsonApiDotNetCore.Annotations`, `JsonApiDotNetCore`). All three were fully written by 13:08:45.493, which is 345 ms **before** the out-of-proc even received `SessionEnd`. Those three files were therefore complete when the out-of-proc began reading.

**Process exit overlap (inferred):** The test host received `TestSession.Terminate` at 13:08:45.722. Process shutdown and `ProcessExit` event handlers fire in the window between 13:08:45.722 and the `EndOfStreamException` at 13:08:45.980 — a gap of only 258 ms. If one or more instrumented assemblies were **not** found by the in-proc scan (leaving their `FlushHitFile` flag at `true`), their `ProcessExit` handler would attempt to write the hit file concurrently with the out-of-proc reading it — the exact scenario that produces an `EndOfStreamException`.

---

## Root cause analysis

The race involves three components that run concurrently across two OS processes:

1. **In-proc collector** (`CoverletInProcDataCollector.TestSessionEnd`) — runs inside the testhost process, scans all loaded assemblies and calls `UnloadModule` to write hit files.
2. **ProcessExit handler** (`ModuleTrackerTemplate.UnloadModule` triggered by `AppDomain.CurrentDomain.ProcessExit`) — also runs inside the testhost process when it shuts down.
3. **Out-of-proc collector** (`CoverletCoverageDataCollector.OnSessionEnd`) — runs in a separate data-collector process, reads the hit files to calculate coverage.

### Step 1 — vstest fires `SessionEnd` in the wrong order

> **Confirmed by source code and directly observed in the diagnostic trace.**

`ProxyOutOfProcDataCollectionManager` is constructed first and therefore subscribes to the `SessionEnd` event before `InProcDataCollectionExtensionManager`. C# multicast delegates fire in subscription order, so when `SendSessionEnd()` raises the event:

1. `ProxyOutOfProcDataCollectionManager.TriggerTestSessionEnd` fires **first** — sends a fire-and-forget `SessionEnd` message to the data-collector process over the per-test-case socket.
2. `InProcDataCollectionExtensionManager.TriggerTestSessionEnd` fires **second** — calls coverlet's `CoverletInProcDataCollector.TestSessionEnd` which writes the hit files.

This ordering is directly visible in the diagnostic trace: callback 1/2 (ProxyOutOfProcDataCollectionManager) completed in 1 ms at 13:08:43.472, and callback 2/2 (InProcDataCollectionExtensionManager) ran for 2,226 ms immediately after, calling `UnloadModule` three times.

The fire-and-forget is confirmed in `DataCollectionTestCaseEventSender.SendTestSessionEnd()` at **vstest v18.7.0** (`src/Microsoft.TestPlatform.CommunicationUtilities/DataCollectionTestCaseEventSender.cs`):

```csharp
public void SendTestSessionEnd(SessionEndEventArgs e)
{
    _communicationManager.SendMessage(MessageType.SessionEnd, e);
    // No ReceiveMessage() — fire-and-forget, no synchronisation with the data-collector side
}
```

(Compare with `SendTestCaseEnd` which waits for a `DataCollectionTestEndResult` reply.)

The data-collector process receives this message and closes the per-test-case channel; it does not call `OnSessionEnd` at this point. `OnSessionEnd` fires later, in response to `AfterTestRunEnd` from vstest.console. So the subscription ordering alone does not cause the race; what follows does.

### Step 2 — In-proc collector silently misses an assembly

> **Confirmed by source code** that the silent-swallow code paths exist. **Strongly supported by the diagnostic trace:** the in-proc collector logged exactly three `UnloadModule` calls, yet the out-of-proc still encountered an unreadable file — meaning at least one additional instrumented assembly was not processed by the in-proc collector. The exact reason it was missed (exception during `GetTypes()`, snapshot race, or other) is not directly visible in the logs; capturing that detail would require adding instrumentation that itself shifts the race timing.

`CoverletInProcDataCollector.GetInstrumentationClass()` (v10.0.1) swallows every exception when `COVERLET_DATACOLLECTOR_INPROC_EXCEPTIONLOG_ENABLED` is not set (the default):

```csharp
catch (Exception ex)
{
    if (_enableExceptionLog)
    {
        // only logged when env var is explicitly set
    }
    return null;   // ← silently returns null; assembly is skipped
}
```

`assembly.GetTypes()` can throw `ReflectionTypeLoadException` when the assembly has type references that cannot be resolved in the testhost's load context. When this happens for an instrumented assembly, its tracker is silently skipped and `FlushHitFile` is never cleared — making it the most likely trigger for the ProcessExit path.

Likewise, the outer `catch` around `unloadModule.Invoke()` also swallows exceptions silently when the flag is off, leaving `FlushHitFile` at `true`:

```csharp
try
{
    unloadModule.Invoke(...);
    injectedInstrumentationClass.GetField("FlushHitFile", ...).SetValue(null, false);
}
catch (Exception ex)
{
    if (_enableExceptionLog) { ... throw; }
    // FlushHitFile is NEVER set to false if Invoke() throws
}
```

### Step 3 — ProcessExit writes the hit file with `FileShare.Read`

> **Confirmed by source code and .NET documentation** — the two-argument `FileStream` constructor default is documented and visible in the .NET runtime source.

Because `FlushHitFile` was never cleared for the missed assembly, the `ProcessExit` handler calls `UnloadModule` when the testhost shuts down. The initial file write uses:

```csharp
using var fs = new FileStream(HitsFilePath, FileMode.CreateNew);
```

The two-argument `FileStream` constructor defaults to **`FileShare.Read`**, meaning any other process may open the same file for reading while it is being written.

### Step 4 — The race window

> **Directly supported by the diagnostic trace.** All three in-proc `UnloadModule` calls completed by 13:08:45.493. The out-of-proc only received `SessionEnd` at 13:08:45.838 (345 ms later) and threw `EndOfStreamException` at 13:08:45.980. The error cannot have been caused by those three files — they were fully written well before the out-of-proc started reading. The only consistent explanation is that a fourth assembly's hit file was being written by `ProcessExit` concurrently with the out-of-proc read.

```
Testhost (shutting down)               Data-collector process
───────────────────────────────────────────────────────────────
vstest.console sends kill →
ProcessExit fires for missed assembly
  → FileStream(path, CreateNew)         AfterTestRunEnd received
  → bw.Write(hitsArray.Length) ...      → OnSessionEnd fires
                                        → Coverage.CalculateCoverage()
                                        → new FileStream(path, FileMode.Open)
                                          (allowed — FileShare.Read!)
                                        → BinaryReader.ReadInt32()
                                          EndOfStreamException  ← here
  → (write continues / process dies)
```

The data-collector opens the file while it is still being created and written (0 bytes or partially written), and `ReadInt32()` fails immediately.

---

## Why it is intermittent

All three of the following conditions must coincide:

1. The in-proc collector silently misses at least one instrumented assembly.
2. The testhost is killed while the data-collector is still inside `OnSessionEnd`.
3. The ProcessExit write and the data-collector read overlap within milliseconds.

Adding diagnostic I/O to the hot path shifts timings enough to make the race disappear. Specifically, enabling `COVERLET_ENABLETRACKERLOG=1` (which adds synchronous file writes inside `UnloadModule`) or `COVERLET_DATACOLLECTOR_INPROC_EXCEPTIONLOG_ENABLED=1` consistently prevented further reproduction in subsequent runs. VSTest's own `--diag` flag does not add overhead to the coverlet hot path and does not appear to suppress the race — the single reproduction was captured with `--diag` active.

---

## What changed between v8.0.1 and v10.0.1

> **Note:** The timing effects listed here are speculative — we identified the changes by diffing every commit between the two tags, but we cannot directly observe how each one affects the race window.

A full diff confirms that the data-collector hot path (`CoverletInProcDataCollector.cs`, `CoverletCoverageCollector.cs`, `CoverageManager.cs`, `CoverageWrapper.cs`, `ModuleTrackerTemplate.cs`) is **byte-for-byte identical** between the two releases — every file simply moved from `src/coverlet.collector/` to `src/legacy/coverlet.collector/`. The race mechanism itself was never touched.

The following changes in `coverlet.core` and the package structure plausibly widen the race window:

| Change | PR | Plausible effect on race |
|---|---|---|
| Package now ships native net8.0 + net9.0 + net10.0 builds; when running net10.0 tests VSTest picks up `build/net10.0/` directly instead of loading net8.0 under a compat layer | #1823 | Different JIT, startup time, and message-pump speed on the native net10.0 binary shift event-arrival probabilities |
| Compiler-generated state machine types (`IAsyncStateMachine`, `IEnumerator<T>`, `IAsyncEnumerator<T>`) are now instrumented; previously all `[CompilerGenerated]` types were excluded | #1864 (issue #1843) | More hit counters per assembly → larger hit files → longer `UnloadModule` writes (988 ms observed for a single assembly) and longer `OnSessionStart` instrumentation (14,832 ms observed) |
| Branch coverage uses trampolines appended at end of method bodies for taken-branch paths | #1865 | More counter slots per method → larger hit files → slower writes |
| `CecilAssemblyResolver` now enumerates all `*.runtimeconfig.json` files in the module directory instead of only the module's own | #1889 | More file I/O per assembly during `OnSessionStart` instrumentation, contributing to the longer session-setup time |
| `[CompilerGenerated]` attribute from `ModuleTrackerTemplate` is now copied to every injected tracker type | #1864 | Forward-looking risk: any future attribute added to `ModuleTrackerTemplate` that references a non-BCL type would propagate into every instrumented assembly and could trigger `ReflectionTypeLoadException` in the in-proc `GetTypes()` scan |

In v8.0.1 the race window was narrow enough to never fire in practice. The changes above, in particular the dramatically larger hit files and slower `OnSessionStart`, plausibly open the window to several hundred milliseconds — wide enough to fire.

---

### Fix 1 — Replace `GetAssemblies()` scanning with a static hit-file registry (root cause)

This is the most important fix. The in-proc collector's `TestSessionEnd` currently discovers which assemblies need flushing by calling `AppDomain.CurrentDomain.GetAssemblies()` and searching each one for the injected tracker class via reflection. If `GetTypes()` throws `ReflectionTypeLoadException` for any assembly (silently swallowed — see Fix 4), or if the snapshot races with a concurrent load, that assembly's hits are never flushed by the in-proc collector and `FlushHitFile` stays `true`.

A simpler and more reliable approach is for `RegisterUnloadEvents()` to record the `HitsFilePath` of the current module in a **shared static collection** at the moment it is called — long before `TestSessionEnd` runs:

```csharp
// Example: add to ModuleTrackerTemplate
private static readonly ConcurrentBag<string> s_registeredPaths = new();

public static void RegisterUnloadEvents()
{
    s_registeredPaths.Add(HitsFilePath);   // register at load time, not scan time
    AppDomain.CurrentDomain.ProcessExit   += UnloadModule;
    AppDomain.CurrentDomain.DomainUnload  += UnloadModule;
}
```

The in-proc collector then iterates `s_registeredPaths` directly instead of calling `GetAssemblies()`, eliminating the reflection scan and its failure modes entirely.

### Fix 2 — Use an atomic rename to make partial hit files impossible (defence-in-depth)

Even if Fix 1 is implemented, writing via `FileMode.CreateNew` leaves a window during which the file exists but is incomplete. Writing to a `.tmp` path and then renaming eliminates that window entirely:

```csharp
// Before — file is visible (and readable via FileShare.Read default) while being written
using var fs = new FileStream(HitsFilePath, FileMode.CreateNew);
...

// After — file only appears at HitsFilePath once the write is complete
string tmpPath = HitsFilePath + ".tmp";
using (var fs = new FileStream(tmpPath, FileMode.CreateNew, FileAccess.Write, FileShare.None))
using (var bw = new BinaryWriter(fs))
{
    bw.Write(hitsArray.Length);
    foreach (int hitCount in hitsArray)
        bw.Write(hitCount);
}
File.Move(tmpPath, HitsFilePath);   // atomic on the same volume on both Windows and Linux
```

`File.Move` on the same volume is an OS-level rename — it is atomic. The out-of-proc collector sees either no file (write not yet complete, `FileNotFoundException` rather than `EndOfStreamException`) or a complete file. The out-of-proc collector already has error handling for this case and can report a clear warning instead of a hard failure.

### Fix 3 — Move `FlushHitFile = false` inside the mutex scope (correctness)

> **Confirmed by source code** — the TOCTOU window is directly visible in `CoverletInProcDataCollector.TestSessionEnd()` at v10.0.1.

Currently `FlushHitFile` is cleared by the in-proc collector *outside* the named mutex, after `UnloadModule` has already released it. A concurrent `ProcessExit` handler can acquire the mutex in that gap, observe `FlushHitFile == true`, and write the file a second time:

```csharp
// CoverletInProcDataCollector.TestSessionEnd — current code
unloadModule.Invoke(null, new[] { (object)this, EventArgs.Empty });
// ← ProcessExit can sneak in here and see FlushHitFile == true
injectedInstrumentationClass.GetField("FlushHitFile", ...).SetValue(null, false);
```

The cleanest fix is to clear `FlushHitFile` inside `UnloadModule` *before* calling `mutex.ReleaseMutex()`, so the flag is always consistent with the mutex state.

### Fix 4 — Surface in-proc failures without requiring an opt-in environment variable

Silent failure makes the bug invisible. Both failure paths are currently gated behind the `COVERLET_DATACOLLECTOR_INPROC_EXCEPTIONLOG_ENABLED` environment variable. `ReflectionTypeLoadException` from `assembly.GetTypes()` and any exception thrown by `unloadModule.Invoke()` should be logged at warning level unconditionally. An instrumented assembly that the in-proc collector cannot flush is a data-loss event and should never be silently ignored.

---

## Recommended priority

1. **Fix 1** (static registry) — eliminates the trigger; the other fixes become largely irrelevant once this is in place.
2. **Fix 2** (atomic rename) — defence-in-depth against any remaining or future timing issues, and protects the ProcessExit path even without Fix 1.
3. **Fix 3** (mutex scope) — correctness fix; avoids the double-write TOCTOU.
4. **Fix 4** (observability) — not a functional fix, but essential for diagnosing any recurrence.

---

## References

- coverlet issue tracking the original report: https://github.com/coverlet-coverage/coverlet/issues/1981
- Earlier race documented (different manifestation): https://github.com/tonerdo/coverlet/issues/572
- File locked variant: https://github.com/coverlet-coverage/coverlet/issues/736
- vstest `ExecutionManager.InitializeDataCollectors` (v18.7.0): https://github.com/microsoft/vstest/blob/v18.7.0/src/Microsoft.TestPlatform.CrossPlatEngine/Execution/ExecutionManager.cs
- vstest `DataCollectionTestCaseEventSender.SendTestSessionEnd` (v18.7.0): https://github.com/microsoft/vstest/blob/v18.7.0/src/Microsoft.TestPlatform.CommunicationUtilities/DataCollectionTestCaseEventSender.cs
- `FileStream(string, FileMode)` default share mode: https://learn.microsoft.com/en-us/dotnet/api/system.io.filestream.-ctor#system-io-filestream-ctor(system-string-system-io-filemode)

bkoelman:
I'm pretty excited about this!

Not only does this PR eliminate the `EndOfStreamException` (which occurs because the in-proc reader starts while the out-of-proc writer hasn't completed yet), but it also enables the reader to detect that the writer is still running, wait for it to complete, or, if it gets killed midway, log a warning. Even with `COVERLET_DATACOLLECTOR_INPROC_FLUSH_DISABLED` active, it reported the same coverage in a stress test on my Windows pc (18757 tests targeting net10/9/8), while 25 sequential runs never reported any warnings. I haven't been able to stress-test on .NET Framework.

See the commit messages for implementation details.

Fixes #1983, fixes #1981.

### DIFF
--- a/src/legacy/coverlet.collector/InProcDataCollection/CoverletInProcDataCollector.cs
+++ b/src/legacy/coverlet.collector/InProcDataCollection/CoverletInProcDataCollector.cs
@@ -2,9 +2,8 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections.Concurrent;
 using System.Diagnostics;
-using System.Reflection;
-using System.Text;
 using Coverlet.Collector.Utilities;
 using Coverlet.Core.Instrumentation;
 using Microsoft.VisualStudio.TestPlatform.ObjectModel.DataCollection;
@@ -42,6 +41,11 @@ namespace Coverlet.Collector.DataCollection
 
       _eqtTrace = new TestPlatformEqtTrace();
       _eqtTrace.Verbose("Initialize CoverletInProcDataCollector");
+
+      // Pre-create the registry bag before any instrumented assembly is loaded.
+      AppDomain.CurrentDomain.SetData(
+          ModuleTrackerTemplate.ModuleTrackerRegistryKey,
+          new ConcurrentBag<EventHandler>());
     }
 
     public void TestCaseEnd(TestCaseEndArgs testCaseEndArgs)
@@ -54,20 +58,19 @@ namespace Coverlet.Collector.DataCollection
 
     public void TestSessionEnd(TestSessionEndArgs testSessionEndArgs)
     {
-      foreach (Assembly assembly in AppDomain.CurrentDomain.GetAssemblies())
-      {
-        Type injectedInstrumentationClass = GetInstrumentationClass(assembly);
-        if (injectedInstrumentationClass is null)
-        {
-          continue;
-        }
+      // Use the AppDomain registry populated by RegisterUnloadEvents at module-load time.
+      var registeredHandlers = (ConcurrentBag<EventHandler>)AppDomain.CurrentDomain.GetData(ModuleTrackerTemplate.ModuleTrackerRegistryKey);
+      if (registeredHandlers is null)
+        return;
 
+      foreach (EventHandler handler in registeredHandlers)
+      {
+        string assemblyName = handler.Method?.DeclaringType?.Assembly?.FullName ?? "(unknown)";
         try
         {
-          _eqtTrace.Verbose($"Calling ModuleTrackerTemplate.UnloadModule for '{injectedInstrumentationClass.Assembly.FullName}'");
-          MethodInfo unloadModule = injectedInstrumentationClass.GetMethod(nameof(ModuleTrackerTemplate.UnloadModule), new[] { typeof(object), typeof(EventArgs) });
-          unloadModule.Invoke(null, new[] { (object)this, EventArgs.Empty });
-          _eqtTrace.Verbose($"Called ModuleTrackerTemplate.UnloadModule for '{injectedInstrumentationClass.Assembly.FullName}'");
+          _eqtTrace.Verbose($"Calling ModuleTrackerTemplate.UnloadModule for '{assemblyName}'");
+          handler.Invoke(this, EventArgs.Empty);
+          _eqtTrace.Verbose($"Called ModuleTrackerTemplate.UnloadModule for '{assemblyName}'");
         }
         catch (Exception ex)
         {
@@ -84,44 +87,5 @@ namespace Coverlet.Collector.DataCollection
     {
     }
 
-    internal Type GetInstrumentationClass(Assembly assembly)
-    {
-      try
-      {
-        foreach (Type type in assembly.GetTypes())
-        {
-          if (type.Namespace == "Coverlet.Core.Instrumentation.Tracker"
-              && type.Name.StartsWith(assembly.GetName().Name + "_"))
-          {
-            return type;
-          }
-        }
-
-        return null;
-      }
-      catch (Exception ex)
-      {
-        if (_enableExceptionLog)
-        {
-          var exceptionString = new StringBuilder();
-          exceptionString.AppendFormat("{0}: Failed to get Instrumentation class for assembly '{1}' with error: {2}",
-              CoverletConstants.InProcDataCollectorName, assembly, ex);
-          exceptionString.AppendLine();
-
-          if (ex is ReflectionTypeLoadException rtle)
-          {
-            exceptionString.AppendLine("ReflectionTypeLoadException list:");
-            foreach (Exception loaderEx in rtle.LoaderExceptions)
-            {
-              exceptionString.AppendLine(loaderEx.ToString());
-            }
-          }
-
-          _eqtTrace.Warning(exceptionString.ToString());
-        }
-
-        return null;
-      }
-    }
   }
 }
--- a/src/legacy/coverlet.collector/InProcDataCollection/CoverletInProcDataCollector.cs
+++ b/src/legacy/coverlet.collector/InProcDataCollection/CoverletInProcDataCollector.cs
@@ -67,7 +67,6 @@ namespace Coverlet.Collector.DataCollection
           _eqtTrace.Verbose($"Calling ModuleTrackerTemplate.UnloadModule for '{injectedInstrumentationClass.Assembly.FullName}'");
           MethodInfo unloadModule = injectedInstrumentationClass.GetMethod(nameof(ModuleTrackerTemplate.UnloadModule), new[] { typeof(object), typeof(EventArgs) });
           unloadModule.Invoke(null, new[] { (object)this, EventArgs.Empty });
-          injectedInstrumentationClass.GetField("FlushHitFile", BindingFlags.Static | BindingFlags.Public).SetValue(null, false);
           _eqtTrace.Verbose($"Called ModuleTrackerTemplate.UnloadModule for '{injectedInstrumentationClass.Assembly.FullName}'");
         }
         catch (Exception ex)
--- a/src/legacy/coverlet.collector/InProcDataCollection/CoverletInProcDataCollector.cs
+++ b/src/legacy/coverlet.collector/InProcDataCollection/CoverletInProcDataCollector.cs
@@ -74,11 +74,9 @@ namespace Coverlet.Collector.DataCollection
         }
         catch (Exception ex)
         {
+          _eqtTrace.Error("{0}: Failed to unload module '{1}' with error: {2}", CoverletConstants.InProcDataCollectorName, assemblyName, ex);
           if (_enableExceptionLog)
-          {
-            _eqtTrace.Error("{0}: Failed to unload module with error: {1}", CoverletConstants.InProcDataCollectorName, ex);
             throw new CoverletDataCollectorException($"{CoverletConstants.InProcDataCollectorName}: Failed to unload module", ex);
-          }
         }
       }
     }
--- a/src/legacy/coverlet.collector/InProcDataCollection/CoverletInProcDataCollector.cs
+++ b/src/legacy/coverlet.collector/InProcDataCollection/CoverletInProcDataCollector.cs
@@ -16,6 +16,7 @@ namespace Coverlet.Collector.DataCollection
   {
     private TestPlatformEqtTrace _eqtTrace;
     private bool _enableExceptionLog;
+    private bool _disableInprocFlush;
 
     private void AttachDebugger()
     {
@@ -34,10 +35,19 @@ namespace Coverlet.Collector.DataCollection
       }
     }
 
+    private void DisableInprocFlush()
+    {
+      if (int.TryParse(Environment.GetEnvironmentVariable("COVERLET_DATACOLLECTOR_INPROC_FLUSH_DISABLED"), out int result) && result == 1)
+      {
+        _disableInprocFlush = true;
+      }
+    }
+
     public void Initialize(IDataCollectionSink dataCollectionSink)
     {
       AttachDebugger();
       EnableExceptionLog();
+      DisableInprocFlush();
 
       _eqtTrace = new TestPlatformEqtTrace();
       _eqtTrace.Verbose("Initialize CoverletInProcDataCollector");
@@ -58,6 +68,12 @@ namespace Coverlet.Collector.DataCollection
 
     public void TestSessionEnd(TestSessionEndArgs testSessionEndArgs)
     {
+      if (_disableInprocFlush)
+      {
+        _eqtTrace.Verbose("COVERLET_DATACOLLECTOR_INPROC_FLUSH_DISABLED is set; skipping in-proc flush, hits will be written by process-exit handlers");
+        return;
+      }
+
       // Use the AppDomain registry populated by RegisterUnloadEvents at module-load time.
       var registeredHandlers = (ConcurrentBag<EventHandler>)AppDomain.CurrentDomain.GetData(ModuleTrackerTemplate.ModuleTrackerRegistryKey);
       if (registeredHandlers is null)
--- a/Documentation/KnownIssues.md
+++ b/Documentation/KnownIssues.md
@@ -2,62 +2,45 @@
 
 ## VSTest stops process execution early
 
-*Affected drivers*: msbuild (`dotnet test`) , dotnet tool(if you're using `--targetargs "test ... --no-build"`)
+*Affected drivers*: msbuild (`dotnet test`), dotnet tool (if you're using `--targetargs "test ... --no-build"`)
 
 *Symptoms:*
 
-* warning or error like:
+* Zero or missing coverage result, often only on CI.
 
-  `Unable to read beyond end of stream`
+* Warning messages in the build output:
 
-  `warning : [coverlet] Hits file:'C:\Users\REDACTED\AppData\Local\Temp\testApp_ac32258b-fd4a-4bb4-824c-a79061e97c31' not found for module: 'testApp'`
+  * ```
+    warning : [coverlet] Hits file not found for module 'testApp': 'C:\Users\REDACTED\AppData\Local\Temp\testApp_ac32258b-fd4a-4bb4-824c-a79061e97c31'. An incomplete write was detected; coverage data for this module is missing.
+    ```
 
-* zero coverage result (often only on CI but not on local)
+  * ```
+    warning : [coverlet] Hits file for module 'testApp' may be incomplete: a merge was in progress when the writer was killed. Coverage data from the merge is missing.
+    ```
 
-  ```bash
-  Calculating coverage result...
-  C:\Users\REDACTED\.nuget\packages\coverlet.msbuild\6.0.0\build\netstandard2.0\coverlet.msbuild.targets(21,5): warning : [coverlet] Hits file:'C:\Users\REDACTED\AppData\Local\Temp\testApp_ac32258b-fd4a-4bb4-824c-a79061e97c31' not found for module: 'testApp' [C:\Users\REDACTED\Documents\repo\testapp\testapp.Tests\testapp.Tests.csproj]
-  C:\Users\REDACTED\.nuget\packages\coverlet.msbuild\6.0.0\build\netstandard2.0\coverlet.msbuild.targets(21,5): warning : [coverlet] Hits file:'C:\Users\REDACTED\AppData\Local\Temp\testApp.Tests_ac32258b-fd4a-4bb4-824c-a79061e97c31' not found for module: 'testApp.Tests' [C:\Users\REDACTED\Documents\repo\testapp\testapp.Tests\testapp.Tests.csproj]
-    Generating report 'C:\Users\REDACTED\Documents\repo\testapp\lcov.info'
-
-  +---------------+------+--------+--------+
-  | Module        | Line | Branch | Method |
-  +---------------+------+--------+--------+
-  | testApp       | 0%   | 0%     | 0%     |
-  +---------------+------+--------+--------+
-  | testApp.Tests | 0%   | 100%   | 0%     |
-  +---------------+------+--------+--------+
-
-  +---------+------+--------+--------+
-  |         | Line | Branch | Method |
-  +---------+------+--------+--------+
-  | Total   | 0%   | 0%     | 0%     |
-  +---------+------+--------+--------+
-  | Average | 0%   | 0%     | 0%     |
-  +---------+------+--------+--------+
-
-  ```
-
-The issue is related to VSTest platform <https://github.com/microsoft/vstest/issues/1900#issuecomment-457488472>
+The issue is related to the VSTest platform ([vstest#1900](https://github.com/microsoft/vstest/issues/1900#issuecomment-457488472)):
 
 > However if testhost doesn't shut down within 100ms (as the execution is completed, we expect it to shutdown fast). vstest.console forcefully kills the process.
 
-Coverlet collects and writes hits data on process exit, if for some reason the process is too slow to close, it will be killed and we cannot collect coverage result.
-This happen also if there are other "piece of code" during testing that slow down process exit. We found problem for instance with tests that use RabbitMQ.
+`coverlet.msbuild` and the dotnet tool collect and write hits data on process exit. If the testhost is killed before that write completes, we cannot collect the coverage result.
+This can also happen if there are other "pieces of code" during testing that slow down process exit. We found problems for instance with tests that use RabbitMQ.
 
 *Solution:*
 
 The only way to get around this issue is to use **coverlet.collector** integration. With the collector, we're injected in test host through a in-proc collector that communicates with the VSTest platform so we can signal when we end our work.
 
+> [!NOTE]
+> `coverlet.collector` is normally not affected by this issue. But if the synchronous flush fails for a module, it falls back to writing on process exit as a safety net, and the same warnings above might be observed.
+
 ## Upgrade `coverlet.collector` to version > 1.0.0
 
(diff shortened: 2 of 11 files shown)


src/legacy/coverlet.collector/InProcDataCollection/CoverletInProcDataCollector.cs
```diff
@@ -35,13 +35,27 @@ private void EnableExceptionLog()
       }
     }
 
+    private void DisableInprocFlush()
+    {
+      if (int.TryParse(Environment.GetEnvironmentVariable("COVERLET_DATACOLLECTOR_INPROC_FLUSH_DISABLED"), out int result) && result == 1)
```
Bertk, line 40: Could you please update the documentation and consider renaming the environment variable? COVERLET_DATACOLLECTOR_INPROC_FLUSH_DISABLED is specific to coverlet.MTP, so it should follow the coverlet.MTP naming convention and be documented in the Environment variables section of the Coverlet.MTP integration guide:
[Documentation/Coverlet.MTP.Integration.md#environment-variables](https://github.com/coverlet-coverage/coverlet/blob/master/Documentation/Coverlet.MTP.Integration.md#environment-variables)

Bertk:
This PR resolves the known issue [„VSTest stops process execution early“](https://github.com/coverlet-coverage/coverlet/blob/master/Documentation/KnownIssues.md#vstest-stops-process-execution-early).  As a result, the corresponding entry in the Known Issues documentation should either be removed or updated to reflect the new behavior.

bkoelman:
While updating the docs, I discovered the following. I'm hesitant to change any code in MTP because I have no way to test it (besides running the repo's tests). Do you think I should apply similar changes or leave it as-is?

{AFA182F2-56CE-4E5F-95BD-8C6E484A86FA}

Bertk:
Many thanks for fixing this long-standing bug and improving the overall reliability of coverlet. Your contribution is greatly appreciated. 👍

Bertk (approved):

src/legacy/coverlet.collector/InProcDataCollection/CoverletInProcDataCollector.cs
bkoelman: I can't anymore, the PR is already merged. I only added it for my own stress testing, it's an unnatural code path. And it does not apply to MTP at all.

bkoelman:
> While updating the docs, I discovered the following. I'm hesitant to change any code in MTP because I have no way to test it (besides running the repo's tests). Do you think I should apply similar changes or leave it as-is?
> 
> {AFA182F2-56CE-4E5F-95BD-8C6E484A86FA}

As this got merged, I suppose the answer is: no, keep the MTP path broken and inefficient?

Bertk:
Sorry, that's my mistake. I thought this had already been resolved.

I'm currently traveling and don't have access to a development environment to verify it myself.

Could you please open a separate PR for the `coverlet.MTP` changes? For testing, you could also use the `coverlet-examples` repository: https://github.com/Bertk/coverlet-examples

The `coverlet-examples` repository contains basic configuration and integration tests, but it is not intended for high-load or stress-testing scenarios..

That will make it easier to review and validate the changes independently.

bkoelman:
> Could you please open a separate PR for the `coverlet.MTP` changes? For testing, you could also use the `coverlet-examples` repository: https://github.com/Bertk/coverlet-examples
>
> The coverlet-examples repository contains basic configuration and integration tests, but it is not intended for high-load or stress-testing scenarios..

Created #1988 to follow up. I don't think it needs to be stress-tested; the timing-sensitive process kill doesn't apply in MTP, as far as I understand.

bkoelman:
@Bertk @daveMueller Can this fix (and follow-up #2009) be released in the coming days or weeks? Or could you provide a rough ETA? I'd like to use the improvements made in v10. Please don't wait until .NET 11 ships.

Bertk:
> @Bertk @daveMueller Can this fix (and follow-up #2009) be released in the coming days or weeks? Or could you provide a rough ETA? I'd like to use the improvements made in v10. Please don't wait until .NET 11 ships.

I'll do my best to speed up another Coverlet release. Since I don't have admin permissions, I'd really appreciate some support from @daveMueller and @MarcoRossignoli when you have a moment. Thanks so much!

PR #2030, opened 2026-09-18
Use `--config-file` CLI arg in `coverlet.MTP`

### LINKED ISSUE #2015: Support `--config-file` flag for testconfig.json location (MTP)
**Describe the bug**
As a continuation on #1887 it would be awesome if coverlet could respect the same switch as the rest of the test configuration so we don't ave to copy it to the output directory. Currently coverlet ignores the config file if it does not live in the output directory.

**To Reproduce**
``` 
dotnet test --config-file testconfig.json --project .\subdir\projectName.csproj --diagnostic --diagnostic-verbosity Trace
```
with a project using coverlet.MTP. (and the correct global.json config to enable MTP)

```
{
  "platformOptions": {
    "Coverlet": {
      "Include": "[Calcasa.*]*",
      "Exclude": "[coverlet.*]*,[Microsoft.VisualStudio.TestPlatform.*]*,[testhost*]*,[coverlet.*.Tests?]*,[*]Coverlet.Core*,[*]Moq*,[*]TestBase*",
      "excludeByAttribute": "Obsolete,GeneratedCodeAttribute,CompilerGeneratedAttribute,ExcludeFromCodeCoverage,ExcludeFromCodeCoverageAttribute",
      "excludeByFile": "**/*Migrations/*.cs",
      "format": "cobertura,opencover",
      "useSourceLink": false,
      "singleHit": false,
      "includeTestAssembly": false,
      "skipAutoProps": true,
      "deterministicReport": false,
      "excludeAssembliesWithoutSources": "MissingAll"
    }
  }
}
```

**Expected behavior**
This should output the correct value in the diagnostic and respect the config items.

**Actual behavior**
It uses the default config.

    * Which coverlet package (collector, console, msbuild, ...) and version was used? coverlet.MTP 10.0.1
    * Which version of .NET is the code running on? 10.0
    * What OS and version, and what distro if applicable? Windows 11
    * Do you know whether it is specific to that configuration? Also fails on Ubuntu 26.04 x64

### LINKED ISSUE #2029: [BUG] `--config-file` via `TestingPlatformCommandLineArguments` ignored by `coverlet.MTP`
**Describe the bug**
When passing `--config-file` via the `TestingPlatformCommandLineArguments` MSBuild property, it appears that the configuration is ignored by `coverlet.MTP`, which means, for example, that if that config file specifies an `include` setting, it will not be applied.

**To Reproduce**
I've put together a minimal reproduction, which you will find attached here as `CoverletRepro.zip`.  Of note in that reproduction:

* There are two class library projects: `LibA` and `LibB`.
* `LibB` depends on `LibA`.
* There are two test projects: `LibA.Test` and `LibB.Test`.
* Each test project has its own `testconfig` file (`testconfig.LibA.json`, `testconfig.LibB.json`), which is passed via `--config-file` in `TestingPlatformCommandLineArguments`.
* In both cases, the config file specifies that _only_ code in the single library under test should be measured for coverage (i.e., `LibA.Test` should only be measuring coverage in `LibA`, `LibB.Test` only in `LibB`).

**Expected behavior**
`LibA.Test` and `LibB.Test` should report 100% coverage, respectively.

**Actual behavior**
`LibB.Test` reports less than 100% coverage, with its Cobertura report including code from `LibA.Test`, which is, in turn, showing up in that report as uncovered.  It's not weird that it isn't covered -- no test in `LibB.Test` covers anything in `LibA` -- but it's weird that `LibA`'s code is being included in `LibB.Test`'s coverage report at all.

    * Which coverlet package (collector, console, msbuild, ...) and version was used?  MTP
    * Which version of .NET is the code running on?  .NET 10.0.401
    * What OS and version, and what distro if applicable?  Windows 11
    * What is the architecture (x64, x86, ARM, ARM64)?  x64
    * Do you know whether it is specific to that configuration?  Given my investigation, it's doubtful that it's platform-specific.

**Additional context**
I've been able to narrow down a likely cause for the problem.  In the `CollectorExtension` class, the `ITestHostProcessLifetimeHandler.BeforeTestHostProcessStartAsync` method is considering both the command-line settings and settings loaded from a separately-searched config file (by calling `LoadConfigurationFileSettings`), but it's ignoring the configuration that may already be present in the `_platformConfiguration` member variable, if that variable means what I think it does.  (I'm still working through my understanding of that.  😃 )

To help confirm that theory, I tried adding the `--coverlet-include` CLI argument to `TestingPlatformCommandLineArguments`.  At that point, the coverage reports came out correctly (i.e., the `--coverlet-include` setting was being respected).

Also worth noting: I tried all of this using `coverlet.MTP` version `10.0.2-preview.38.gda5c649126` (the latest nightly release I could find), and that exhibited the same behavior.

[CoverletRepro.zip](https://github.com/user-attachments/files/32398065/CoverletRepro.zip)

alexthornton1:
When determining the configuration to be used in each test module execution, we're now using the `--config-file` CLI argument, loading configuration from the file it refers to (when that argument is present), rather than falling back to the legacy file search.


Bertk (approved):

Bertk:
Thank you for the regression fix.

Issue #1843, opened 2026-03-18
[BUG] Missing Coverage after moving to MTP


mellis-ao:
**Describe the bug**
After moving to coverlet.MTP from coverlet.msbuild (with xUnit) our coverage has dropped significantly using the opencover format.

**To Reproduce**

Generating an opencover report using the older VSTest system and again using MTP version yields notably less entries in resulting file without other code changes.

**Expected behavior**

Identical coverage results between coverlet.msbuild and coverlet.MTP with the same code / setup.

**Actual behavior**

MTP opencover results file has about 35% of the files the MSBuild version of the package does.

**Configuration (please complete the following information):**

For VSTest:
* coverlet.msbuild v8.0.0
* Microsoft.NET.Test.Sdk v18.0.1
* xunit.v3 v3.2.2
* xunit.runner.visualstudio v3.1.5

For MTP:
* coverlet.MTP v8.0.0
* Microsoft.Testing.Platform v2.1.0
* xunit.v3.mtp-v2 v3.2.2

Both running on .Net 10, issues present via azure pipeline running ubuntu-latest and locally on Windows 11.

**Additional context**

Command run for VSTest on windows:

`dotnet test <Project> --configuration <Config> --nologo /p:SingleHit=true /p:CollectCoverage=true /p:CoverletOutputFormat=opencover   `

Summary header of VSTest opencover result:

`<Summary numSequencePoints="25080" visitedSequencePoints="1346" numBranchPoints="1970" visitedBranchPoints="172" sequenceCoverage="5.36" branchCoverage="8.73" maxCyclomaticComplexity="5097" minCyclomaticComplexity="242" visitedClasses="215" numClasses="607" visitedMethods="446" numMethods="6914" />`

Command run for MTP on windows:

`dotnet test --project <Project> --configuration <Config> --report-xunit-trx --coverlet --coverlet-output-format opencover --coverlet-include <Own Modules> --coverlet-single-hit`

Summary header for MTP opencover result:

`<Summary numSequencePoints="2596" visitedSequencePoints="399" numBranchPoints="808" visitedBranchPoints="51" sequenceCoverage="15.36" branchCoverage="6.31" maxCyclomaticComplexity="467" minCyclomaticComplexity="11" visitedClasses="79" numClasses="226" visitedMethods="112" numMethods="624" />`

If there is anything you need from me to test or more info let me know.

Same behaviour when running using `--solution`, the logs indicate the module is being instrumented but not picking up all the files.

PR #2033, opened 2026-09-21
Add regression test for #1727 (state machine excluded via ExcludeByAttribute=CompilerGeneratedAttribute)

### LINKED ISSUE #1727: [BUG] Constructorless ControllerBase implementations produce no code coverage
**Describe the bug**
When producing code coverage for a WebApi/MVC project, if a Controller has no explicit constructor, it may not produce any code coverage results for the entire file.

These Controllers/Classes seem to be affected by the `ExcludeByAttribute=CompilerGeneratedAttribute` switch - as removing this option produces code coverage for these files.

**To Reproduce**
Sample Repository: https://github.com/chrish619/coverlet-coverage-issue-20250103/

Controller must inherit from either ControllerBase, or another constructor-less Controller abstraction.
`ExcludeByAttribute` in either MSBuild property flags, or a runsettings file, must contain `CompilerGeneratedAttribute`.

**Expected behavior**
WebApi/MVC Controllers *without* an explicit constructor, should produce code coverage results.

**Actual behavior**
WebApi/MVC Controllers *without* an explicit constructor produces no code coverage results when
`ExcludeByAttribute=CompilerGeneratedAttribute` is used.
These Controllers may use Actions with `[FromServices]` dependency injection - whilst also not declaring a constructor.

WebApi/MVC Controllers *with* an explicit constructor will always produce code coverage results.

* Which coverlet package and version was used? 6.0.2/6.0.3
* Which version of .NET is the code running on? .Net8/.Net9
* What OS and version, and what distro if applicable? Windows 10
* Do you know whether it is specific to that configuration? Specifically caused by `ExcludeByAttribute=CompilerGeneratedAttribute`

Eljees:
Fixes #1727.

## Root cause

When `ExcludeAttributes` (`ExcludeByAttribute` on the CLI) is set to `CompilerGeneratedAttribute`, `Instrumenter.IsTypeExcluded` used to exclude *any* type carrying that attribute — including the compiler-generated `async` state-machine type. Since the state machine is where the actual method body lives, excluding it meant the whole `async` method silently dropped out of the coverage report, even though nothing else about it was excluded.

This was already fixed upstream by 3a241bc (#1940, merged 2026-05-21) for the closely related #1843 — `IsTypeExcluded` now recognizes compiler-generated state-machine types (`IAsyncStateMachine` / `IEnumerator` / `IAsyncEnumerator`) and no longer treats `CompilerGeneratedAttribute`/`GeneratedCodeAttribute` as grounds for excluding *them* specifically, while still honoring explicit exclusions for everything else. That fix has not shipped in a NuGet release yet (`coverlet.collector` 10.0.1 predates the merge by 3 days), and — more relevantly for this PR — none of the four existing `AsyncAwait_Issue_1843_*` tests actually exercise the exact configuration reported in #1727 (`ExcludeAttributes: ["CompilerGeneratedAttribute"]` on a constructorless controller-like class with no explicit constructor). This PR only adds that missing regression test; `src/` is unchanged.

## Testing

Added `AsyncAwait_Issue_1727_ExcludeByCompilerGeneratedAttribute_StateMachineStillInstrumented` to `CoverageTests.AsyncAwait.cs`: instruments `AsyncAwait.ConfigureAwait()` with `excludeAttributes: ["CompilerGeneratedAttribute"]` and asserts the method body lines are still covered (i.e. the state machine was not excluded). `TestInstrumentationHelper.Run<T>()` needed a way to pass `ExcludeAttributes` through to `CoverageParameters` for this, which it previously hard-coded to empty — added as an optional trailing parameter, so all ~20 existing call sites in the file are unaffected.

Verified red before the fix and green after:
- With the #1940 carve-out temporarily removed from `Instrumenter.IsTypeExcluded`, the new test fails: `Not all requested line found, 89, 90`.
- With the fix restored, the new test passes in isolation, and the full `CoverageTests.AsyncAwait.cs` class (24 tests) passes as well.

(One pre-existing test in that class, `AsyncAwait_Issue_669_2`, depends on an outbound HTTP request and is independently flaky under network-restricted sandboxes — unrelated to this change; it is not touched by this diff.)


Bertk (approved):

Bertk:
Thank you for the contribution.

